Issue #35, opened 2015-11-25
Improve build tasks for preferences


gornostal:
* `gulp build` should create a build dirrectory with all necessary files
* `gulp build` should concatenate all css and js files into one
* `gulp` should create `index.html` file in `.tmp`. It will be used by Ulauncher

gornostal:
Priority 1

gornostal:
working on this

PR #44, opened 2015-12-07
#35 - updated gulp file to build to the same path for dev and prod

gornostal:


data/preferences/src/components/about-page/ulauncher.svg.html
```diff
@@ -0,0 +1 @@
+../../../../media/icons/hicolor/ulauncher.svg
```
gornostal: It's a symlink. `ng-include` didn't work with `.svg` files, so I did this trick with `.svg.html`

gornostal:
@EugeneChu, please review.

Now both dev and prod environment will load index.html from the same path.

gornostal:
OK. I'm merging it.

PR #279, opened 2018-09-01
Python3 migration

### LINKED ISSUE #217: Migration to python3.
The End Of Life date for Python 2 is in 2020. 
https://stackoverflow.com/questions/4836375/end-of-support-for-python-2-7

cjbassi:
I started the process to port to Python 3, but this is still a WIP. A lot of the tests are failing, but the program starts and runs. Let me know if any of the commits should be squashed. I tried to keep things as separate as possible for the time being.

The only other thing is that extensions also have to port to python 3.

Closes #217 

TODO:
- [x] Errors
    - [x] EOFError
    - [x] preferences window errors
- [ ] Testing
    - [ ] packages
        - [ ] AUR
    - [x] extensions
    - [ ] docker
- [ ] fix tests


gornostal:
Hi @cjbassi 
Thanks a lot for helping to port Ulauncher to Python 3. I really appreciate your efforts.

There's only one thing I'm worried about, which is updating dependencies for all supported Linux distros.
I think it's not going to be a problem with Arch, but I'm not sure about Ubuntu, Fedora, SUSE, and CentOS.

Last time I tried to port the app to Python 3, I got stuck because Ubuntu 14.04 didn't have python3-levenshtein package. Plus, some other distros didn't have a few more required packages.

While you're working on code changes I'll try to check the dependencies.

As for Ulauncher extensions, ideally maintainers should port them to. But I think we can make extension library backward compatible.

cjbassi:
Also, the build process still relies on Python 2 since node-gyp doesn't support Python 3 yet.

friday:
So this would depend on both Python 2 and 3? Perhaps then extensions can keep using v2 unless they specify v3 in the manifest? That way extensions maintainers can defer the upgrade until everyone are using the Python 3 Ulauncher version.

(providing distro support won't be an issue)

erayerdin:
Hi guys,

I was wondering what is the situation on this issue. I have been developing a project called [pyairmore](https://github.com/erayerdin/pyairmore/) and would like to integrate it to Ulauncher.

rumjantsevv:
Is it already possible to write extensions on python3 ?

gornostal:
@rumjantsevv it's not ready yet
There's a branch `python3`, but I don't recommend using it. I'm planning to change the way extensions are downloaded

rumjantsevv:
@gornostal oh, i see.
Do you have any list of tasks  should be done before publishing py3 version?
Can you publish it, so community can help you with the migration.
Also will be useful if you will tell a bit your thoughts about py3 version, how it should be done, what do you except from it, so people will also develop their parts of code in the way you planned.

gornostal:
@rumjantsevv I will publish a list of contribution-friendly tasks as soon as I have a new schema of manifest.json defined an a working proof of concept.

The problem is that in current version of Ulauncher I didn't take into consideration forward compatibility of extensions. Meaning that unless I make changes to the manifest file and to how extensions are downloaded, the existing extensions will crash with the new Ulauncher release.

PR #281, opened 2018-09-07
Fix typo

friday:
I found this in the generated documentation. I don't know how the generation works but since you do, this could be of interest.

http://docs.ulauncher.io/en/latest/extensions/events.html#ulauncher.api.shared.event.KeywordQueryEvent


gornostal:
Thanks

Issue #284, opened 2018-09-07
API suggestion: add event.get_keyword_id()


friday:
`event.get_keyword()` is very useful, since extensions may use multiple keywords, and in this case there's a need to distinguish between them in `KeywordQueryEventListener`.

However, it gets the user defined keyword, not the keyword id. Since users can change these it has to be converted back to the keyword id.

This is not hard by any means, but it feels like a hack having to do it.
https://github.com/brpaz/ulauncher-hash/pull/4/commits/be88b73b7cc14416893c3a417ecf67900b3d3577

goodwillcoding:
@friday send a PR?

gornostal:
Hey @friday 
Feel free to rebase your PR over `dev`. I think we can now merge it.

friday:
I will try to do that soon. Don't have the time right now I'm afraid.

brpaz:
Any chance for this to be added?

friday:
This is fixed in the Ulauncher v6 branch and PR #870. It won't be released for a while, but we made a tracking issue if you want to be notified #869

PR #287, opened 2018-09-11
Add event.get_keyword_id()

friday:
Implements #284 

Criticism is welcome. I'm not a python dev and don't know the Ulauncher internals or your conventions well

Update: Know the tests are broken. Will have a look at this soon

### DIFF
--- a/tests/api/server/test_DeferredResultRenderer.py
+++ b/tests/api/server/test_DeferredResultRenderer.py
@@ -4,6 +4,7 @@ import pytest
 from ulauncher.api.server.DeferredResultRenderer import DeferredResultRenderer
 from ulauncher.api.server.ExtensionController import ExtensionController
 from ulauncher.api.server.ExtensionManifest import ExtensionManifest
+from ulauncher.api.server.ExtensionPreferences import ExtensionPreferences
 from ulauncher.api.shared.action.BaseAction import BaseAction
 from ulauncher.api.shared.event import BaseEvent, KeywordQueryEvent
 from ulauncher.search.Query import Query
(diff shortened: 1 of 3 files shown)


gornostal:
Thanks for the PR @friday 
Implementation of `get_keyword_id` looks good to me. However, I need to bump API version and run some checks on different extensions before merging it.

friday:
I'm not in a hurry, since this can be done in the client code as well. Don't think this should be API-breaking but you would know better. Perhaps it makes sense to coordinate this with other features. 

As mentioned I'm aware I broke the tests (and also haven't added any for `get_keyword_id()`). Will look at the broken test(s) soon.

friday:
Fixed the tests, but you may want to do it differently.

friday:
Should work now. It actually did before too, with just a lint error. I fixed it with 4b5900b. Personally I think this commit is dumb and more likely to obfuscate errors than fix them. Feel free to revert or skip it.

I updated https://github.com/friday/ulauncher-gnome-settings to use the new API, so you can test it, at least of you have Gnome Settings (formerly Gnome Control Center).

gornostal:
Cool. Thank you!

gornostal:
@friday sorry, I had to revert.

I didn't notice at first that you pass `preference` object into event. That causes an error.
Also, an event object is supposed to be lightweight. I would not pass anything like that.

It's better to have the function that returns keyword ID in the extension. You have the entire extension object with preferences in every `on_event` callback.

friday:
Alright. So every extension that needs the id has to do this then?

I didn't notice any errors while linting or building. I'm also unable to get the docker image running (I'm running it on my native environment), but I don't think that should matter.

gornostal:
> I didn't notice any errors while linting or building
I've changed a few things after which it caused an error.

> So every extension that needs the id has to do this then?

I think we could add a helper function `get_keyword_id(preferences, query)` to `ulauncher.api.client.utils`

PR #288, opened 2018-09-12
Comply with XDG spec on .desktop location, dedup .desktop search per XDG_DATA_DIR and fix search to not collapse same Name

goodwillcoding:
This fix addresses 2 related situations. Both fixes are needed to help Ulauncher find .desktop folders in the same way as XDG spec defines it.

Fix 1:
Dedup the scanned .desktop files following the XDG spec where only the .desktop file found in the first XDG_DATA_DIR paths is the one we use. This is not happening currently and files located in the $HOME/.local/share/applications were not found/prefered in some scenarios where they file in /usr/share or other folder were either found first or updated on user watch (i.e. upgrade of software). Specific issue encountered is that I had copied and modified a specific desktop file from `/usr/share/applications` to `~/.local/share/applications` but it was not  prefered.

Changes for this fix are only in util/desktop/reader.py

To clarify with an example:
Assuming XDG_DATA_DIR="/foo:/bar" and both /foo/firefox.desktop and /bar/firefox.desktop exist, we should always pick /foo/firefox.desktop to respect the XDG spec.

Fix 2:
If two different .desktop files have the same Name= value only one currently is shown in the search. However, per XDG spec the Name value does not need to be unique, and only the .desktop file name found in first XDG_DATA_DIR path should be (see Fix 1). Specifically I had firefox and firefox-esr packages installed side by side which had a firefox.desktop and firefox-esr.desktop both with the same name of "Firefox Web Browser". WIthout this fix only one was showing (one that was scanned after the first one and replaced first result in the sqlite db.

Changes for this fix are only in search/apps/AppdDb.py

### DIFF
--- a/tests/search/apps/test_AppDb.py
+++ b/tests/search/apps/test_AppDb.py
@@ -13,27 +13,94 @@ class TestAppDb(object):
     @pytest.fixture
     def db_with_data(self, app_db):
         values = [
-            {'name': 'john', 'description': 'test', 'search_name': 'john',
-                'desktop_file': 'john.desktop', 'icon': 'icon'},
-            {'name': 'james', 'description': 'test', 'search_name': 'james',
-                'desktop_file': 'james.desktop', 'icon': 'icon'},
-            {'name': 'o.jody', 'description': 'test', 'search_name': 'o.jody',
-                'desktop_file': 'o.jdy.desktop', 'icon': 'icon'},
-            {'name': 'sandy', 'description': 'test', 'search_name': 'sandy',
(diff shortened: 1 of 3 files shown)


goodwillcoding:
@gornostal : if you need more explanation, I am happy to provide it

goodwillcoding:
@gornostal ... ugh, broke tests ... will fix tomorrow when I get the build chain working correctly. Meanwhile any feedback is welcome.

goodwillcoding:
@gornostal tests fixed, did a tad of reformatting on fixtures cause it was a tad hard to read them, I can try to put them back if you want

Issue #133, opened 2017-05-18
Add a button to enable/disable the shortcuts


NetBender:
At the moment we can only edit or remove them.

gornostal:
Sounds useful.

IgorMael:
I think I can do it.
It's basically 

-  In UI, add a new checkbox toggle visibility" icon beside  the edit and remove ones
-  In Python code save that flag to the ShortcutsDb
-  Do some css changes to the item visibility on the list
-  In Python, change to not show hidden itens on search

right?

gornostal:
Please hold on with this one @IgorMael 

I haven't added `contribution-friendly` label to this issue because now I'm not sure if enhancement is really useful. Maybe for few people, most of the user can just remove the shortcut if they don't want to use it.

PR #298, opened 2018-09-29
Feature: Option to enable/disable shortcuts

goodwillcoding:
Rationale: Alt+<num/letter> is not required for everyone, since it's
often faster to type the completion then use Alt key.
I found this to be most problematic on laptops where the Alt key
is small and lifting fingers to get to Alt+<key>.

PR below includes tests for all the new and updated scenarios.
Also the enabling of shortcuts is still the default in keeping with current behaviour.

This PR might also address #133 (can not tell from it's phrasing)


goodwillcoding:
@goodwillcoding is there anything you need from me on this PR. Or do you think it's too much?

gornostal:
>  is not required for everyone, since it's often faster to type the completion then use Alt key.

I agree with that. I'm not sure what problem this improvement solves exactly. It only removes Alt keys from the UI and disables them, right?

goodwillcoding:
> It only removes Alt keys from the UI and disables them, right?

That's correct, with togglable option. The default is still to show/enable Alt keys

goodwillcoding:
I think? it's what #133 was requesting.

gornostal:
> I think? it's what #133 was requesting.

Ah, that's actually a different thing. The requestin #133 is about disabling shortcuts in the Preferences. We have shortcuts there that people may want to disable so they don't appear among search results.

Sorry about the confusing title.

goodwillcoding:
@gornostal ... all good. any thought on the feature itself?

gornostal:
Sorry, but I don't think this improvement is necessary.
The reason is that I want to keep preferences as simple as possible, and also to have as minimum code as possible in order to make maintenance easier.

Once we migrate to python3 I will merge #300. I think that's a really good improvement

goodwillcoding:
kie, I'll just keep my branch of this code.

goodwillcoding:
@gornostal ... if we want to pair down the preferences code, should we consider moving it into Gtk and out of JS/Vue? That would definately cut down on maintenance.

gornostal:
> should we consider moving it into Gtk and out of JS/Vue?

That was the first thing I've tried and I realized that customizing UI in GTK is pain. Whe I re-implemented Preferences in JS it was much easier to add stuff and to make the UI look pretty.

gornostal:
I'm going to close this PR now

PR #299, opened 2018-09-30
Development toolchain cleanup

goodwillcoding:
Hi @gornostal,

This PR has changes I found helped me develop. Each category of change is broken up in individual commit in case you only like some. If you do not like any, I understand, I'll sidecar them for my own development. The changes are:

1. Change all bash sheebangs from '#!/bin/bash` to '#!/usr/bin/env bash'
 
**Rationale:** not all linux systems have /bin/bash and common practice to fix this is to use /usr/bin/env

2. Added dev-clean.sh and dev-nkpy.sh

**Rationale:**
 
dev-clean.sh removes the files installed by dev-run.sh, which means that system level working ulauncher can now be used.

dev-nkpy.sh is useful to remove pyc/pyo and __pycache__ folder. This ESPECIALLY helps when the pyc files are generated inside the docker container during the test run and end up being owned by `root:root` with permission of `rw-r-r both` in docker and on the host system which prevents their regeneration when their respective py files are changed.

3. Stop DistUtilsExtra.auto from poluting stdout with a large set of mostly useless messages about which files it does not recognize (and hence can not handle). It forces one to always scroll past them during development which can be frustrating. Message look like this
```
WARNING: the following files are not recognized by DistUtilsExtra.auto:
Dockerfile.build
Dockerfile.build-arch
Dockerfile.build-rpm
PKGBUILD.template
build-utils/aur-update.py
...
```
If you still need these messages during packaging I can add environmental variable work around which will show them when needed. **Action item: let me know if you need this**

4. I applied pep8 to setup.py (minor change if it bothers you I'll revert it)

### DIFF
--- a/build-utils/build-deb.sh
+++ b/build-utils/build-deb.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ######################
 # Builds Ulauncher.deb
--- a/build-utils/build-doc.sh
+++ b/build-utils/build-doc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -ex
 
--- a/build-utils/build-preferences.sh
+++ b/build-utils/build-preferences.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #############################################
 # Builds Ulauncher Preferences UI with nodejs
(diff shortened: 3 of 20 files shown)


gornostal:
Thanks for the improvements. It looks good to me :+1: 

One thing I noticed is that you configured pep8 to use 80 line length (probably it's the default), but for Ulauncher I'm using 120 characters. I was running flake8 linter though.
Since this PR affects only config.py, I'm going to merge it anyway.

It would be nice if you could create `.pep8` file or something that the linter uses to override that setting.

goodwillcoding:
Ah @goodwillcoding ... ok I see this in setup.cfg
```
[tool:pytest]
pep8maxlinelength = 120
pep8ignore = E402
```

I'll see what generic file helps with this.

PR #300, opened 2018-09-30
WIP: First pass at also searching desktop file keywords and descriptions using updated fuzzy search

goodwillcoding:
Hi @gornostal,

I took a pass at adding ability to also search keywords and description and have had to updated the fuzzy search to make it work. I'd be happy to get your feedback on it, branch is fully working.

Current changes are.

1. add first keywords, then description to search_name. description has punctuation stripped from it for easier search.

2. updated fuzzy search to compute match a number by doing the levenshtein distance between the query and EACH word in search_name. All the entries which have distance above the limit specified to SortedList are added up together. This number is then divided by the number "matchy" words that been found to be above limit to create a ratio which serves as the score. This helps avoid the situation where some desktop entries have a lot more words because they have extensive keywords and description.

3. A large artificial boost is added to the if a query matches word in the search_name exactly, but only 1 time. This helps avoid score going when the same word is found multiple times (i.e. "this application is for document reading, it handles document well, document, document, document)

4. A smaller artificial boost is applied to the complete but partial word boost (i.e. query of word in "password".

I am still playing around with the algo, specifically considering adding only including entries with total_limit of N because right now 9 results are almost always returned.

Obvious downsides:

1. Algo is a tad slower because it does more. I have not seen major issues on my system but who knows ¯\\\_(ツ)_/¯

2. Given the size of description and keywords. highlighting is obviously and issue. Solutions like "extra lines of text below mainy entry" or "informational sidebar" can be considered. Personally, I do not need either, but if had the choice I'd go for a small sidebar on the right that has name, keyword, description over "more lines in the description" because of laptop vertical space which always seems to be lacking.

If you would like to tweak the algo together consider creating an intergration branch we can both push to.

P.S. I have not done tests yet because this is more of research branch for now.

### DIFF
--- a/ulauncher/search/SortedList.py
+++ b/ulauncher/search/SortedList.py
@@ -10,6 +10,9 @@ class SortedList(object):
     """
 
     def __init__(self, query, min_score=30, limit=9):
+        print ""
+        print ""
+        print "=> => => =>"
         self._query = query.lower().strip()
         self._min_score = min_score
         self._limit = limit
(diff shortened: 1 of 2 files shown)


friday:
Awesome taking this on @goodwillcoding! :+1: I haven't actually tested the pr but you seem to know what you are doing in general so I'm sure this is no exception.

Perhaps relevant: Keywords in manifest.json have descriptions too. This doesn't mean the logical conclusion is they should be matched, or that this couldn't be added after this, but I just wanted to bring it up.

friday:
Went a complete different route on this, with several fixes to the code (V6 branch), and think that was the right move, because it's working great now. Not splitting up the query per word though, which would be good to support keywords in particular, but I think the fuzzy search is better left simple and effective.

Issue #295, opened 2018-09-24
Default terminal for htop


yegorf1:
Can't change default terminal for htop

**Steps to reproduce:**
Just open htop from ulauncher and it will be openned from `gnome-terminal`, but my default terminal is Tilix

**Ulauncher version**: v4.1.1.r1
**Linux distro**: Elementary loki 0.4
**Desktop environment**: Pantheon

goodwillcoding:
@yegorf1 how does elementary choose it's default terminal. I only ask because only DE I know Terminal can be looked up is XFCE (using an exo util)

yegorf1:
@goodwillcoding, I changed my default terminal with these commands: 

```Bash 
gsettings set org.gnome.desktop.default-applications.terminal exec tilix 
gsettings set org.gnome.desktop.default-applications.terminal exec-arg '' 
``` 

But ulauncher uses `gnome-terminal` while elemenatery default terminal is `pantheon-terminal`

goodwillcoding:
@yegorf1 yeah I agree with @friday assessment, `org.gnome.desktop.default-applications.terminal` is deprecated for GNOME (or GNOME based DEs). Other DE's have their own magical ways to specify default terminal with various OS workaround in place. The fix @friday suggested is probably the most sane one with some alternative of auto detection ... maybe @yegorf1 or @friday ya wanna take a swing at getting that done?

yegorf1:
@friday's fix looks best in this situation I think. Closing issue for now

PR #301, opened 2018-09-30
Add default terminal setting for desktop entries

friday:
See #295 

@goodwillcoding suggested auto-detecting *could* be possible. I don't think it is, but I may have missed something. The `TERM`-variable is not for this and the dconf entries are deprecated. So even if I could use them I think that would be making matters worse.

Either way, I made it into a setting "terminal-exec", which is added to the advanced section of the preferences window as "Terminal Launch Command" (not thrilled about the name but I needed to fine one that implied that the exec argument should be included).

It updates with the "change"-event, so when the focus is lost after having changed, rather than when it actually changes.

screenshot from 2018-09-30 23-38-15

The exec flag is usually `-e` or `-x` it seems (with somewhat different meanings). For gnome-terminal it's `--` though.

### DIFF
--- a/data/preferences/src/components/pages/Preferences.vue
+++ b/data/preferences/src/components/pages/Preferences.vue
@@ -87,7 +87,7 @@
     <table>
       <tr>
         <td>
-          <label for="terminal-exec">Terminal launch command</label>
+          <label for="terminal-exec">Terminal Launch Command</label>
         </td>
         <td>
           <b-form-input
--- a/data/preferences/src/components/pages/Preferences.vue
+++ b/data/preferences/src/components/pages/Preferences.vue
@@ -86,6 +86,17 @@
 
     <table>
       <tr>
+        <td>
+          <label for="terminal-exec">Terminal launch command</label>
+        </td>
+        <td>
+          <b-form-input
+            style="width:250px"
+            id="terminal-exec"
+            v-model="terminal_exec"></b-form-input>
+        </td>
+      </tr>
+      <tr>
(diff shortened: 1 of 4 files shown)


goodwillcoding:
@friday .. yeah, what I meant by autodetection is by checking which DE is running via XDG_CURRENT_DESKTOP and GDMSESSION, see: https://askubuntu.com/questions/72549/how-to-determine-which-window-manager-is-running. 

Then we would need to manage a list in an ongoing manner of looking up or launching the default terminal for each DE (for example for XFCE this is `exo-open --launch TerminalEmulator`)

I'd say that's a lot of work unless someone wants to do it.

goodwillcoding:
@friday ... Very nice!

I'd say you need to add tests for new code branches the methods in `LaunchAppAction.py` and `PreferencesUlauncherDialog.py`

Please see:
`tests/api/shared/action/test_LaunchAppAction.py`
`PreferencesUlauncherDialog.py`

There are existing tests there for you to model after.

friday:
Thanks for the pointers. I'll try to have that added. I'm new to Python so pytest and fixtures etc is far from second nature for me.

varac:
@friday @gornostal What's the state of this PR ? I'd love to get it merged since no terminal starts for me right now and I don't know how to properly fix this, so configuring the terminal n the ulauncher config would be best.

gornostal:
Thanks for this PR @friday 
I'll include it in the next release

gornostal:
Released in 5.7.0

Issue #305, opened 2018-10-04
Wish: Add to documentation: Hint about desktop entries indexing


kstenschke:
First of all thankyou very much for your efforts on ulauncher, i am happy to have finally found what looks at the moment like the best linux application launcher to me :+1: 
IMO the 3 features that make ulauncher the perfect tool, are shortcuts, the extension system AND that desktop entries in ~/.local/share/applications/ are indexed as applications.

The documentation at the ulauncher website, github and http://docs.ulauncher.io/en/latest/index.html should IMO really contain at least a short hint about desktop entries.

friday:
> please go ahead fork it, and submit a PR with the changes you think are needed and we can consider them

I think this was well intended as a pointer to where to make the change, but the formulation was somewhat coercive and unfortunate. Forking and submitting PR's is great, and while the possibility could be presented, it should not be *expected* from people creating issues imo. Creating the issue is a contribution in the first place.

gornostal:
> First of all thankyou very much for your efforts on ulauncher, i am happy to have finally found what looks at the moment like the best linux application launcher to me +1

I'm happy to hear that :smiley: 

Since #308 is resolved I'm going to close this issue too.

PR #308, opened 2018-10-10
Update index.rst

kstenschke:
Added description of launch-targets other than applications: desktop-entries, shortcuts, extensions

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,6 +11,30 @@ Everything else is on
 `Github Wiki <https://github.com/Ulauncher/Ulauncher/wiki/>`_.
 
 
+Launch-Targets
+==============
+
+Additionally to launching Applications, Ulauncher can launch the following targets:
+
+Desktop-Entries
+---------------
+
+Ulauncher indexes desktop-entries in ``~/.local/share/applications`` the same as applications.
+Using desktop-entries, launchers that require no query (no space or arguments other than the name/keyword) can be created easily.
+
+
+Shortcuts
+---------
+
+Shortcuts configure Web-URLs using a Query-String or Script, to be opened in the default browser.
+
+
+Extensions
+----------
+
+Extensions are custom scripts, that extend Ulauncher with arbitrary features, such as displaying options generated from the extension in Ulauncher, and evaluating (launching) these options via Ulauncher or the extension.
+
+
 Custom Color Themes
 ===================
 


docs/index.rst
```diff
@@ -11,6 +11,30 @@ Everything else is on
...
 
+Launch-Targets
+==============
+
+Additionally to launching Applications, Ulauncher can launch the following targets:
+
+Desktop-Entries
+---------------
+
+Ulauncher indexes desktop-entries in ``~/.local/share/applications`` the same as applications.
```
goodwillcoding: @kstenschke so desktop entries in `~/.local/share/applications` is the same entries in `/usr/share/appplications`, i.e. Application desktop entries. Freedesktop format searches in multiple folders (XDG_DATA_DIRS). Please read https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html

docs/index.rst
```diff
@@ -11,6 +11,30 @@ Everything else is on
...
+Using desktop-entries, launchers that require no query (no space or arguments other than the name/keyword) can be created easily.
+
+
+Shortcuts
```
goodwillcoding: @kstenschke ... `Shortcuts` terminalogy applies to lauching both web searches and extensions so this would be confusing as well.

docs/index.rst
```diff
@@ -11,6 +11,30 @@ Everything else is on
...
+---------
+
+Shortcuts configure Web-URLs using a Query-String or Script, to be opened in the default browser.
+
+
+Extensions
```
goodwillcoding: Extensions are described in detail on the main Ulauncher site. The docs you are updating are for developers who by the time of writing an extensions should have read what an extension is.

goodwillcoding:
@kstenschke ... I'll leave to @gornostal but the auduence you are addressing in these docs are developers who before developing and extensions would need read the information on the main site which has most of the information you have up here (and some infromation is a bit off as I mentioned above, like desktop entries).

kstenschke:
I suggested adding this description especially because desktop-entries are a simple but powerful option to create custom launchers, which i myself did not read about before i wrote extensions and than found out that desktop launchers  covered all i needed, without further reading the extension documentation. Additionally i noticed issues from other  people lookimg for ways to add launch-items without the need of pressing space after the keyword (=possibly desktop entries). Therefor to my impression, mentioning desktop entries seems to be indeed missing from and relevant to the documentation.

Am Mittwoch, 10. Oktober 2018 schrieb goodwillcoding:
> @kstenschke ... I'll leave to @gornostal but the auduence you are addressing in these docs are developers who before developing and extensions would need read the information on the main site which has most of the information you have up here (and some infromation is a bit off as I mentioned above, like desktop entries).
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/Ulauncher/Ulauncher/pull/308#issuecomment-42861259

-- 
Sent from my Jolla

friday:
I'm not authorized to accept or reject PR's, but if I was I would reject this. Besides the pointers already made, the format is not the same as the rest of the page (it's an index page with mostly links). It introduces new terminology like "Launch-Targets", which is something I would avoid doing without consulting the maintainer. Also desktop-entries is how all apps get indexed in all launchers or menu apps. They're not something separate from / in addition to apps like the pr suggests. Also line 10 "Currently only docs about extensions are available here." would no longer be true with this PR.

Reviewing PR's often needs more effort from maintainers than making the changes themselves. In particular "drive-by" contributions when the author doesn't want to invest time into understanding how everything works, styleguides etc. The "constructive" feedback loop with code reviews is nicer, but really time consuming.

We should keep #305 open, and now that this PR is linked to it it could serve as a starting point for a new PR or a discussion.

I did take the liberty of deleting the wiki page you added, since it was a duplication of this PR which itself hasn't been approved, and won't be without modifications. The wiki is mostly for contributions to Ulauncher, while the docs is mostly for extensions. I think the docs is the right place, but more effort should be spent on making the docs general (not just for extensions), so that the information here would have a place to fit in.

docs/index.rst
gornostal: I think that's fine

gornostal:
I think the information would be useful for those who want to have a better understanding of Ulauncher features. 
@kstenschke  I've brought back the wiki page, updated it according to comments from @goodwillcoding https://github.com/Ulauncher/Ulauncher/wiki/Ulauncher-Features

I agree with @friday that http://docs.ulauncher.io is primarily for documentation about extensions. So I'm going to close this PR.
I think the wiki is the place where that info should live.

Thanks for the contribution.

PR #309, opened 2018-10-14
WIP: first pass at Python2 and Python3 compatability

goodwillcoding:
@gornostal this gets ulauncher working on both py2 and py3. I prolly missed a few things (likely some unicode/str bits as always). The compat library is based on Pyramid framework (a project I am a developer) for so it's a well tested approach.

Have not done the packaging and dependencies yet, the real questio there is whether there will be 2 packages now (one for py2 and one for py3)


gornostal:
Thanks for the PR @goodwillcoding 

Is it an updated #279 or you've made all migration updates yourself?

> whether there will be 2 packages now (one for py2 and one for py3)

I would not go that path. As soon asl py3 version is stable I'm going to release it in the Dev channel. And then a 2-4 month later will drop support for py2 and merge the code to master

gornostal:
@goodwillcoding before you've created this MR I've tried to pull and run code from #279 (Thanks @cjbassi for the contribution). It works, but lacks some bug fixes that were made to the develop branch after the MR.

I've created `python3` branch where I pushed my changes. I'm going to cherry pick some stuff from your branch, from #279 and `dev` branch in the next week. Hopefully will get tests passing and dependencies updated so I can test the app in different distros.

I see you maintained compatibility with python 2, which is cool, but I don't see why we may want to release and maintain py2+py3 compatible version. Other than to let the users choose which python they want to use with Ulauncher. Am I missing something?

Thanks

goodwillcoding:
@gornostal maintaining both was the take away I got from the other (contributors) discussion. I'll read it over again and see if I got confused

goodwillcoding:
@gornostal ... okie, I re-read the discussion and I got confused. Specifically I the reason for keeping both is to let the extension maintainer time to adapt (i.e. release the py3 packages side by side or something).

The changes were from scratch, thought I did look over cjbassi contributions. Since you got the branch now, just assign things to be you need done.

gornostal:
My idea was to make a dev release with py3 and drop py2 completely in a new version. But keep py3 version in dev for a while until enough extension developers migrate to the python 3.

If we keep Ulauncher compatible with both py3 and 2, will it help users or extension developers in any way?

goodwillcoding:
@gornostal ... my general experience with "3rd party plugin" bits (firefox add-ons, gnome shell extensions) was that the least stressfull path is to have a grace period of supporting both. 

Specifically, what could be done is to suggest to user when they launch py2 version that they switch to py3 (refork the processes) ... and report any plugins that do not work somewhere (maybe a google form).

Of course that's a bit more work so maybe we should not do it.

gornostal:
I gave it another thought and decided that the gradual period when both versions are supported is not worth it -- too time consuming to do it right.

In order to have gradual period with py 2 and 3 supported the following must be done:
* when packaging Ulauncher, include dependencies for py2 and 3
* install Ulauncher into both `/usr/lib/python2.7` and `/usr/lib/python3`
* check if extension supports py3, and then run the extension with py3. Run it with py2 otherwise
* 2 times more testing

gornostal:
Migration to Python 3 is now merged to dev. Closing this PR

PR #317, opened 2018-10-23
Swapped out fuzzy match code

raddessi:
Moved over to the fuzzywuzzy module for matching and match indexing

### DIFF
--- a/ulauncher/search/SortedList.py
+++ b/ulauncher/search/SortedList.py
@@ -1,5 +1,5 @@
 from ulauncher.util.SortedCollection import SortedCollection
-from ulauncher.util.fuzzy_search import get_score
+from fuzzywuzzy import fuzz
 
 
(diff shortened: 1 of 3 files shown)


raddessi:
This does seem to have some odd behavior also, I'm sure I'm not familiar enough with this codebase yet. When typing in "chr" (expecting to match "Google Chrome") chrome is in fact selected as the highlighted entry but it is not at number 1, instead for me it is at number 5. I'll dig when I have time, unless you can shed some light on what is going on.

gornostal:
@raddessi 

> it is not at number 1, instead for me it is at number 5. 

That's probably because of another feature of Ulauncher :)

It remembers which apps you selected/opened before for a certain query.
Then if you enter the same query again it will select the app that you ran previously. Even if it's not the item no.1

Issue #355, opened 2019-03-14
Open file using file browser in background


1earch:
### Steps to reproduce
1. Search for a file to open using the `~` shortcut
2. Press `<Enter>` to open the file

### Current behavior
The file is opened, but Ulauncher is not hidden until file is closed.

### Expected behavior
The file should be opened in background and Ulauncher should be hidden.

### Environment
**Ulauncher version**: 4.3.2.r8
**Linux distro**: Arch Linux
**Desktop environment**: i3wm

gornostal:
Implemented in 5.3.0

PR #356, opened 2019-03-14
Open file in background when using file browser

1earch:
Fixes issue #355 

Change from `subprocess.call` to `subprocess.Popen`:

- `call` waits for the child process to end
- `Popen` runs the child process in background

### Tested environment
ArchLinux, i3wm 4.46.1


gornostal:
Thanks for the fix @1earch

Issue #347, opened 2019-02-12
Store themes and .db files in ~/.local/share/ulauncher


gornostal:
See https://github.com/Ulauncher/Ulauncher/issues/152#issuecomment-462654460

Basically, deleting the cache dir should not change Ulauncher behavior. Not it changes it because the .db files with user patterns are stored in there.

kisaragi-hiu:
IMHO, extensions should not be stored in `~/.cache` either.

Browsers treat downloaded websites as cached data because it is unfeasible to store all websites the user visits on disk: it'd take up too much space, and much of it will rapidly become outdated. The download bandwidth saved isn't worth the space required.

Ulauncher extensions are not like that: the saved download bandwidth *is* worth the space required. They are not updated rapidly (more bandwidth saved), and when added up they don't take up as much space as websites (not much space is required). This makes it reasonable to store them as persistent data, not cached.

gornostal:
Implemented in 5.3.0

Issue #152, opened 2017-08-17
Non-standard cache path


hickscorp:
In my `~/.cache` directory, two folders are being created by ULauncher, one not following the standard convention:

- `~/.cache/ulauncher` which is correct, and probably created by the Webkit internal based on the app id.
- `~/.cache/ulauncher_cache` which has that redundant `_cache` suffix.

It would be nice to fix that to have a more standard cache directory structure.

The responsible line is in the [confg constants](https://github.com/Ulauncher/Ulauncher/blob/86939732c978a97a4941bdd2d3f6848277b1e275/ulauncher/config.py#L20) but changing that would be a mistake: a migration would have to be written to properly move existing cache files and then destroy the old `ulauncher_cache` folder *if empty*.

gornostal:
There is a bug in WebKit2 that removes cache folder if it has the same name as the executable. Hence `~/.cache/ulauncher_cache`.
It's actually mentioned in that file you linked

branb:
I'm not sure if ```~/.cache``` is the proper base directory for *everything* that is currently inside of it. As stated in the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

> ```$XDG_CACHE_HOME``` defines the base directory relative to which user specific **non-essential data files** should be stored. If ```$XDG_CACHE_HOME``` is either not set or empty, a default equal to ```$HOME/.cache``` should be used.

At the moment, extensions and themes are saved there as well, and I'm not too sure what's inside of the ```.db``` files. Ulauncher and all it's extensions should be able to work as expected when you delete the cache folder at any point of time during it's runtime. The first sentence from the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#varcacheApplicationCacheData) explains it best:

> ```/var/cache``` is intended for cached data from applications. Such data is locally generated as a result of time-consuming I/O or calculation. The application must be able to regenerate or restore the data. Unlike ```/var/spool```, the cached files can be deleted without data loss. The data must remain valid between invocations of the application and rebooting the system.

I think at least the extensions, themes and maybe some - if not all - of the ```.db``` files should be stored into ```~/.local/share/ulauncher/``` instead. This way the WebKit-Bug would be completely irrelevant as well, because even though it may be annoying, anything stored in ```~/.cache``` should not cause any issues when deleted no matter if by bug, intentionally or whatever.

[...]

PR #417, opened 2019-08-17
Treat some data as persistent and save them in XDG_DATA_HOME

kisaragi-hiu:

#152, #347

Implement #347.

> A cache [...] stores data so future requests for that data can be served faster [...]. — https://en.wikipedia.org/wiki/Cache_(computing)

These existing uses of `CACHE_DIR` record history, so they should not be considered cache:

- QueryHistoryDb
- AppStatDb
- Logging

I've also made it so that extensions are saved in `DATA_DIR` (the new variable pointing to `xdg-data-home`). While extensions can be "recomputed" (i.e. downloaded again), it takes a relatively long time to download them, and not too much space to store them persistently. This is in fact what bit me initially: I made my system clear the cache on shut down because I don't care about a cache from the previous session, but with Ulauncher extensions being treated as cache I'd have to wait an unpredictable amount of time after login before they can be used again.

- [X] Write unit tests for your changes when applicable
- [x] If your changes alter the behavior or introduce new functionality, please update the documentation accordingly
- [X] All tests are passing

- Arch Linux, Python 3.7.4, KDE Plasma 5.16.4

### DIFF
--- a/docs/extensions/tutorial.rst
+++ b/docs/extensions/tutorial.rst
@@ -4,7 +4,7 @@ Development Tutorial
 Creating a Project
 ------------------
 
-Ulauncher runs all extensions from ``~/.cache/ulauncher_cache/extensions/``.
+Ulauncher runs all extensions from ``~/.local/share/ulauncher/extensions/``.
 
 Create a new directory there (name it as you wish) with the following structure::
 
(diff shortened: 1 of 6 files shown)


gornostal:
Thanks a lot for the MR @kisaragi-hiu :+1:

Issue #349, opened 2019-02-20
Window is positioned in top-left corner in Fluxbox


goodevilgenius:
**Steps to reproduce:**
1. Launch ULauncher
2. Look for window in center of screen
3. Find it in top-left corner instead

**Ulauncher version**: 4.3.1.r4-1
**Linux distro**: Fedora 29
**Desktop environment**: Fluxbox, with dual monitors

**Logs**:

https://gist.github.com/goodevilgenius/c841b75ffe6b5efb0f333b4dd1f47627

Issue #205, opened 2018-01-24
Window position


PierfrancescoArdino:
![screenshot from 2018-01-24 09-02-37](https://user-images.githubusercontent.com/9986725/35320856-9365a3b6-00e5-11e8-8aa7-efbc3d3a4b5a.png)

The position of the window seems wrong, moreover to write a command on the input field I've to first click on it.

I'm not 100% but it seems to happen after launching a second command. Before then, the widget seems to work as expected. 

**Ulauncher version**: Ulauncher-git latest
**Linux distro**: Arch Linux
**Desktop environment**: Gnome on Wayland

gornostal:
could you attach logs please?

gornostal:
I've seen this issue only when `GDK_BACKEND` is set to `wayland` explicitly.

PierfrancescoArdino:
> 2018-01-24 09:19:55,863 | DEBUG | ulauncher.util.display: get_screens() | Found 1 monitor(s).
2018-01-24 09:19:55,863 | DEBUG | ulauncher.util.display: get_screens() |   Monitor 0 - X: 0, Y: 0, W: 1920, H: 1080
2018-01-24 09:19:55,864 | DEBUG | ulauncher.ui.windows.UlauncherWindow: show_results() | render 0 results
2018-01-24 09:20:02,366 | DEBUG | ulauncher.util.display: get_screens() | Found 1 monitor(s).
2018-01-24 09:20:02,367 | DEBUG | ulauncher.util.display: get_screens() |   Monitor 0 - X: 0, Y: 0, W: 1920, H: 1080
2018-01-24 09:20:02,367 | DEBUG | ulauncher.ui.windows.UlauncherWindow: show_results() | render 0 results
2018-01-24 09:20:04,681 | DEBUG | ulauncher.util.display: get_screens() | Found 1 monitor(s).
2018-01-24 09:20:04,682 | DEBUG | ulauncher.util.display: get_screens() |   Monitor 0 - X: 0, Y: 0, W: 1920, H: 1080
2018-01-24 09:20:04,682 | DEBUG | ulauncher.ui.windows.UlauncherWindow: show_results() | render 0 results
2018-01-24 09:20:05,815 | DEBUG | ulauncher.util.display: get_screens() | Found 1 monitor(s).
2018-01-24 09:20:05,815 | DEBUG | ulauncher.util.display: get_screens() |   Monitor 0 - X: 0, Y: 0, W: 1920, H: 1080
[...]

PR #419, opened 2019-08-23
Make window visible before requesting move.

subdavis:

#349 #205

On LXDE Ubuntu 18.04 with i3-gaps, I was consistently having an issue with the window jumping all over the place.   The location would be correct on a fresh launch, then wrong every time after that.

I traced the issue to [the Python gtk.Window docs](https://developer.gnome.org/pygtk/stable/class-gtkwindow.html#move)

> The move() method asks the window manager to move the window to the position specified by x and y. Window managers are free to ignore this. In fact, most window managers ignore requests for initial window positions (instead using a user-defined placement algorithm) and honor requests after the window has already been shown.

As far as I can tell, my window manager (and most of them) are ignoring the request for initial position before the window is presented. I made this change and the launcher shows up at the right location every time now.

- [ ] Write unit tests for your changes when applicable (NA)
- [ ] If your changes alters the behavior or introduce new functionality, please update the documentation accordingly (NA)
- [x] All tests are passing

* LXDE
* Ubuntu 18.04
* i3-gaps

### DIFF
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -247,9 +247,9 @@ class UlauncherWindow(Gtk.Window, WindowHelper):
 
     def show_window(self):
         # works only when the following methods are called in that exact order
-        self.position_window()
         self.window.set_sensitive(True)
         self.window.present()
+        self.position_window()
         if not is_wayland_compatibility_on():
             self.present_with_time(Keybinder.get_current_event_time())
 


subdavis:
@gornostal any thoughts on this?

gornostal:
@subdavis thanks for the PR. I haven't seen this issue, but I'll the PR anyway.

But if it breaks something on other distros I will have to revert it

gornostal:
The fix released in Ulauncher v5.3.0~beta2

PR #457, opened 2019-10-29
Fixed PyGIWarning: Gtk was imported without specifying a version first

voneAbides:
Fixed warning:
PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.


friday:
#458 is a duplicate of this which also follows the python style guidelines and hence passes the tests.

Issue #638, opened 2020-11-14
Would you consider merging support for podman?


troycurtisjr:
I am running on Fedora which uses podman as the container engine instead of docker. In fact, some hoops need to be jumped through in order to run docker correctly considering its lack of support for cgroups2 (or at least so I've read). With very minimal tweaks I have managed to get the development build container to run correctly in podman. If I put this in a PR would you be opposed to merging it?

### Describe the solution you'd like
Allowing the build scripts to use docker or podman equally well.

gornostal:
Merged the PR. Closing this issue now.

Issue #337, opened 2019-01-22
Ulauncher window closes unexpectedly during query


jbriales:
**Steps to reproduce:**
IF upon triggering Ulauncher the mouse pointer is over the Ulauncher floating window
AND after entering some query the window size changes (because of a smaller number of items)
AND the mouse pointer falls out of the Ulauncher floating window area,
THEN the Ulauncher window will close "unexpectedly".

Does this make sense? If some of the missing information below is important I can reproduce it later (I'm writing this on a rush).

**Ulauncher version**: TODO
**Linux distro**: Fedora28
**Desktop environment**: i3-gnome

---

**For contributors**

Introduce a new CLI option `--keep-on-focus-out`, which disables keeps the ulauncher window visible even if it receives focus out event.

gornostal:
Hi @jbriales 

There's `focus_follows_mouse` setting in i3, which by default is set to `yes`.
You may want to set it to `no` to fix the issue.
More info here https://i3wm.org/docs/userguide.html#_focus_follows_mouse

Hope that helps. Please reopen the issue if needed.

troycurtisjr:
Are you saying that you'd prefer to keep focus follows mouse as incompatible with ulauncher? 

This setting is one of the biggest things I miss when I have to work on Mac and Windows. Would you consider an option to not close when ulauncher loses focus? I think there are several explicit actions which can cause ulauncher to hide already: hitting Enter to select an option, Esc to cancel, etc.  

Such an option would allow ulauncher to work with focus follows mouse (and sloppy focus as mentioned in issue #58 ), and would also allow for the use case described in #459 .  Additionally I suspect it would address #242.

Finally, this is not an unusual setting, and as mentioned in this Issue, it is the default for i3 with is definitely a power user WM (which I would argue is a similar demographic as ulauncher itself).

troycurtisjr:
Additionally it might address #364 as well.

ToxaZ:
@troycurtisjr sure. Especially considering that my wrapper no longer working for new version. I haven't done much research on that, but now the code I wrote is useless.

gornostal:
@troycurtisjr @ToxaZ I've reopened this one.

Added details to the description about a new CLI option that prevents window from hiding.

gornostal:
Just released v5.9.0 that has a new setting:
![image](https://user-images.githubusercontent.com/1202543/99916045-4547eb80-2d10-11eb-8509-5cf4eed798c5.png)

[...]

PR #640, opened 2020-11-21
Add option to not hide when losing focus due to the mouse

troycurtisjr:

Addresses #337.
Also address #638.

This add a user preference option which will prevent ULauncher from closing when focus is lost because of the mouse. This makes ULauncher compatible with focus follows mouse and sloppy focus modes.

The issue talked about making this option as a CLI command. However, that makes it more challenging to set the preference for the user if they are using a system installed ULauncher. In that case they would have to modify the desktop file to add the shortcut, and the modification would likely get overridden on the next update. I was able to make this a user preference so that the customization is stored in the preferences file.

I tried to make the UI element short and meaningful about what it did and not how it was implemented. Internally the option is grab-mouse-pointer since that is the actual implementation, but the effect of grabbing the mouse pointer is probably not obvious to the general user. I am most certainly interested in feedback on the naming.

I also included a couple of small tweaks which allowed me to use the container infrastructure with podman (via podman's docker compatibility wrapper). I opened issue #638 for this and I can split it out into a separate PR if desired, or remove it entirely if you aren't interested in supporting it.

Lastly I couldn't find any documentation to update with the new option. I'd be happy to do so if you give me a pointer of where to look.

- [x] All tests are passing

Distro: Fedora 32
Desktop: Gnome 3.36

### DIFF
--- a/data/preferences/src/api/fixture.js
+++ b/data/preferences/src/api/fixture.js
@@ -14,6 +14,7 @@ export default function(url, params) {
         autostart_allowed: true,
         autostart_enabled: true,
         clear_previous_text: true,
+        grab_mouse_pointer: true,
         blacklisted_desktop_dirs: ['/var/tmp', '/tmp/var/log/bin/bash/root'].join(':'),
         available_themes: [{ text: 'Dark', value: 'dark' }, { text: 'Light', value: 'light' }],
         theme_name: 'light',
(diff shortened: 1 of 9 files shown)


gornostal:
The code changes are simple and clear. I like that :+1: 
Thanks a lot for this PR @troycurtisjr!

sliddjur:
I use Gnome 3.36.8, x11 and  ulauncher 5.
I have multiple language inputs. and I want to change language once I have opened the uLauncher window.
I enabled "dont hide after losing mouse focus". But I still cannot click anything outside the ulauncher window. 
My usecase would be to be able to switch language input to use with the translate-shell plugin.

friday:
@troycurtisjr I always forget what this option does, and it looks like another option I want to add so I'm renaming/clarifying it. Let me know if this isn't explaining it well/correct. I never use these focus modes and don't know anything about them. 
image

PR #643, opened 2020-11-23
Try fixing the ulauncher-toggle crash

Surendrajat:

Just the path of `sh` in shebang of `ulauncher-toggle` is changed

- [ ] Follow the [Python Code Style Guides](https://github.com/Ulauncher/Ulauncher/wiki/Python-Code-Style-Guides)
- [ ] All tests are passing

### Tested environment (distro, desktop environment, and their versions)
Distro: elementray OS 5.1.7 (Ubuntu 18.04)
DE: pantheon

### DIFF
--- a/bin/ulauncher-toggle
+++ b/bin/ulauncher-toggle
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/usr/bin/env sh
 
 dbus-send \
     --session \


bin/ulauncher-toggle
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/usr/bin/env sh
```
gornostal, line 1: Unfortunately this breaks the build for Fedora. I've already tried this.
I'll think what we can do here

friday:
We deactivated the fedora builds and fixed this, but used /usr/bin/env bash rather than sh, since `sh` runs in the users default shell, which might not be fully compatible with the shell script.

It's released in Ulauncher 5.12.0

Issue #270, opened 2018-08-20
Feature: Support keywords in search


ivashin:
Many applications provide additional keywords in their .desktop files to facilitate finding them. E.g. if you type "Excel" into Unity Dash it shows LibreOffice Calc because libreoffice-calc.desktop contains "Microsoft Excel" in Keywords section.

It would be very useful to parse these keywords and use them in search similarly. https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys contains specification details.

It also makes to include comment from desktop entry in search as well, albeit perhaps with a lower priority.

friday:
This one is important to me as well. Part of what I expect the launcher to handle is to find the apps I rarely use and don't remember the name of. For example I search for "usb" not specifically the name "Etcher" or "UNetbootin" since I use these apps too rarely to remember.

gornostal:
Yes, that sounds like a nice feature to me too.

friday:
Comment aka "Description" (it's the same thing) is now supported in the Ulauncher v6 branch, so I'm renaming this to remove that. Keywords work a bit different from this and need to be matched on a word by word basis.

friday:
I should clarify why this isn't something I am looking into implementing, or at least not soon:

Implementing this would make searching more complex, slower (maybe not noticeably so though) and give you false positives you didn't intend to match based on keywords.

Many of these apps have a lot of keywords. Gnome Online Accounts (settings panel) for example has these: `Google;Facebook;Twitter;Yahoo;Web;Online;Chat;Calendar;Mail;Contact;ownCloud;Kerberos;IMAP;SMTP;Pocket;ReadItLater`
So if you search for Chat or Mail or Calendar or "Pocket" you match that. In some cases that will be helpful. In some cases not.

Gnome lock panel has these: `screen;lock;diagnostics;crash;private;recent;temporary;tmp;index;name;network;identity;privacy`

XFCE WMTweaks has these: `windows;behaviour;settings;preferences;cycling;cycle;switching;focus;raises;accessibility;key;dragging;move;moving;hide;frame;title bar;maximized;tile;screen;edge;hot corner;snapping;mouse wheel;roll up;workspaces;placement;compositor;compositing;enable;disable;shadows;decorations;opacity;resize;inactive;preview;compiz;transitions;`

[...]

PR #652, opened 2020-12-01
Search parameters include application description

### LINKED ISSUE #613: Can I recognize and search even "comment" in ~/.local/share/applications/*.Desktop File?
## Question
Can I recognize and search even "comment" in ~/.local/share/applications/*.Desktop File?
Is there a way? Is it possible on Ulauncher?

[Desktop Entry]
Version=1.0
Name=유튜브
Comment=**Youtube (Ice SSB Webapp)**

BharatKalluri:

Partly solves #270 . Also closes #613 

Currently the search uses the name, although description is present in the AppDB. With this change we also consider the description in the search. 
If the match score of description is greater than min_score (which is 50 for app search) then the score is halved and sent back as a result item, just to make sure it always ranks below entries which match with the name.
One thing I would like your opinion on is that - now the get_score fn returns ResultItems which have a score less than the min_score provided. This seems weird to me, but I could not think of any other elegant way. Let me know your thoughts :smile: 

- [x] All tests are passing

distro: Fedora 33
desktop environment: GNOME on XOrg
version: running from the dev branch, version

### DIFF
--- a/tests/search/test_SortedList.py
+++ b/tests/search/test_SortedList.py
@@ -75,15 +75,3 @@ class TestSortedList:
         assert ri6 in res_list
 
         assert res_list[0] == ri5  # ri5 stays first, because it has the highest score (100)
-
-
-class TestSortedListOrder:
-    def test_order_if_name_and_description_match(self):
-        r1 = ResultItem(name='Problem Reporting', description='An application to report problems')
-        r2 = ResultItem(name='VS Code', description='Code editing, redefined')
-
-        sorted_list_instance = SortedList(query='re', min_score=50)
-        sorted_list_instance.append(r1)
-        sorted_list_instance.append(r2)
-        assert sorted_list_instance[0].get_name() == 'Problem Reporting'
-        assert sorted_list_instance[1].get_name() == 'VS Code'
--- a/tests/search/test_SortedList.py
+++ b/tests/search/test_SortedList.py
@@ -75,3 +75,15 @@ class TestSortedList:
         assert ri6 in res_list
 
         assert res_list[0] == ri5  # ri5 stays first, because it has the highest score (100)
+
+
+class TestSortedListOrder:
+    def test_order_if_name_and_description_match(self):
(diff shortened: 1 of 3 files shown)


troycurtisjr:
@BharatKalluri Are you still interested in getting this merged? There are conflicts with the current code, but more importantly for the upcoming [v6](https://github.com/Ulauncher/Ulauncher/tree/v6) there are more substantial changes to the fuzzy search ( and much improved!) that may be applicable. So if you are still interested in getting this integrated, it may be more worthwhile to target the v6 branch.

friday:
It's even easier to implement now. I've been reluctant to do it since it could have a confusing or negative effect on the results, and it's hard to know how to weight it.

BharatKalluri:
Hey, sorry for the late response!
I agree with @friday , let me know how we want to tackle this problem. I could make a separate PR for v6 based on the decisions we take.

PS: Super excited about the v6 release! 🥳  I'll probably dive into issues and see if I can make any contributions!

friday:
> Hey, sorry for the late response! I agree with @friday , let me know how we want to tackle this problem. I could make a separate PR for v6 based on the decisions we take.
> 
> PS: Super excited about the v6 release! partying_face I'll probably dive into issues and see if I can make any contributions!

That's great!
I'm focusing on the architectural/API changes and refactoring myself since those require more focus and in-depth knowledge of the codebase. Some of them so far have been quite big and it makes it a bit difficult to coordinate work perhaps, but smaller additions like this should be pretty easy.

The V6 milestone has a list of issues, most of which are pretty complex, because they are things I picked as "this will likely break compatibility, so better do it in one go". You can also see in the milestone list at the bottom (you need to expand it) that there are some things that are broken now.

You should look at how `self._executable` is added and handled here and add more meta (like description and GenericName), but I suggest to wait with keywords, as those work differently and add more problems.

https://github.com/Ulauncher/Ulauncher/blob/v6/ulauncher/modes/apps/AppResult.py#L78-L81

Adding this is super easy, it's just the weighting and testing part that prevented me, because it's important that we don't get false positives.

Issue #382, opened 2019-05-25
Programs requiring pkexec don't launch


LinuxOnTheDesktop:
I thought this was reported already - at least for Mint's program `mintsources' - but I cannot find that report.

### Steps to reproduce
1. Be on Mint (or some other OS?)
2. Try to run a program that launches via `pkexec`, such as bleachbit or minsources. (Bleachbit is not installed by default in Mint. Also, Bleachbit creates system menu items to run the program as root and as a normal user, respectively. Accordingly, Ulauncher shows two entries for Bleachbit; and only the entry for running as root fails to work.)

### Current behavior

Nothing happens, i.e. nothing is launcher.

### Expected behavior

An authentication window should appear and then, if the user authenticates successfully, the program in question should run.

### Environment
**Ulauncher version**: 4.4.0.r1
**Linux distro**: Mint 19.1
**Desktop environment**: Cinnamon

### Logs

<script src="https://gist.github.com/LinuxOnTheDesktop/b65348955b84303b87976373731997cc.js"></script>

friday:
That's not a log file. Also, what does the `Exec` line say in the desktop entry for Bleachbit? It's probably named `bleachbit.desktop` and located in `/usr/share/applications/`

LinuxOnTheDesktop:
Log file: I was trying to provide the terminal output that your issue template requests. I have done so again [here](https://gist.github.com/LinuxOnTheDesktop/ebe9ac4d521e75f6d69e1a14eeb8a5b3) and (in case there is any problem with that link, which is to a 'gist'), [here](https://pastebin.com/bTjL90xh).

> what does the Exec line say in the desktop entry for Bleachbit?

Relevant parts of `bleachbit.desktop`:

```
TryExec=bleachbit
Exec=bleachbit
Icon=bleachbit
Categories=GTK;System;
Keywords=clean;performances;free;privacy;
StartupNotify=true
```

Relevant parts of the more relevant file, viz., `bleachbit-root.desktop`:

```
TryExec=pkexec
Exec=pkexec bleachbit
Icon=bleachbit
Categories=GTK;System;
Keywords=clean;performances;free;privacy;
StartupNotify=true
```

friday:
That log file gave it away. Seems to be this issue: https://askubuntu.com/questions/608419/trying-to-execute-a-process-with-pkexec-asynchronously-and-get-refusing-to-rend

[...]

PR #728, opened 2021-04-14
Naive solution to pkexec issues (for Python 3.8 or higher)

WhosyVox:
It's not a great solution but here's a jumping off point to support pkexec in Ulauncher.

Part of the fix requires a Pexec flag only present in Python 3.8 or later, though this is only to prevent the application from closing when Ulauncher's main process does. As far as I know this won't stop the application or fix working in older versions.

Sadly I couldn't find a solution to prevent redirecting part of the application std output to the terminal, but if anyone can modify this then please go ahead.

This was tested as working under Linux Mint 20.1 using driver-manager and timeshift.

### DIFF
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -1,6 +1,7 @@
 import logging
 import pipes
 import subprocess
+import shlex
 
 from ulauncher.utils.desktop.reader import read_desktop_file
 from ulauncher.utils.Settings import Settings
@@ -26,6 +27,7 @@ class LaunchAppAction(BaseAction):
     def run(self):
         app = read_desktop_file(self.filename)
         command = app.get_string('Exec')
+        
         terminal_exec = settings.get_property('terminal-command')
         if app.get_boolean('Terminal') and terminal_exec and command:
             logger.info('Run command %s (%s) in preferred terminal (%s)', command, self.filename, terminal_exec)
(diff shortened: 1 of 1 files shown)


ulauncher/api/shared/action/LaunchAppAction.py
```diff
@@ -26,13 +27,22 @@ def keep_app_open(self):
...
         if app.get_boolean('Terminal') and terminal_exec and command:
             logger.info('Run command %s (%s) in preferred terminal (%s)', command, self.filename, terminal_exec)
             subprocess.Popen(terminal_exec.split() + [pipes.quote(command)])
         else:
             logger.info('Run application %s (%s)', app.get_name(), self.filename)
             try:
-                app.launch()
+                if (command.startswith("pkexec")):
+                    subprocess.Popen(
+                        args=shlex.split('sh -c "'+command+'"'),
```
friday, line 40: `args=shlex.split('sh -c "'+command+'"'),` could be changed to `args=['sh', '-c', command],` couldn't it? Less code and you wouldn't need to import shlex.

friday:
This is for the #382 issue

ulauncher/api/shared/action/LaunchAppAction.py
WhosyVox: I imagine if you're suggesting so, that's probably true! I'm inexperienced with Python and tried to hack code together until it worked. My hope was that someone might see this, tidy it up/fix it a little and resolve the initial problem. ULauncher could completely replace the mint menu if this ever gets fixed.

friday: I haven't even tried the PR, but it should do exactly the same thing with these line swapped. I'm not using apps that need root permissions a lot.

friday:
Fixed in a very similar way in #778.
Thanks again for the PR and testing! Sorry you had to wait so long.

Issue #559, opened 2020-05-13
minimum keystrokes


tulsidas:
## Question
Is there a way to configure the minimum keystrokes to open an app?

For example, I want to open **KeePassXC**
I open Ulauncher and I have to type `keepa` in order for it to show me the application which I can choose to start.
I don't have any other applications starting with `kee` so shouldn't it with those 3 chars already show me KeePassXC? Is there a way to have a *cache* of most used apps so I can start KeePassXC with only `k`, since I open it daily?

Thanks!

mahmadmujtaba:
Open Ulauncher app and go to shortcuts tab to define a shortcut for keepass xc eg kp.

KuenzelIT:
@tulsidas  When I open ULauncher and type `ke` I get KeePassXC directly as the first result. Maybe the latest version already fixed this?

tulsidas:
Tried it with 5.8.0 and works fine, thanks!

Issue #542, opened 2020-04-13
unable to find apps by just input 2 or 3 letters


mchilli:
Since yesterday or version 5.7.3, maybe 5.7.2 also, I could'nt find i.e. VLC Player by only typing "vl", i only find it if I type "vlc media p". Maybe ulauncher searches first in the description since the new version?
In previous versions <= 5.7.1 I find the app mostly by typing only one letter!

Update:
Ok, when I change [this line](https://github.com/Ulauncher/Ulauncher/blob/ee9e9da4a202fe88b9c309d453821ae7f6ae6389/ulauncher/search/apps/AppDb.py#L188) back to the previous one `return '%s %s' % (name, exec_name)`, ulauncher work fine again!

### Environment
**Ulauncher version**: 5.7.3
**Linux distro**: Linux Mint 19.3
**Desktop environment**: Cinnamon 4.4

lgierth:
Same for me, it looks like #516 not just added `description` to the search, but also removed `exec_name` :)

lgierth:
Another issue is that I can now type "LibreOffice Calc" and it won't show up, but type just "LibreOffice" and it will

gornostal:
> Same for me, it looks like #516 not just added description to the search, but also removed exec_name

@lgierth Indeed. I'm going to create a fixed version soon

gornostal:
After taking another look the `exec_name` is still used in search, but I'm going to revert that change anyway.

gornostal:
Should be fixed in 5.7.4
Please reopen if needed

Pencil4:
**Thank you** for fixing this so promptly!!

lgierth:
Thank you :)

Issue #336, opened 2019-01-20
Sorting issue - existing command should be the first one (fuzzy search improvements)


sukovec:
Hello. I installed ulauncher yesterday (came from gmrun). So far so good. But today I wanted to run gimp, so I inserted "gimp" into line and pressed enter... and gvim has been launched. [Can be seen on image](https://sh.sukovec.cz/upload/gimp-gvim.png)

**Steps to reproduce:**
- Install gvim & gimp
- Insert "gimp" into ulauncher

**Ulauncher version**: 4.3.1.r4
**Linux distro**: Arch, 4.20.1-zen1-1-zen x86_64
**Desktop environment**: i3wm

sukovec:
Okay, it looks like I mis-interpreted purpose of this software - I was looking on it as "cool and good looking command launcher" but it's probably more like program launcher?

gornostal:
> it's probably more like program launcher

Yes, that's right.

gornostal:
The issue is related to #232. 
Will be fixed along with it.

sukovec:
One time, it happened to me that I wrote something too fast after launch and expected program was not in the list (re-writing it after a while reveals that finally). Is it also related with #232 or should I raise new bug? I cannot see anything more related...


> > it's probably more like program launcher
> Yes, that's right.
So, I cannot expect some support to just exec plain commands? Sometimes I just need to run simple command (like `i3 restart`, `reboot` or `shutdown`) - for which is small launcher window more suitable (as I don't need to create new terminal windows in tiling layout).

gornostal:
> ... wrote something too fast after launch and expected program was not in the list (re-writing it after a while reveals that finally)

That can happen when you start Ulauncher for the first time and start typing a query while it's indexing apps in the background. It's not a bug if that's your case.

> Sometimes I just need to run simple command (like i3 restart, reboot or shutdown) 

You can do that with Ulauncher. There are several ways.
First, you can check out if there's an extension for that here https://ext.ulauncher.io/.
I saw several extensions that provide commands for shutdown, reboot, log out.
If that doesn't suit you, you can open Preferences, go to Shortcuts and create a shortcut that will run a shell command. You will still have to use some prefix like `cmd i3 restart`, because by default Ulauncher query is used to search apps

friday:
This may also be of interest @sukovec: https://github.com/Ulauncher/Ulauncher/issues/191#issuecomment-457000841

friday:
[...]

Issue #324, opened 2018-12-08
About Ulauncher .desktop search (fuzzy search improvements)


jbriales:
I have created my own .desktop files (in /home/jesusbriales/.local/share/applications/) for some of my command-line applications.
One example is `web-whatsapp`.
When I open Ulauncher and intend to launch this app, writing 'web-w' will already prompt `web-whatsapp` among options to run.
However, the output for other search inputs are a little unexpected to me:
- 'what' -> prompts `Atom`
- 'whats' -> prompts `Photos`, `Tweaks` and `web-thesaurus`
- 'whatsa' -> prompts `web-thesaurus`, `Photos` and `Tweaks`
- 'whatsap' -> prompts `web-thesaurus`, `web-whatsapp` and `web-maps`
- 'whatsapp' -> finally prompts `web-whatsapp` and others after it
So, I know Ulauncher uses some kind of fuzzy search, although I'm not familiar with its details yet.
I was wondering though if the behavior above is right/expectable, as just writing 'what' should be (I think) much closer to `web-whatsapp` than to any other of the options listed by Ulauncher.

friday:
The current behavior is not optimal, but finding what is optimal isn't easy.

There's both matching and scoring/sorting involved. Matching uses the binary name too, so you can type "nautilus" or "eog" to get the file manager and image viewer for Gnome although they've both given different names in the desktop entries. There's also some feature to sort and highlight the most apps you open the most,

If you get these strange matches only for desktop entries you just created, you could try to narrow it down by tweaking them. You may also want to try to close ulauncher and open `ulauncher -v` from a terminal so you see when it detects changes and updates or if it fails to do so.

I'm not sure the behavior is safe if you copy a desktop entry for one app, rename it and change the file after it has been indexed (it seems like the kind of thing that could cause something like this). If restarting Ulauncher doesn't do it, I would try closing Ulauncher and deleting the db files in `~/.cache/ulauncher_cache/` just to be sure. Note that this is based entirely on hunches. I don't know the indexing and db code well.

Btw, you know you can create shortcuts for searching websites, right? It looks a bit like you're doing something similar with `web-thesaurus` and `web-maps`. Shortcuts would avoid one page load and take you directly to the search results.

jbriales:
Hi @friday, thanks for the hints! I'll have a look at it and come back if I get any useful insight.
Also, what do you mean by creating shortcuts for searching websites? That might interest me.

friday:
[...]

PR #742, opened 2021-05-20
Sort results by frequency

voidstarstar:

Possibly related to #559, #324, #336, #232, #317, #542

This will sort the fuzzy search results first by their frequency score and then by their fuzzy search score.  Previously they were sorted only by their fuzzy search score.  The frequency score is the number of times the user has selected the app in the past using the `AppStatDb`.

This should have the (hopefully desired) effect of putting the most frequently used apps on the top of the list and thereby requiring fewer keystrokes to open.

Note: This does not affect the quality of the fuzzy search results, only their order.  However, since this sorting is done over all results, rather than just the final 9 that are printed, the actual results shown to the user will differ and prefer frequent apps.

I believe this PR is orthogonal to an improved fuzzy search (e.g. #317) and would not cause issue with a swapped out algorithm.  I have some thoughts for an improved algorithm, but that would be for a separate issue/PR.

This PR should not be merged as is since it lacks updated tests.  I wanted to open the PR to at least verify if this is a welcomed feature.  It may need to hide behind a preference flag too.

- [X] Use `dev` as the base branch
- [X] Follow the [Python Code Style Guides](https://github.com/Ulauncher/Ulauncher/wiki/Python-Code-Style-Guides)
- [ ] All tests are passing

Ubuntu 20.04, Gnome Flashback

### DIFF
--- a/ulauncher/search/SortedList.py
+++ b/ulauncher/search/SortedList.py
@@ -1,5 +1,7 @@
 from ulauncher.utils.SortedCollection import SortedCollection
 from ulauncher.utils.fuzzy_search import get_score
+from ulauncher.search.apps.AppStatDb import AppStatDb
+from ulauncher.search.apps.AppResultItem import AppResultItem
 
 
 class SortedList:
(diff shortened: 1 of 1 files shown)


voidstarstar:
Is the highlighted result always the one that was most frequently used?  If so, then this PR should cause that to always be at the top of the list, which I think is more intuitive.

troycurtisjr:
Wouldn't this approach cause completely unrelated apps to what the user is typing to pop up, merely because it is frequent? Shouldn't frequency be used to break a tie or boost, but not override, the match score?

friday:
Superseded by #1175

PR #768, opened 2021-08-09
Move conftest.py to the project root

friday:
Currently pytest needs to run one of these ways:
1. `python -m pytest`
2. `PYTHONPATH=`pwd` py.test tests`
3. [`test-pytest`](https://github.com/Ulauncher/Ulauncher/blob/dev/scripts/tests.sh#L21) which I guess is the only officially supported way, and wraps the above and sets additional parameters for Display.

However if you run `py.test tests` without setting `PYTHONPATH` (even if we don't support this way), it will test the wrong thing, which can be a cause of great confusion. It will import Ulauncher modules that are installed on the system and run the tests on those. I don't feel comfortable leaving it like that.


gornostal:
> ... it will test the wrong thing, which can be a cause of great confusion

I don't think that's a problem. As long as instructions are clear on how to run tests, user should be able to figure this out

gornostal (approved):

friday:
> I don't think that's a problem. As long as instructions are clear on how to run tests, user should be able to figure this out

Maybe it's not an actual problem, but if this change causes no harm then at least I feel better with it. It would be bad if contributors are lead to believe their code doesn't break tests if it actually does.

gornostal:
Well, it's good to keep tests in a merge request for that reason 🙂

Issue #769, opened 2021-08-09
Ulauncher does not respect github symlinks


tchar:
Hello.
Looks like that when there are symlinks in a repo, ulauncher does not follow the `git clone` method where symlinks are restored and just create files with the symlink inside

Example:
file: `US.svg`
symlink: `USD.svg` -> `US.svg`

- `git clone`: Restores `USD.svg` symlink
- Ulauncher: Creates a file named `USD.svg` with contents `US.svg` (as is showing in github)

**I have the impression though the same will happen with git submodules, haven't tested it though.**

### Steps to reproduce
1. Create a mock extension repo with
```json
[
    { "required_api_version": "^2.0.0", "commit": "master" }
]
```
2. Create a symlink as described above
3. Get the extension through Ulauncher
4. Go to cloned extension
5. cat the symlink

### Current behavior
Symlinks not cloned properly

### Expected behavior
Symlinks should be restored

### Environment
**Ulauncher version**: 5.11.0, API: 2.0.0
**Linux distro**: Ubuntu 20.04.2
**Desktop environment**: Gnome3.36.8

### Logs
Well I have fixed this issue in my repo by removing symlinks and I cannot force ulauncher to clone a specific commit of my repo
If you insist on logs I have to create a template repo to provide logs.

friday:
Thank you for reporting this, but Ulauncher doesn't use or depend on git. It downloads the extensions as zip. See https://developercommunity.visualstudio.com/t/symbolic-links-are-lost-when-downloading-a-git-rep/653461 for another case like this.

We will not introduce a dependency just for this. The problem you were solving with symlinks can be solved in other ways. The only thing I think we can do is add a clarification to the documentation that we don't support symlinks, but I neither think this is a common problem or that people would remember reading that.

tchar:
Hey @friday 

Although that other case may look like similar, I actually did some experiments and found the issue.
Here is a python sample code to reproduce the issue and see that actually the problem lies with python's `ZipFile` and that links are just fine within the downloaded zip.

```python
from ulauncher.api.server.ExtensionDownloader import download_zip, unzip
from shutil import move
import os

# url of my repo with svg
url = 'https://github.com/tchar/ulauncher-albert-calculate-anything/archive/65e34ded08a4d88a66ec9fcd29bec41e57b32967.zip'

# Download with ulauncher's download_zip function
filepath = download_zip(url)

filename = os.path.basename(filepath)
[...]

Issue #273, opened 2018-08-22
Feature: Automatic install of extension dependencies


brpaz:
Hello. 
I think the most common issue people have when installing extensions is because of missing dependencies.

It would be nice if Ulauncher could install dependencies automatically. 
Probably the simplest and more native way, since Ulauncher is built with Python  would be to use requirements.txt file. Ulauncher could detect if the requirements.txt file is present In the extension soure code and if yes it would run ```pip install -r requirements.txt```  automatically before starting the extension.

gornostal:
Hi @brpaz 

Thanks for the suggestion. I understand the problem with the missing dependencies and approach you've described should work.

friday:
For a CLI dependency the below may work. Not really optimal for more complex dependencies, or if it's critical that they get the most recent version, though you could probably adapt it to work for some of these cases too.

```python
import os
from distutils.spawn import find_executable

cmd = 'binary_name'
extDir = os.path.dirname(os.path.realpath(__file__))

# test if client is installed on system and is in PATH or change to local version
if not find_executable(cmd):
    cmd = '{}/{}'.format(extDir, cmd)

# test/get local version
if not find_executable(cmd):
    # Simple example assuming the url is directly to a binary. More likely you'll get an archive or a .deb which needs to be extracted
    urllib.urlretrieve('https://...', cmd)
    os.chmod(cmd, 0o755)
```

brpaz:
Think this is an important issue.Most of the issues opened in my extensions repos are because of missing dependencies.

I try to keep the dependencies required listed in the README but it doesn't seem to help much. ;(

We need a more automatic way.

friday:
Yeah, the average user doesn't bother reading instructions it seem. Just posts straight to Disqus that it doesn't work, even if the dependency is mentioned in the title. :P

Would it be feasible to include it in the source? Your Lorem Ipsum extension for example has a tiny dependency, which is not likely that people have. For GPL dependencies you may have to use GPL too this way though. Not sure.

Also, when failing to meet a dependency, you could perhaps show a notification so users become aware of the issue: https://github.com/Ulauncher/ulauncher-kill/blob/master/main.py#L73

friday:
For `pip` this should work:

```python
import subprocess
import sys

def ensure_import(package):
    try:
        __import__(package)
    except ImportError:
        subprocess.call([sys.executable, "-m", "pip", "install", "--user", package])
    return __import__(package)
```

Example:
```
[...]

PR #770, opened 2021-08-11
Add support for extensions containing symlinks

tchar:
#769 

Changed [ulauncher.api.server.ExtensionDownloader.unzip](https://github.com/Ulauncher/Ulauncher/blob/90452694b6766e8bcc3235ab762d1bde1d9a770b/ulauncher/api/server/ExtensionDownloader.py#L185) to handle symlinks. As discussed in #769 ZipFile handles symlinks differently depending on the context.

This change loops over the entries in a ZipFile object and extracts them one by one. If the entry does not have symlink attributes, or it is a symlink it just proceeds, if the entry is has symlink attributes and is created as a file, it creates the symlink based on the file contents. Symlinks that point outside of the extension directory, are left as a file (original behavior).

#### Some Notes
I can create the symlink by avoiding its extraction and then reading its contents and removing it, but symlinks are rare and I wanted to keep the changes minimal. If this bugs you, I can introduce the boilerplate needed (about ~20 sloc extra) to avoid that.

- [ ] All tests are passing

Just a note. There is no actual testing for `ulauncher.api.server.ExtensionDownloader.unzip`, It is mocked as far as I can tell. If you would like some tests on the unzip, I can write them, but I would need to introduce a sample zip file to be tested against (Python's ZipFile doesn't handle well symlinks so it is hard to emulate it in memory)

- My computer
    - Linux: 5.11.0-25-generic
    - Distro: Ubuntu 20.04.2 LTS
    - DE: Gnome 3.36.8 (X11)
 - Docker image provided by ulauncher

### DIFF
--- a/ulauncher/api/server/ExtensionDownloader.py
+++ b/ulauncher/api/server/ExtensionDownloader.py
@@ -182,7 +182,20 @@ def unzip(filename: str, ext_path: str) -> None:
     temp_ext_path = mkdtemp(prefix='ulauncher_dl_')
 
     with ZipFile(filename) as zipfile:
-        zipfile.extractall(temp_ext_path)
+        for zipinfo in zipfile.infolist():
+            filepath = zipfile.extract(zipinfo, temp_ext_path)
+            # Check if file is symlink from its attributes or zipfile did extract it as a symlink
+            # see https://learning-python.com/cgi/showcode.py?name=ziptools/ziptools/ziptools/zipsymlinks.py
+            if (zipinfo.external_attr >> 28) != 0xA or os.path.islink(filepath):
+                continue
+            with open(filepath) as f:
+                symlink = f.read()
+            symlink_abs = os.path.abspath(os.path.join(os.path.dirname(filepath), symlink))
+            # Don't create symlink if it points outside of extension directory
+            if not symlink_abs.startswith(temp_ext_path + os.sep):
+                continue
+            os.remove(filepath)
+            os.symlink(symlink, filepath)
 
     for dir in os.listdir(temp_ext_path):
         move(os.path.join(temp_ext_path, dir), ext_path)


friday:
Appreciated! I hope I can get to this soon.

ulauncher/api/server/ExtensionDownloader.py
```diff
@@ -182,7 +182,20 @@ def unzip(filename: str, ext_path: str) -> None:
     temp_ext_path = mkdtemp(prefix='ulauncher_dl_')
 
     with ZipFile(filename) as zipfile:
-        zipfile.extractall(temp_ext_path)
+        for zipinfo in zipfile.infolist():
+            filepath = zipfile.extract(zipinfo, temp_ext_path)
+            # Check if file is symlink from its attributes or zipfile did extract it as a symlink
+            # see https://learning-python.com/cgi/showcode.py?name=ziptools/ziptools/ziptools/zipsymlinks.py
+            if (zipinfo.external_attr >> 28) != 0xA or os.path.islink(filepath):
```
troycurtisjr, line 189: I see, this is due to the fact that to git, and thus inside this zip archive, the symlink is really just a text file with the path as contents. So this snippet is trying to convert that normal text file into a true symlink, as long as it is within the extension directory. Could you add a few comments calling out the fact that most/all of the times zips from Git will have the symbolic links represented as text files? That might help future us understand what is going on here and why.

ulauncher/api/server/ExtensionDownloader.py
```diff
@@ -182,7 +182,20 @@ def unzip(filename: str, ext_path: str) -> None:
...
+                continue
+            with open(filepath) as f:
```
troycurtisjr, line 191: I might be concerned about whether or not this is a directory, but I think the symlink check above is sufficient to be sure that if we get here what we have is a small text file containing the path to point to. :+1:

troycurtisjr (changes requested):
@tchar would you be up for adding a unit test for this change? I could definitely see changes being made accidentally that break the capability. Also I think it is unusual enough it might be a while before we actually figure out it got broke.

ulauncher/api/server/ExtensionDownloader.py
tchar: Actually it is not only to git but to any normal  zip file. The current unzip function cannot handle any of these due to `ZipFile` not handling symlinks. You can test that by creating locally a zip with symlinks inside and pass it to Ulauncher's unzip function (that is what I actually tested in the first place). The `zipinfo.external_attr >> 28` check is generic as can be seen [here](https://learning-python.com/cgi/showcode.py?name=ziptools/ziptools/ziptools/zipsymlinks.py) at the `isSymlink` function.

I tested my function against a directory downloaded with Ulauncher's download_zip function and also a normal zip created locally by me totally unrelated to git. It works in both cases.

ulauncher/api/server/ExtensionDownloader.py
tchar: Directories return False for (zipinfo.external_attr >> 28) != 0xA.

Actually my first tries had a safeguard like this
```python
# zipinfo.is_dir(), only python >= 3.6
if (zipinfo.external_attr >> 28) != 0xA or zipinfo.is_dir() or os.path.islink(filepath):  
    ...
```

tchar:
> @tchar would you be up for adding a unit test for this change? I could definitely see changes being made accidentally that break the capability. Also I think it is unusual enough it might be a while before we actually figure out it got broke.

Yeah sure I can add tests, but I need to add a couple of zips somewhere in the `tests` directory and actually extract them (without mocker) when testing. I can do that, no problem.

I would do it in-memory but since ZipFile struggles with symlinks in the first place I can't create them in-memory using it.

friday:
I forked your extension and reverted to the state when you had symlinks: https://github.com/friday/ulauncher-albert-calculate-anything

Then instead of downloading the zip, I downloaded the tarball: https://github.com/friday/ulauncher-albert-calculate-anything/tarball/master

I just followed the tarball documentation on how to extract:
```
import tarfile
tf = tarfile.open("friday-ulauncher-albert-calculate-anything-v0.8.0-27-g65e34de.tar.gz", mode="r")
tf.extractall()
```

This worked straight away for me, and is much cleaner imo, so I suggest that way instead. The only issue is that the root directory has the commit hash in it. `g65e34de`, so this needs to be handled instead.

friday:
Oh actually, you can just change the file extension on the download url from zip to tar.gz and it doesn't include the commit hash. I couldn't find that documented anywhere though.

friday:
Alternative implementation using tar #773

tchar:
> I forked your extension and reverted to the state when you had symlinks: https://github.com/friday/ulauncher-albert-calculate-anything
> 
> Then instead of downloading the zip, I downloaded the tarball: https://github.com/friday/ulauncher-albert-calculate-anything/tarball/master
> 
> I just followed the tarball documentation on how to extract:
> 
> ```
> import tarfile
> tf = tarfile.open("friday-ulauncher-albert-calculate-anything-v0.8.0-27-g65e34de.tar.gz", mode="r")
> tf.extractall()
> ```
> 
> This worked straight away for me, and is much cleaner imo, so I suggest that way instead. The only issue is that the root directory has the commit hash in it. `g65e34de`, so this needs to be handled instead.

Pff I was hoping to get some PR to fix some of the mess in my code.

Yeah tarfile looks better, you could wait though to make the changes in this branch so we don't make multiple PRs for the same thing.

friday:
> Yeah tarfile looks better, you could wait though to make the changes in this branch so we don't make multiple PRs for the same thing and I didnt sit and try to contribute for no reason.

Please understand that getting your contributions merged is never guaranteed. I only asked for the PR because you said you had written the code already. A PR just makes that work more formal and easy to review, but I never wanted to encourage you to put time on this. My plan was to look for an alternative to avoid introducing the extra complexity.

Most of the PRs which are on the list that haven't been merged so far will never be merged. There are many considerations involved in making these decisions that maintainers have to weight, because if it's a mess or seems risky, it'll be their mess to work with later. And if you have a long experience working with software development you do get pickier (for a good reason). It also takes a lot of time to review PRs and suggest changes, and in this case it took very little time for me to make that PR after doing the research, so it was easier for me that way.

tchar:
> > Yeah tarfile looks better, you could wait though to make the changes in this branch so we don't make multiple PRs for the same thing and I didnt sit and try to contribute for no reason.
> 
> Please understand that getting your contributions merged is never guaranteed. I only asked for the PR because you said you had written the code already. A PR just makes that work more formal and easy to review, but I never wanted to encourage you to put time on this. My plan was to look for an alternative to avoid introducing the extra complexity.
> 
> Most of the PRs which are on the list that haven't been merged so far will never be merged. There are many considerations involved in making these decisions that maintainers have to weight, because if it's a mess or seems risky, it'll be their mess to work with later. And if you have a long experience working with software development you do get pickier (for a good reason). It also takes a lot of time to review PRs and suggest changes, and in this case it took very little time for me to make that PR after doing the research, so it was easier for me that way.

Yeah I had written the code but the PR requires from me to download the repo, branch dev, install docker and deps, format the code accordingly to the repo's specs, pass the tests etc. The prototype I had did not need any of that work. And most of the work was basically configuring my environment with linting deps etc for the PR, not writing the code.

I am just pointing out that not making comments on the existing PR, fixing it in the maintainer's way, discourages people from spending any time contributing to any project. The work is not the code, that's easy, the work is making the code ready for review etc.

I am not arguing on the pickyness actually this is correct but it can be done through discussion and code review.

Anyways. You can close this PR, since you already fixed it.

friday:
This happened to me many times @tchar. For example in https://github.com/tkashkin/GameHub/pull/171. I had to make the PR to prove it could be done, and hopefully it helped the maintainer make his counter-suggestion. The solution he went with was better, so I was happy that my use case was covered, and I appreciate that he did that instead of telling me how to change my code to become his code. That back and forth would have been a waste of his time and mine.

The experience definitely didn't encourage me to continue contributing to that project, but I blame my PR and limited knowledge of Vala async event handling, and not the project maintainer for doing the right thing.

tchar:
@friday 

I partly agree with you.

Tarfile is indeed a better solution if you don't care changing dependencies.

However given that you are a project maintainer and you have experience with softeng I thought you had a reason to keep `ZipFile` and already knew that this problem exists. I simply went with ZipFile because you said you don't want to alter dependencies. And this is just experience with talking with multiple project maintainers because one doesn't like x and the other doesn't like y.

Anyways my main problem with these kind of projects is that I have to maintain a separate branch for a ton of projects because [either a feature is not implemented](https://github.com/Ulauncher/Ulauncher/issues/273) and I have to [implement](https://github.com/Ulauncher/Ulauncher/compare/dev...tchar:requirements?expand=1) it myself or there are [memory issues causing  a lightweight program to be as heavy as gnome](https://github.com/tchar/qtile/commit/4127cf35bcce2a989647cfade159c1571edf6c51) or just simply devs don't want to support a feature.

friday:
ZipFile and TarFile are both standard Python modules written in Python with no external dependencies. Switching from one to the other doesn't change dependencies. Switching to use git or to use the `unzip` or `tar` commands via subprocess changes the dependencies, because then everyone using Ulauncher has to install those. Although most users probably have them, it's better to avoid adding dependencies.

The rest is good criticism imo. I also want to improve things like this. Extension dependencies is one of the more important issues for Ulauncher to handle imo, along with HiDPI support and fixing crashes. We are prioritizing this. This is a big and important internal change and something I want to fully understand before I integrate it.

Issue #477, opened 2019-12-19
Extensions: How to escape html-special characters (<, >, ...) in ResultItems


dhelmr:
## Question

Hi, for [my extension](https://github.com/dhelmr/ulauncher-duckduckgo-bangs), I try to figure out how html-special characters like `<`, `>` and so on should be escaped in the `name` attribute of`ExtensionSmallResultItem` or `ExtensionResultItem`.

* if the item name only contains `>`, everything is fine:
![image](https://user-images.githubusercontent.com/16763338/71189990-00a40c80-2284-11ea-9293-896336be36c0.png)
* If I try to display a sequence that starts an invalid tag (e.g.  `< ...`  or `<>`), ulauncher does not display anything (see error log below)
![image](https://user-images.githubusercontent.com/16763338/71190126-47920200-2284-11ea-9715-d40fc3841759.png)
* when I escape them with `&lt;` or `&#60;`, these strings are displayed as they are:
![image](https://user-images.githubusercontent.com/16763338/71189564-4f04db80-2283-11ea-9144-f5812b80ca45.png)

So my question is: How can I display such characters properly in the `name` attribute? I think ulauncher tries to parse it as html and thus fails if it contains invalid tags? Can this somehow be turned off? Sorry if I have overseen something in the docs. 

I am using ulauncher 5.4.0

## Error Log

```
Failed to set text '<span foreground="#2c74cc">wa</span> | Wolfram|alpha: Search for<span foreground="#2c74cc"> <</span>b><</b>' from markup due to error parsing markup: Error on line 1 char 104: “<” is not a valid name
```

Thanks in advance!

gornostal:
@dhelmr Ulauncher was replacing `&` with `&amp;` that's why escaping didn't work for you.
I could not track down why I've added that. It seems to work fine without that replacement, so I deleted it in this commit: https://github.com/Ulauncher/Ulauncher/commit/241fd593248a07fcca76c739940226b884915bf1

I'll post an update here when the fix is available in dev and stable release

dhelmr:
Alright, thank you :)

gornostal:
@dhelmr the issue is fixed in 5.4.1. Make sure you  `html.escape` whatever string to pass as a title and it should work.

Please reopen if needed.

dhelmr:
Hi, sorry for not responding earlier, but I only just now found the time to take a look at this again.
[...]

PR #771, opened 2021-08-11
Fix ampersand breaking highlighting

### LINKED ISSUE #587: Ampersand (&) is not handled properly
**Steps to reproduce:**

On Ubuntu, type "Software", which should match the "Software & updates app. Because ampersand character is escaped, the app name is not displayed.

Screenshot from 2020-07-12 14-50-45

**Ulauncher version**: 5.8.0
**Linux distro**: ubuntu 20.04
**Desktop environment**: i3

**Logs**:
```
(ulauncher:9242): Gtk-WARNING **: 14:43:53.729: Failed to set text '<span foreground="#2c74cc">Software</span> & Updates' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

friday:
This should fix #587 which was broken by fixing #477, which imo is a less important use case to support

Before:
before

`Failed to set text '<span foreground="#fff">Date &</span>amp; Time Settings' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;`

As can be seen, the issue is that the highlighter splits up the `&amp;` entity with a html tag. By replacing `&amp;` to `&` and then back again after highlighting this is fixed. In addition this escapes unescaped `&`.

After:

It breaks #477 again for https://github.com/dhelmr/ulauncher-duckduckgo-bangs, but not for my extension. I think I originally [disabled hightlighting like this](https://github.com/friday/ulauncher-clipboard/blob/master/main.py#L25) to solve that, but then didn't even need to do that when switching to the small result view. I think this is a reasonable workaround/limitation for the rare exceptions of extensions out there that want to display special characters.

### DIFF
--- a/tests/utils/test_text_highlighter.py
+++ b/tests/utils/test_text_highlighter.py
@@ -9,4 +9,5 @@ def test_highlight_text():
     assert hl('hell wo', 'hello world') == '<i>hell</i>o<i> wo</i>rld'
     assert hl('ttesti', 'testik_ls-ttestk') == '<i>testi</i>k_ls-ttestk'
     assert hl('dome', 'Documents') == '<i>Do</i>cu<i>me</i>nts'
-    assert hl('e tom', 'São tomé & príncipe') == 'São<i> tom</i>é & príncip<i>e</i>'
+    assert hl('e tom', 'São tomé & príncipe') == 'São<i> tom</i>é &amp; príncip<i>e</i>'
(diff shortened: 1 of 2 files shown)


troycurtisjr (approved):
LGTM :+1:

PR #773, opened 2021-08-12
Switch extension download link to use tar instead of zip (fixes symlinks)

### LINKED ISSUE #769: Ulauncher does not respect github symlinks

friday:
This fixed #769 in my basic tests.

Apart from the one line that actually extracts the files I just search for and replaced all occurrences of "zip" without fully understanding/checking the flow, but it worked in my tests and doesn't break the tests.

Can you verify it's working for you @tchar? I could install my fork of your extension reverted to when you had symlinks: https://github.com/friday/ulauncher-albert-calculate-anything, and the usd symlink was working. I didn't actually run your extension though.

### DIFF
--- a/tests/api/server/test_ExtensionDownloader.py
+++ b/tests/api/server/test_ExtensionDownloader.py
@@ -26,7 +26,7 @@ class TestExtensionDownloader:
     def gh_ext(self, mocker):
         gh_ext = mocker.patch('ulauncher.api.server.ExtensionDownloader.GithubExtension').return_value
         gh_ext.get_ext_id.return_value = 'com.github.ulauncher.ulauncher-timer'
-        gh_ext.get_download_url.return_value = 'https://github.com/Ulauncher/ulauncher-timer/archive/master.zip'
+        gh_ext.get_download_url.return_value = 'https://github.com/Ulauncher/ulauncher-timer/archive/master.tar.gz'
         gh_ext.get_last_commit.return_value = {
             'last_commit': '64e106c',
             'last_commit_time': '2017-05-01T07:30:39'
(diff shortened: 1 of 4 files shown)


tchar:
Yeah works, for me.

I would insist on using the `/tarball/` link. I could not find the `/archive/` in a documentation or smth.

friday:
> Yeah works, for me.
> 
> I would insist on using the `/tarball/` link. I could not find the `/archive/` in a documentation or smth.

Good point. I pushed that now. I had that concern, then forgot it. /archive/ is the one Github's own download button uses, but I never found it in any API documentation either.

friday:
Thank you for testing @tchar :+1:

tchar:
Another note which you have to take into consideration.

Links in the tar that refer outside of the extension's directory, will be extracted normally, not sure if this could be a security issue or smth (this has to be tested).

tchar:
For example if there is a link inside that refers to /dev/null will be extracted and will refer to /dev/null, if you want to allow extensions to have access outside of their "contained space" then no worries, otherwise you have to test that.

friday:
Ulauncher is not running as sudo, so this isn't really a problem. Any file that this potential symlink vulnerability could read the extensions can already read.

Issue #700, opened 2021-02-28
Prioritize exact matches?


ckp95:
I'm not sure if this is a bug but it's a bit annoying. When I type things, exact matches aren't prioritized. For example, when typing "thunar`, the Thunar program is not the top match (this is on a fresh install):

![image](https://user-images.githubusercontent.com/43073868/109437023-acb1bd80-7a1a-11eb-878e-c81dfa0effee.png)

I couldn't find a setting to prioritize exact matches.

In practice this isn't a huge problem because Ulauncher remembers frequently chosen options, so Thunar appears at the top after having chosen it before. But it's still a bit annoying for rarely used programs.

KuenzelIT:
Probably a duplicate of #232

friday:
This is fixed in 5.12.0, that was just released for Arch and Ubuntu/Debian packages. The Fedora build will be released soon too.

Issue #558, opened 2020-05-12
Searching with a program's partial name does not actually return the program


Vinno97:
### Is your feature request related to a problem? Please describe

A number of programs have names that are comprised of compound words or a portmanteau. To open a program like _FreeOffice TextMaker_ or _balenaEtcher_, I'd like to just type "office" or "etcher" and be presented with the corresponding programs. 
For some reason, the current fuzzy search algorithm is only able to find these programs if I start my query with "free" or "balena", respectively. Also, for some reason "etcher" does bring up programs like _Zo**ter**o_, _Do**c**k**er**_ and _**Ch**arac**ter**s_.

### Describe the solution you'd like
A fuzzy-search algorithm that 1) returns a program if the query matches a part of the program's name and 2) prioritizes this program (since it _obviously_ has the best match)

### Additional context

My first instinct was that it does not search for matches inside a program's name, but searching "ens" brings up _psensor_. As someone with no internal knowledge of the program, this is quite confusing.

PR #774, opened 2021-08-13
Fuzzy improvements

friday:
This changes the behavior of fuzzy matching

Levenstein distance is the amount of changes needed between one string of text to become the second one. The previous implementation didn't use it in that way, leading to low scores for what I would consider to be positive matches like "wif" for "Wi-Fi", because it was actually checking how many letters to change in "wif" to become "Wi-Fi gnome-control center" (which is most of them), and hence giving it a low score.

1. Instead of concatenating "Wi-Fi gnome-control center" or "Files nautilus" and running the query on that, it runs separately on "Wi-Fi" and "gnome-control center", and picks the one with the best score.
2. #558 Make a custom method for matching queries with results that counter-weights the "how many letters to change to become" part so that short queries can match with long results (also removed tests that expect short queries to favor short results)
3. Change the weighting so that app names are considered slightly more important than executable names, which are sometimes the same across several applications in the same suit (like "Thunar Bulk Rename" in #700).
4. Set a much higher threshold for what's considered a match (it needed to be low before, but now because of 1. and 2. those low scores are just noise). #324

Will probably need to tune this later, but it's a heck of an improvement already. The code is not super pretty, because it's overloading `search_name` and `get_search_name()`. This is a hard problem to solve, since the class they're in is used both for apps and the file browse mode, and only the former really needs this method and the name of the executable.

In my fork I also added the description (and tags, but not properly implemented), and "normalized" both the results and queries, so that only letters and numbers were kept. So for example "Add/Remove Software" and "Wi-Fi" stripped the slash and dash, both in the query and the results used for matching, but I think of these are more controversial changes that can wait.

### Checklist
- [x] All tests are passing

### DIFF
--- a/ulauncher/search/SortedList.py
+++ b/ulauncher/search/SortedList.py
@@ -49,7 +49,7 @@ class SortedList:
         name, exec_name, *_ = f'{search_fields}\n'.split('\n')
         score = max(
             get_score(self._query, name),
-            get_score(self._query, exec_name)
+            get_score(self._query, exec_name) * .8
         )
 
         if score >= self._min_score:
--- a/ulauncher/search/SortedList.py
+++ b/ulauncher/search/SortedList.py
@@ -43,7 +43,15 @@ class SortedList:
             self.append(item)
 
     def append(self, result_item):
-        score = get_score(self._query, result_item.get_search_name())
+        # get_search_name() returns a string with the app display name, but it may contain a
+        # second line in which case that line is the name of the executable
+        search_fields = result_item.get_search_name()
+        name, exec_name, *_ = f'{search_fields}\n'.split('\n')
+        score = max(
+            get_score(self._query, name),
+            get_score(self._query, exec_name)
+        )
+
         if score >= self._min_score:
             result_item.score = -score  # use negative to sort by score in desc. order
             self._items.insert(result_item)
--- a/ulauncher/search/apps/AppDb.py
+++ b/ulauncher/search/apps/AppDb.py
@@ -169,11 +169,4 @@ def search_name(name, exec_name):
     Returns string that will be used for search
     We want to make sure app can be searchable by its exec_line
     """
-    exec_name_split = set(exec_name.split('-'))
-    name_split = set(name.lower().split(' '))
-    common_words = exec_name_split & name_split
-
-    if common_words or not exec_name:
-        return name
-
-    return '%s %s' % (name, exec_name)
+    return f'{name}\n{exec_name}'
--- a/ulauncher/search/apps/AppDb.py
+++ b/ulauncher/search/apps/AppDb.py
@@ -144,7 +144,7 @@ class AppDb:
         :rtype: :class:`ResultList`
         """
 
-        result_list = result_list or SortedList(query, min_score=50, limit=9)
+        result_list = result_list or SortedList(query, min_score=75, limit=9)
 
         if not query:
             return result_list
--- a/ulauncher/search/apps/AppDb.py
+++ b/ulauncher/search/apps/AppDb.py
@@ -69,7 +69,7 @@ class AppDb:
         :param Gio.DesktopAppInfo app:
         """
         name = app.get_string('X-GNOME-FullName') or app.get_name()
-        exec_name = app.get_string('Exec') or ''
+        exec_name = get_exec_name(app.get_string('Exec') or '')
         description = app.get_description() or ''
         if not description and (app.get_generic_name() != name):
             description = app.get_generic_name() or ''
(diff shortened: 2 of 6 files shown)


ulauncher/search/SortedList.py
```diff
@@ -43,7 +43,15 @@ def extend(self, items):
             self.append(item)
 
     def append(self, result_item):
-        score = get_score(self._query, result_item.get_search_name())
+        # get_search_name() returns a string with the app display name, but it may contain a
+        # second line in which case that line is the name of the executable
+        search_fields = result_item.get_search_name()
+        name, exec_name, *_ = f'{search_fields}\n'.split('\n')
```
friday, line 49: I can't decide if I think this line is clever or ugly. I'm leaning towards being counted as clever in other languages, but not "Pythonic". The problem is that result_item.get_search_name() returns just the name for SortedLists that are not a applications.

troycurtisjr: I vote clever as well.

ulauncher/search/apps/AppDb.py
```diff
@@ -155,28 +155,18 @@ def find(self, query, result_list=None):
...
-    # drop "/usr/bin/"
-    match = re.match(r'^(\/.+\/)?([-\w\.]+)([^\w\/]|$)', exec_name.lower(), re.I)
-    if not match:
-        return name
-
-    exec_name = match.group(2)
-    exec_name_split = set(exec_name.split('-'))
-    name_split = set(name.lower().split(' '))
-    common_words = exec_name_split & name_split
+    match = re.match(r'^(env\s+)?([^\s=]+=[^\s=]+\s+)*([^\s=]+\/)*(?P<bin>[^\s=]*)', exec, re.I)
```
troycurtisjr, line 163: The env vars could be quoted values with spaces, right? 
```
>>> from ulauncher.search.apps.AppDb import get_exec_name
>>> get_exec_name('env MYVAR="value with spaces" /usr/bin/foo --option')
'with'
```

troycurtisjr: though it's no worse than it was already and that seems to have worked well enough

ulauncher/utils/fuzzy_search.py
```diff
@@ -48,19 +48,24 @@ def get_matching_indexes(query, text):
     return sorted(positions)
 
 
-def get_score(query, text):
+def get_score(query="", text=""):
```
troycurtisjr, line 51: defaulting both args seems to indicate it is meaningful and expected to call as `get_score()`, or maybe call as `get_score(text="something")`. While it will return a score of `0` for `get_score()`, which is reasonable in some sense, I'm not sure it really matches with the semantics of the function. Would it not be better to error if this was ever called with anything other than exactly 2 arguments?

ulauncher/search/apps/AppDb.py
friday: True. I didn't know how to support that without actually looping through character by character, because it could be single quotes as well, or single quotes with double quotes inside (or reversed). Possibly also backticks to run things through the shell. I think if we want to support that it'll add up to a lot more complex than a regex can support.

I ran `cat /usr/share/applications/*.desktop | grep Exec=` Ulauncher is the only app I have that even uses env.

Gio.DesktopAppInfo.get_executable() actually returns `env` instead of the app name for Ulauncher

ulauncher/utils/fuzzy_search.py
friday: Good point. I can't even remember why I did that.

friday: It didn't break any behavior or test, so pushed it as 7f720f6

ulauncher/search/apps/AppDb.py
troycurtisjr: > Gio.DesktopAppInfo.get_executable() actually returns env instead of the app name for Ulauncher

ha! Yeah that is totally unhelpful for us. I suppose that is technically correct, since the `env` executable is in fact the executable that will be executed when the desktop file is run. It just happens that env will immediately turn around and exec a different executable based on its arguments! 

Regardless it does seem very uncommon to include any environment vars in desktop files, I also could not find any other examples

friday: It's also not absolutely critical that it gets the executable imo. It's just for apps which has known named like "Gedit", "Nautilus" etc that are named "Text Editor" and "Files" in their desktop entries. I thought about ditching this method entirely to go for Gio.DesktopAppInfo.get_executable(). If I wrote the initial implementation knowing all this, that's what I would do.

ulauncher/utils/fuzzy_search.py
```diff
@@ -48,19 +48,24 @@ def get_matching_indexes(query, text):
...
-    # increase score if a word from text starts with a query
+    query_len = len(query)
+    if not query or not text:
+        return 0
+    if text.startswith(query):
+        return 100
+    # Wrap and counter-weight distance so that short queries can match better with long texts.
+    # With regular distance "Fir" and "Firefox" is only 3/7 similar
+    # We want them to be counted as near 100% matches (if you have an actual app named "Fir"
+    # it's better that is listed first).
```
troycurtisjr, line 66: Doesn't the `text.startswith(query)` check above contradict this comment and the provided example? It mentions how `fir` will match to `firefox` and `fir`, and that `fir` should be scored higher. However, in practice with this `get_score()` implementation it will return 100 for both queries since it is a prefix for both. tbh I think we have to be very clever to get those two cases sorted right while only looking at each match independently. 

Perhaps some logarithmic score increase for the number of exact characters matched in the same order from the beginning of the word which quickly rises to perfect score but perhaps never reaching it. Though maybe Levenshtein's already does something similar? It's been a while since I've looked at it in detail.

Regardless this falls squarely in the "requires more tuning" that is likely an ongoing effort, and in this case at least the 2 exact prefix matches show up near the top, even if the exact relative order between them is indeterminate.

I think the algorithm seems good to go with, but I did want to point out that the comment is slightly misleading.

friday: Good point. You are right. I should change the comment or code. Not sure which yet

friday: I should probably remove text.startswith(query) actually. It's already covered below.

friday: Pushed as 9db6eea. I should have been more patient to merge this PR haha :D

I removed text.startswith(query), rewrote the comment to be technically correct and added a text to show that the 95% weighting does work, just as long as the query doesn't match the start of a word.

PR #778, opened 2021-08-15
Prevent apps from exiting if Ulauncher crashes or is terminated

### LINKED ISSUE #754: Killing Ulauncher kills all apps launched with Ulauncher
### Steps to reproduce
1. Start ulauncher
2. Start some program using ulauncher
3. Kill ulauncher

### Current behavior
Program started with ulauncher gets killed when ulauncher is killed.

### Expected behavior

If Ulauncher crashes or is killed, all programs launched with Ulauncher should keep running.

To start an application, ulauncher should fork --- parent process of ulauncher keeps running, child process of ulauncher launches specified app and then exits.

### Environment
**Ulauncher version**:  5.11.0
**Linux distro**:  Archlinux
**Desktop environment**:  swaywm

friday:
Hopefully fixes #754 for good now. It was more complicated than I originally thought as the issue reappeared after the first fix in my fork

* Switch from using `app.launch()` to using subprocess to open all apps, so we can control how they open
* Open with `start_new_session=True`, which prevents child-processes from being killed if you're running the app from the terminal and ctrl+c (SIGINT)
* The above doesn't work when the app is running through systemd, so also test if `systemd-run` exists, and run with that to set a different cgroup (this also fixes the case above, but not if `systemd-run` is missing, which is the reason we need both ways)
* Unrelated to #754 it also properly strips field codes (%f, %F, %u, %U, etc) from the exec line now, which fixes some cases when using custom terminal.

### DIFF
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -1,6 +1,7 @@
 import logging
-import pipes
 import subprocess
+import re
+import shlex
 
 from ulauncher.utils.desktop.reader import read_desktop_file
 from ulauncher.utils.Settings import Settings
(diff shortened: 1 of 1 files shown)


kushraj:
still happening on sway for me, On reloading sway config, sometimes Ulauncher exit automatically freezing or killing windows opened by it . 
ulauncher 5.14.0-1
sway 1:1.6.1-1
wlroots 0.14.1-2

PR #782, opened 2021-08-19
Changes to Travis build

friday:
This is untested. Merging the PR is the only way to test. The PR is mostly to collect the changes somewhere. I think I will merge it rather than wait for reviews, but I will wait until after 5.12.0.

Changes:
1. The aur package `ulauncher-git` builds from the main git branch and is not connected to releases. It's not to be considered a pre-release repo, but a way to build Ulauncher from the git branch. Making releases to this package is undesirable. It should by Arch package conventions behave exactly like it does without releases or building from a specific tag or tarball, so we should never make releases to this package.
2. Replacing `-` with `~` should only be for the debian package, and it should happen always. Not just if the label is "beta" or "dev". So now it's only done inside of the build-deb script, and we can know that it doesn't have a ~in all other cases. The same does apply to Arch versions, but we don't need to handle that because of 1 ^
3. If we use a git tag containing "test" it should skip uploads now. In addition it's possible to push only the tag, but not the branch it's on. That way it's just a "stray" tag we can delete afterwards.


friday:
Hope I didn't break anything. I did test my changes only in isolation.

Issue #649, opened 2020-11-27
Emacs/Shell keybindings 


troycurtisjr:
### Is your feature request related to a problem? Please describe
I have the emacs keymap enabled in my Gnome session. The typical text editing commands work fine in ULauncher (C-b, C-f, etc), but I would also like to have C-n for down and C-p for up to scroll through the list.  

I plan on trying to add it in, but I want to make sure the PR would be accepted before bothering.

### Describe the solution you'd like
Ideally we'd be able to detect the current keymap somehow and do the right thing automatically. If that doesn't work, then I would propose adding an options to the preferences.

friday:
If this is something that can be handled without adding new complicated settings to Ulauncher or locking in to specific key combinations supported that would be great. I'm definitely reluctant to add settings for all these keys, and also to support both emacs and vim, not to mention adding the Vim way with insert mode and normal mode in Ulauncher.

friday:
5.13.0 is out now, so this should be fixed

PR #796, opened 2021-08-31
Implement standard shell/emacs ctrl-n/ctrl-p scrolling.

troycurtisjr:
#649 

It turns out that there was no need for any settings or fancy logic. Ctrl-n and Ctrl-p have no effect in current ulauncher and so are free for use. This PR simply support the Ctrl-n & Ctrl-p option specifically because Ulauncher is responsible for intercepting and responding to keypresses to scroll through results. For other emacs editing features, the user can enable the emacs keymap in Gtk as Ulauncher is a Gtk application. With custom themes it is even possible to enable it only for Ulauncher and not for other apps. However, all of that is out of scope for what Ulauncher is in charge of. Just like it doesn't implement "backspace" in the input field, neither does it implement all the normal emacs line editing commands. 

- [x] All tests are passing

Fedora 34, Gnome 40

### DIFF
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -178,13 +178,14 @@ class UlauncherWindow(Gtk.Window, WindowHelper):
         keyval = event.get_keyval()
         keyname = Gdk.keyval_name(keyval[1])
         alt = event.state & Gdk.ModifierType.MOD1_MASK
+        ctrl = event.state & Gdk.ModifierType.CONTROL_MASK
         Search.get_instance().on_key_press_event(widget, event, self._get_user_query())
 
         if self.results_nav:
-            if keyname in ('Up', 'ISO_Left_Tab'):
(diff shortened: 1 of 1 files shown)


friday:
Thank you, and sorry it took so long to merge. I think I fixed my notifications (didn't get any for this).

I'm a little bit reluctant to add fixed keyboard combinations in case there's another use case this clashes with. For example say editor X and editor Y uses the same combination to mean different things.

troycurtisjr:
Yeah that definitely makes sense in general. I think this particular
combination is slightly more common than typical since it is default in
most shells, available as a keymap in Gnome, and part of the default keymap
in MacOS. Not that any of those things really prevent some conflicts, like
screen using Ctrl-A and tmux using Ctrl-B as their default "escape" binding
even though they are directly targeted at use with shell environments and
both of those conflict with default shell shortcut sequences. In the end
there are only so many keystrokes that can be used 😆!

The alternative to fixed combinations is more configuration options, which
I'm certainly all for, especially for a productivity app like Ulauncher.

Issue #369, opened 2019-05-08
Upload to Debian and Ubuntu repositories?


fossfreedom:
Hi there,

  I am the project lead for Ubuntu Budgie and I'm considering currently whether to add ULauncher as a core component of our distro.  First question - would you be happy with this?  Any reservations that you may have?

The python3 conversion that you have done is a key requirement for us - so it is exciting that you have just started this update.  brilliant news :)

To make ulauncher a key component of UB I will need to create a Debian package that meets current Debian expected standards.  I will then run with this and ask Debian whether they are happy to include your project in their repositories.

There are a number of questions though with the project source and installation structure that need to be addressed though.  Thoughts on the following issues highlighted by Debian lintian?:

1. usr/share/ulauncher/preferences/README.md --> Why is this an installed file?
2. usr/share/ulauncher/preferences/static/fonts/FontAwesome.otf also in fonts-glewlwyd --> why are static fonts being used? Can ulauncher use user defined fonts?
3. font-outside-font-dir usr/share/ulauncher/preferences/static/fonts/FontAwesome.otf --> several font files are included in the source and installed in a odd location (should be /usr/share/fonts) ... similar to 2 - maybe have the ability in the setup package to not install fonts files?
4. binary-without-manpage usr/bin/ulauncher  & binary-without-manpage usr/bin/ulauncher-toggle  --> anything installed in /usr/bin should have the ability to "man ulauncher" and "man ulauncher-toggle"
5. desktop-entry-lacks-keywords-entry usr/share/applications/ulauncher.desktop
6. desktop-command-not-in-package usr/share/applications/ulauncher.desktop env --> would you please consider adding the GDK_BACKEND=x11 environment variable via the python script rather than via the desktop file?
7. debian subfolder - please can you consider moving the debian subfolder and scripts to another branch.  Usually the master branch is just for source files - packaging is handled in other ways such as another branch
8. release versioning. Any chance that in future releases, release versions are tagged "v5.0.0" whereas dev releases are tagged something like "dev xyz" ?  Debian has a concept of watching for new releases. Currently I see v5.0.0r5 which is a dev release will confuse the watch system because it cannot distinguish between a final release and a interim dev drop

gornostal:
Hi @fossfreedom 
I would be very happy to see Ulauncher as a core component of Ubuntu Budgie. 
I have no reservations like that, much less exclusive ones.

[...]

PR #797, opened 2021-09-01
some minor Debianization file updates

SwampRabbit:

#369 

minor updates to Debianization files

Debian 11 (Bullseye) and MX Linux 21 Beta 1

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,6 +9,8 @@ Build-Depends: debhelper (>=9~),
 Maintainer: Aleksandr Gornostal <ulauncher.app@gmail.com>
 Standards-Version: 4.1.4
 X-Python3-Version: >= 3.5
+Homepage: https://ulauncher.io/
+Vcs-Browser: https://github.com/Ulauncher/Ulauncher
 
 Package: ulauncher
 Architecture: all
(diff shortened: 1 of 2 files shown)


friday:
Thank you! I don't know a lot about debian packaging so I trust that you know what you're doing.

One issue I am aware of is that we're generating the changelog with the same entry every time now. I have been planning to look at actually using a changelog and semi-generate it from commit messages, but this likely won't happen soon.

SwampRabbit:
@friday I'm no genius, but I hope I know what I'm doing, I'm on the MX Linux team.  ;)

I'm not sure about generating it from actual GitHub commits though, sorry can't help with that, the only reservation I would have with doing that would be that the entries could get insanely long.  Doesn't hurt anything though.

What I'd do is just run the dch command for each major/stable release to get a new entry, then at least add the same thing your adding on the release page.

I'll be doing a bit more on your debianization (its pretty good already) when I can swing some time, I maintain a lot of stuff.

friday:
> The only reservation I would have with doing that would be that the entries could get insanely long. Doesn't hurt anything though.

The last two release change notes were generated from commits actually. So it's not a problem. There's a manual step where I filter out irrelevant commits and can edit the messages. I just need to put these messages in a format that can be accessed by the deploy script.

> I'll be doing a bit more on your debianization (its pretty good already) when I can swing some time, I maintain a lot of stuff.

Thank you for this. :+1:

PR #799, opened 2021-09-03
Add ulauncher.service file to the debian package.

troycurtisjr:
The CI failure after merging my setup.py fixes which include the service file failed because the debian package script didn't copy over the service file. This adds that tweak.

I used the build scripts to build a debian package on a Ubuntu 16.04 VM (which appears to be the version used in the CI container) and installed it to test the fix. I also installed the generated package on Ubuntu 20.04. Note that to get the service file to work on the older Ubuntu additional configuration has to be done in order to provide ulauncher with the right environment, but works out of the box on the newer Ubuntu.

- [x] All tests are passing

Ubuntu 16.04, Gnome 3.14
Ubuntu 20.04, Gnome 3.30

### DIFF
--- a/scripts/build-deb.sh
+++ b/scripts/build-deb.sh
@@ -55,6 +55,7 @@ build-deb () {
         setup.py \
         ulauncher \
         ulauncher.desktop.in \
+        ulauncher.service \
         $tmpsrc \
         --exclude-from=.gitignore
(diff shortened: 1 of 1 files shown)


friday:
Nice!

Obvious fix in retrospect :D I was trying to replicate and fix the issue, but it actually failed in my Docker image for some unrelated encoding error, so I never even got that far.

Issue #780, opened 2021-08-17
Wrong environment variables (QT_QPA_PLATFORMTHEME=qt5ct not set)


matheusfillipe:
## Problem 

When I launch an application from ulauncher seems that some environment variables are lost.

I have `QT_QPA_PLATFORMTHEME=qt5ct` on my /etc/environment, ~/.profile, ~/.zshrc and tried launching ulauncher with ``QT_QPA_PLATFORMTHEME=qt5ct ulauncher``

Still when I launch ulauncher it is different from launching the application from command line (where the variable is set).

### From ulauncher
![image](https://user-images.githubusercontent.com/24435787/129811929-af71df33-314f-45a8-a6b7-7b70023cfa14.png)

### From Terminal
![image](https://user-images.githubusercontent.com/24435787/129811987-a0bb6bd6-c688-434f-a56d-b1646c6d9855.png)
(The theme and font are very different)

## Environment

I am on qtile window manager, a very minimal and simple window manager. This problem doesn't happen for me (same OS) on KDE plasma.  I am building ulauncher with this little script based on the PKGBUILD:

```sh
build() {
  python setup.py build
  ./ul build-preferences
}

prepare() {
  pkgver=$(git describe --long --tags | sed -r 's/([^-]*-g)/r\1/;s/-/./g')
  sed -i "s/%VERSION%/${pkgver%.*.*}/g" setup.py
  find -iname "*.py" | xargs sed -i 's=\(^#! */usr/bin.*\)python3 *$=\1python='
}

package() {
  export PYTHONHASHSEED=0
  sudo mv build/share/applications/ulauncher.desktop /usr/share/applications
  sudo python setup.py install --root="/" --optimize=1 --skip-build
  sudo rm -rf /usr/share/ulauncher/preferences/{no*,src,bow*,gul*,pack*}
  sudo mv ulauncher.service /usr/lib/systemd/user/
}

prepare
build
package

```

This problem also happened with qtile from aur. `qtile-git`

OS: Arch Linux x86_64, CPU: Intel i7-7700HQ (8) @ 3.800GHz, GPU: NVIDIA GeForce GTX 1050 Ti Mobile, GPU: Intel HD Graphics 630, Memory: 15.5 GiB

friday:
Which Ulauncher version? You're building from the dev branch? We don't have a stable release yet, but in the process of releasing one including #778, which could be causing this?

matheusfillipe:
Yes I am building from dev. 
[...]

Issue #363, opened 2019-03-31
Applications launched under wayland all use Xwayland


beanaroo:
### Steps to reproduce
1. Start ulauncher under wayland session with `GDK_BACKEND=x11`
2. Launch any application

### Current behavior
`GDK_BACKEND` is being inherited by spawned applications.

### Expected behavior
This is undesirable as many applications do support wayland and I would prefer to only use x11 for ulauncher itself as a workaround.

### Environment
**Ulauncher version**: 4.3.2.r8
**Linux distro**: Arch
**Desktop environment**: SwayWM

### Logs
Nothing interesting

beanaroo:
This also has unwanted side effects like searches opening a new (x11) instance of Firefox instead of using the already open (wayland) instance of Firefox

friday:
If you close Ulauncher and open it from the terminal `env GDK_BACKEND=x11` shouldn't be set I think. This also means keyboard binding shouldn't work, but maybe you can add it [this way](https://github.com/Ulauncher/Ulauncher/wiki/Hotkey-In-Wayland).

If this works, you could override `GDK_BACKEND` by copying `/usr/share/applications/ulauncher.desktop` to `~/.local/share/applications/` and removing `env GDK_BACKEND=x11`. If you have Ulauncher set up to autostart, then you also need to remove it from `~/.config/autostart/ulauncher.desktop`.

gornostal:
@friday that solution may not work because Ulauncher starts apps using the same env vars as it was started with.
I'm not sure what would be the best way to fix it. Probably unset `GDK_BACKEND` before launching apps from Ulauncher.

friday:
> Ulauncher starts apps using the same env vars as it was started with.

I suggested starting Ulauncher without `GDK_BACKEND=x11`, so I don't understand how that argument isn't in favor of the workaround I suggested (which still may not work - I haven't tried).

> I'm not sure what would be the best way to fix it. Probably unset GDK_BACKEND before launching apps from Ulauncher.

If you unset it, it won't revert to any user preference it may have been set to (outside of Ulauncher). For example someone might use XWayland for all their desktop apps (likely as a workaround for something else). Then it would still open a new instance of an app rather than a new window right?

[...]

PR #800, opened 2021-09-04
Wayland fix

troycurtisjr:
This fixes ulauncher on wayland when started from systemd. It also makes sure GDK applications launched by ulauncher don't get forced into using Xwayland unnecessarily.

- [x] All tests are passing

Fedora 34, Gnome 40

### DIFF
--- a/ulauncher.service
+++ b/ulauncher.service
@@ -6,7 +6,7 @@ Documentation=https://ulauncher.io/
 Type=simple
 Restart=always
 RestartSec=1
-ExecStart=/usr/bin/ulauncher --hide-window
+ExecStart=env GDK_BACKEND=x11 /usr/bin/ulauncher --hide-window
 
 [Install]
 WantedBy=graphical-session.target
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import subprocess
 import re
 import shlex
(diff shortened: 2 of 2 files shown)


friday:
Very nice! I just tested and this also fixes the use case I couldn't get to work in https://github.com/Ulauncher/Ulauncher/issues/780#issuecomment-902032788

So now if you want Ulauncher to have specific environment variables set you can run

`systemctl --user edit ulauncher`, then add the variable in the designated area:



Then save and run `systemctl --user restart ulauncher`. Then it should work :)

WhyNotHugo:
What's the goal of this change?

The `env GDK_BACKEND=x11` part break running `ulauncher` on wayland; it'll only render via XWayland, and even in that case, render very blurry due to the lack of scaling support on XWaylan.

friday:
> What's the goal of this change?
> 
> The `env GDK_BACKEND=x11` part breaks running `ulauncher` on wayland; it'll only render via XWayland, and even in that case, render very blurry due to the lack of scaling support on XWaylan.

See https://github.com/Ulauncher/Ulauncher/issues/363#issuecomment-484353470

It's possibly that we could be able to run it correctly positioned without XWayland.

WhyNotHugo:
ulauncher is positioned correctly for me without this PR on swaywm (I'm not entirely sure _how_, since the compositor is supposed to handle positioning).

With this PR, however, ulauncher renders very blurry due to XWayland's lack of support for scaling. It'll only really render properly on non-HiDPI displays.

PR #810, opened 2021-09-06
Add ability to show all apps

### LINKED ISSUE #809: Add ability to customize which desktop apps are filtered
### Is your feature request related to a problem? Please describe

Following a discussion on #343 with @friday , the behavior of ulauncher is now valid. It does filter out desktop apps that are not supposed to be shown (based on `OnlyShowIn`/`NotShowIn`).

However for some desktop environments (most probably more simple window managers such as [sway](https://github.com/swaywm/sway), i3, ...), it filters out many apps that could be useful to the environment. For example, some sway users like myself rely on gnome apps which are now filtered out with version >=5.12.0.

### Describe the solution you'd like

As I don't want to break the new (and valid) behavior, but still would like my gnome apps to be shown, a new configuration could make sense.

Something like: `show-apps-in-env=sway;gnome;kde` in `settings.json`. Basically an array (`;` separated) containing the environments (what is set in `OnlyShowIn`) that are whitelisted by the `filter_app` function.

### Additional context
Add any other context about the feature request here.

jtheoof:
This commit adds a configurable flag: `show-all-apps` to display all
apps despite the current desktop environement and the values:

- `OnlyShowIn`
- `NotShowIn`

from the `.desktop` files.

Closes #809

### Checklist
- [x] Update the documentation according to your changes (when applicable) -> Not sure if it needs to more documented than it is (in the settings).

### DIFF
--- a/data/preferences/src/api/fixture.js
+++ b/data/preferences/src/api/fixture.js
@@ -16,6 +16,7 @@ export default function(url, params) {
         clear_previous_text: true,
         grab_mouse_pointer: true,
         blacklisted_desktop_dirs: ['/var/tmp', '/tmp/var/log/bin/bash/root'].join(':'),
+        disable_desktop_filters: false,
         available_themes: [{ text: 'Dark', value: 'dark' }, { text: 'Light', value: 'light' }],
         theme_name: 'light',
         is_wayland: true,
@@ -48,6 +49,9 @@ export default function(url, params) {
       setTimeout(resolve, 0)
     } else if (isMatch(url, '/set/blacklisted-desktop-dirs')) {
       console.log('/set/blacklisted-desktop-dirs', params)
+      setTimeout(resolve, 0)
+    } else if (isMatch(url, '/set/disable-desktop-filters')) {
+      console.log('/set/disable-desktop-filters', params)
       setTimeout(resolve, 0)
     } else if (isMatch(url, '/open/web-url')) {
       console.log('/open/web-url', params)
--- a/data/preferences/src/components/pages/Preferences.vue
+++ b/data/preferences/src/components/pages/Preferences.vue
@@ -146,6 +146,22 @@
           ></editable-text-list>
         </td>
       </tr>
+      <tr>
+        <td class="pull-top">
+          <label for="disable-desktop-filters">Show All Apps</label>
+          <small>
+            <p>
+              Display all applications, even if they are configured to not show in the current desktop environment.
+            </p>
+            <p v-if="disableDesktopFiltersChanged">
+              <i class="fa fa-warning"></i> Restart Ulauncher for this to take effect
+            </p>
+          </small>
+        </td>
+        <td class="pull-top">
+          <b-form-checkbox id="disable-desktop-filters" v-model="disable_desktop_filters"></b-form-checkbox>
+        </td>
+      </tr>
     </table>
   </div>
 </template>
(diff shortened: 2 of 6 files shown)


data/preferences/src/components/pages/Preferences.vue
```diff
@@ -146,6 +146,23 @@
...
+              Ulauncher will display regular applications no matter the value of OnlyShowIn / NotShowIn
+              from the .desktop files.
+            </p>
+            <p v-if="showAllAppsChanged">
+              <i class="fa fa-warning"></i> Restart Ulauncher for this to take effect
+            </p>
+          </small>
+        </td>
+        <td class="pull-top">
+          <b-form-checkbox id="show-all-apps" v-model="show_all_apps"></b-form-checkbox>
```
friday, line 163: I agree with you about the naming, and do want to get the id right from the start rather than having to change and break people's configs. This is why the review took so long (sorry).

How about `disable-desktop-filters` for the id? That's very technical and direct, which I think is good for the id. The description and name shown for users can still be the same. I prefer it this less technical and without the double negatives I suggested for the id. We can change it at any time, should we want but I think you described it really well :+1: .

friday (changes requested):
Nice work :+1:  Sorry for the delay.

data/preferences/src/api/fixture.js
```diff
@@ -16,6 +16,7 @@ export default function(url, params) {
         clear_previous_text: true,
         grab_mouse_pointer: true,
         blacklisted_desktop_dirs: ['/var/tmp', '/tmp/var/log/bin/bash/root'].join(':'),
+        show_all_apps: true,
```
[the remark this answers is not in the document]
troycurtisjr: +1 on defaulting to following the spec and not showing all apps

data/preferences/src/components/pages/Preferences.vue
```diff
@@ -146,6 +146,23 @@
           ></editable-text-list>
         </td>
       </tr>
+      <tr>
+        <td class="pull-top">
+          <label for="show-all-apps">Show All Apps</label>
+          <small>
+            <p>
+              Ulauncher will display regular applications no matter the value of OnlyShowIn / NotShowIn
+              from the .desktop files.
```
troycurtisjr, line 155: A potential wording change:

> Ulauncher will display all applications, even if they are configured to not be shown in the current desktop environment.

(mostly to drop the reference to the OnlyShowIn/NotShowIn parameters, as someone not familiar with .desktop files might not understand the implications.

data/preferences/src/api/fixture.js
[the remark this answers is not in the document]
jtheoof: Good point, will fix that.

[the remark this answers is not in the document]
jtheoof: Fixed

data/preferences/src/components/pages/Preferences.vue
jtheoof: I've renamed pretty much everything to `disable-desktop-filters`.  Let me know if everything is good.

data/preferences/src/components/pages/Preferences.vue
jtheoof: Fixed

jtheoof:
Thanks for the feedback @troycurtisjr @friday . I resolved the comments.

friday (approved):
Looks perfect :+1: Thank you for the the changes :pray:

PR #812, opened 2021-09-10
Ensure child processes are reaped.

### LINKED ISSUE #811: Zombie processes created when launching apps
### Steps to reproduce
1. Run `ulauncher -v` (no extensions)
2. Launch `ulauncher` from within ulauncher
3. `ps aux | grep -i ulauncher`

```
shai       10753  4.0  0.4 1876436 75876 pts/0   Sl+  01:24   0:00 /usr/bin/python3 /usr/bin/ulauncher -v
shai       10791  2.0  0.0      0     0 ?        Zs   01:24   0:00 [ulauncher] <defunct>
```

### Current behavior

The process is launched as a child of the ulauncher process.
When the process terminates, a zombie process is created
This for example breaks `killall --wait` 
I've only noticed it happening recently, so might be something that was introduced in 5.12.x/5.13.x.
Might be related to https://github.com/Ulauncher/Ulauncher/pull/778

### Expected behavior

Should not create zombie processes

### Environment
**Ulauncher version**: 5.13.0
**Linux distro**: Linux Mint 20.2 (Based on Ubuntu 20.04), Kernel 5.14.1
**Desktop environment**: Cinnamon

### Logs
https://gist.github.com/shaicoleman/26c199b81573ca430c43062ca979eee5

troycurtisjr:
Fixes #811.

I did replicate the issue described in #811. Interestingly there was only ever one defunct process, but it would stick around, sometime even chewing up 5% or so CPU until the next process was launched & terminated.  I'm not *super* happy with this solution, as I tend to try to reduce threads typically. However, I didn't come up with a great alternative. Some ideas I thought of:

- A single thread that would wait on any child (`waitpid(0,0)`) in a loop until there were none. Then some kind of singleton control to restart the thread again if needed with the next child. However, we don't want this to reap the extension children, so that doesn't work.
- Using the asyncio lib's subprocess support. Internally this uses an event loop implementation, and isn't something we are using elsewhere. It seems like there was a lot more nuance to get right there, especially with the fact we have the Gtk event loop already running
-  Gio.Subprocess: I strongly considered this implementation, and is fact something I've implemented in my test branch where I've implemented unix sockets (namely to git rid of the thread-per-extension in an effort to help trim down resources). However, that would mean changing to a new launching mechanism and all the change/testing that would entail, and I didn't think it was worth the potentially risky change.

### Checklist
- [X] Update the documentation according to your changes (when applicable)

### DIFF
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -7,12 +7,18 @@ import shutil
 from pathlib import Path
 
 from ulauncher.utils.desktop.reader import read_desktop_file
+from ulauncher.utils.decorator.run_async import run_async
 from ulauncher.utils.Settings import Settings
 from ulauncher.api.shared.action.BaseAction import BaseAction
 
 logger = logging.getLogger(__name__)
 settings = Settings.get_instance()
 hasSystemdRun = bool(shutil.which("systemd-run"))
+
+
+@run_async
+def wait_for_pid(pid):
+    os.waitpid(pid, 0)
 
 
 class LaunchAppAction(BaseAction):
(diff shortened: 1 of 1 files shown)


friday:
I don't think this implementation looks bad at all. If we could find a way not to need it, that would be better, but since we apparently do need it it's good in my book :+1: 

I also reproduced it by launching Ulauncher from Ulauncher. The dead process didn't use that much CPU for me (starting at 1.5% and dropping steadily down to 0.1% after a minute), and the next time I launched any app the ghost process was closed. Very weird behavior.

Gio.Subprocess could be worth looking up. We did switch *from* another GTK API for launching apps though, as it wasn't enough and didn't give up the control we needed.

PR #813, opened 2021-09-13
Use GLib.spawn_async() for app launching to auto-reap children.

troycurtisjr:
This fixes an issue added in my previous PR #812. With that PR, a clean exit from ulauncher isn't possible unless all launched applications are closed. This is because each process had a thread blocked on `os.waitpid()`, which means the threads wouldn't exit until the process they watched exited. For an orderly exit, mostly easily reproduced with a Ctrl-C when running ulauncher from the terminal, things just hanged.

Here I've used GLibs async spawning code because I couldn't find how to ensure children from Python's `subprocess` were reaped without using the waitpid as I was. I considered using `Subprocess.poll` with a sleep, but then I still needed to signal to these threads that they needed to give up, and I needed to make sure the signaling was done in any clean shutdown scenario. This all seemed like overkill since we don't actually want the exit status.

I've tested in all the cases that I could come up with:
- Make sure no zombies happen when the launched processes exit
- Make sure the child processes don't exit with ulauncher exits
-  Works with apps using terminal, with and without a custom terminal command
- Try the above launching ulauncher from the command-line and from systemd

### Checklist
- [-] Update the documentation according to your changes (when applicable)
- [-] Write unit tests for your changes (when applicable)

### DIFF
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -1,24 +1,22 @@
 import logging
 import os
-import subprocess
 import re
 import shlex
 import shutil
 from pathlib import Path
+import gi
+
+gi.require_version("GLib", "2.0")
+# pylint: disable=wrong-import-position
+from gi.repository import GLib
 
 from ulauncher.utils.desktop.reader import read_desktop_file
-from ulauncher.utils.decorator.run_async import run_async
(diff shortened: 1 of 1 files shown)


friday:
Thank you again! Good thing I didn't know what a convoluted weird solution this would end up being for the full launch code, or I might not have started trying fix it.

PR #815, opened 2021-09-14
Call XInitThreads() before any other GUI code initializes.

### LINKED ISSUE #667: "Unknown sequence number while processing queue"
### Steps to reproduce
1. Start ulauncher from the terminal with verbose (-v) flag
2. Keep it running until the toggle command stops working

### Current behavior
See 2

### Expected behavior
No crash :)

### Environment
**Ulauncher version**: 5.9.0
**Linux distro**: Arch
**Desktop environment**: Gnome Desktop (GTK+ 3.24.24)

### Logs
```
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python: xcb_io.c:269: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
[1]    5802 abort (core dumped)  ulauncher -v
➜  ~ [233907:233907:1212/143025.760540:ERROR:shared_context_state.cc(74)] Skia shader compilation error
```

troycurtisjr:
I ran into #667 a few times recently as I was testing various things (normally it is hidden by auto systemd restarts :smile: ), so I decided to dig in a bit. Ultimate I took the direct approach, and just did what the error implies we should do. Call `XInitThreads()`.  It seems that multi-threaded access is fine, it just needs to have the right locks initialized.

I can't reproduce this bug consistently, but my changes have been running well for a while, so it doesn't seem to break anything that I can find. I'm posting this PR a bit early/premature so that other willing folks could test in their own setup to see if anything improves. Hopefully so!

Fixes #667 (hopefully)

### Checklist
- [-] Update the documentation according to your changes (when applicable)
- [-] Write unit tests for your changes (when applicable)

### DIFF
--- a/ulauncher/config.py
+++ b/ulauncher/config.py
@@ -16,6 +16,7 @@ from functools import lru_cache
 from gettext import gettext
 from xdg.BaseDirectory import xdg_config_home, xdg_cache_home, xdg_data_dirs, xdg_data_home
 
+
 DATA_DIR = os.path.join(xdg_data_home, 'ulauncher')
 # Use ulauncher_cache dir because of the WebKit bug
 # https://bugs.webkit.org/show_bug.cgi?id=151646
@@ -59,23 +60,6 @@ def get_data_path():
         raise ProjectPathNotFoundError(abs_data_path)
 
     return abs_data_path
-
-
-def is_wayland():
-    return os.environ.get('XDG_SESSION_TYPE', '').lower() == 'wayland'
-
-
-def is_wayland_compatibility_on():
-    """
-    In this mode user won't be able to set app hotkey via preferences
-    Set hotkey in OS Settings > Devices > Keyboard > Add Hotkey > Command: ulauncher-toggle
-    GDK_BACKEND is typically unset in Wayland sessions to allow GTK apps to self-select
-    """
-    return is_wayland() and (gdk_backend() == '' or gdk_backend().lower().startswith('wayland'))
-
-
-def gdk_backend():
-    return os.environ.get('GDK_BACKEND', '')
 
 
 @lru_cache()
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -90,7 +90,7 @@ def main():
     """
     Main function that starts everything
     """
-    if is_wayland() and gdk_backend().lower() != 'x11' and not is_wayland_compatibility_on():
+    if is_wayland_compatibility_on():
         warn = """
                     [!]
         Looks like you are in Wayland session
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -90,17 +90,6 @@ def main():
     """
     Main function that starts everything
     """
-    if is_wayland_compatibility_on():
-        warn = """
-                    [!]
-        Looks like you are in Wayland session
-        Please run Ulauncher with env var
-        GDK_BACKEND set to 'x11' like this:
-
-        GDK_BACKEND=x11 ulauncher
-        """
-        print(warn, file=sys.stderr)
-        sys.exit(1)
 
     # start DBus loop
     DBusGMainLoop(set_as_default=True)
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -4,6 +4,10 @@ import signal
 import logging
 import time
 from threading import Event
+# This xinit import must happen before any GUI libraries are initialized.
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports,unused-import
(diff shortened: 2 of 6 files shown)


tkapias:
**I confirm that this solves the problem.**

The bug was triggering simply if I flicked the toggle repeatedly every seconds. Now everything works.

PS:  _to the maintainers, please put forward a short documentation for a local build without Docker, I'm on Debian Testing and the only problem was with the AppIndicator3 package. While with Docker everything is longer and messy._

friday:
Wow. This next release is going to be such a major improvement, with this and the HiDPI fix (which was something I kind of figure out looking at improving something else).

I do think this should be wrapped in a condition and not run if `is_wayland` is true. Conditional imports will unfortunately not be appreciated by pylint I think without overriding.

friday:
> PS: _to the maintainers, please put forward a short documentation for a local build without Docker, I'm on Debian Testing and the only problem was with the AppIndicator3 package. While with Docker everything is longer and messy._

This has already been done as part of even more simplifications, but it won't be released until Ulauncher 6.0. We're holding that off to focus on fixing things like this.

troycurtisjr:
Ah yes a conditional is a good idea, though I think is_wayland is true when
running on XWayland, right? I'll look into it this evening to confirm. My
plan would be to put such logic into the xinit.py module. Of course I would
also factor out any common Wayland detection logic somewhere that main and
xinit can both use.

friday:
> Though I think is_wayland is true when running on XWayland, right?

I'm not sure.

Does this fix all these threading issues or do we need to run this code every time we create a thread that will/can call x11 indirectly?

I kind of think it's weird that `x11.XInitThreads()` is in the module, which means it'll always/only run once, right? I would approve this solution as long as it doesn't break other functionality though.

troycurtisjr:
Right, it only needs to be called once per application execution, I'm not
sure if anything bad happens if it is called more than that or not. The
important thing is that it MUST be called before any other X calls are
made. Once I make a determination on the wayland part, then this single
call will be sufficient for any future threads. My understanding is that
the various X calls aren't protected by any mutexes by default, but doing
this indicates you intend to call X from different threads so it
initializes the appropriate locks and enables their use for all future X11
calls.

I also think that expanded explanation may be worth adding to the xinit
module so I will include it in the comment block as well.

troycurtisjr:
OK, so I dug some into the Wayland stuff. One thing I noted is that if you are running in Wayland, calling `XInitThreads()` does no harm, but I suspect it also doesn't do much good. If GDK is running with Wayland backend there are multiple weird behaviors above and beyond the potentially meaningless call to `XInitThreads`. This is probably why we had this:
```
    if is_wayland() and gdk_backend().lower() != 'x11' and not is_wayland_compatibility_on():
        warn = """
                    [!]
        Looks like you are in Wayland session
        Please run Ulauncher with env var
        GDK_BACKEND set to 'x11' like this:
        GDK_BACKEND=x11 ulauncher
        """
        print(warn, file=sys.stderr)
        sys.exit(1)
```
Clearly the intention of this code is the bail if the unsupported Wayland backend is in use. However, there must have been some changes since the original implementation since the results of actual values are a bit surprising. Assuming we are running under Wayland (thus `is_wayland() == True`, the only thing interesting here), the behavior for the existing code for different values of GDK_BACKEND is:

 - `''`: Segfault in GTK. Note this is the var set to the empty string, not the lack of the var
 - `x11`: Launches as you'd expect, with all the right behavior (uses XWayland)
 -  `wayland`: Launches, but the position is not centered, using Wayland backend
 -  `x11,wayland`: Prints out the warning above and bails, incorrectly. if allowed to execute, GDK picks X11, like we want
 - `wayland,x11`: Launches, but the position is not centered, using Waylan backend
 
There are additional permutations that GDK might accept, but don't make much sense on the Linux systems we support.

Basically the only time that code block is actually triggered is a time it definitely shouldn't be!

It seems that we have several places throughout the code which attempts to have some Wayland compatibility. The position oddity makes running under Wayland kinda strange IMO, but maybe that doesn't happen in all environments? Regardless it seems clear that we expect to be able to run with Wayland, warts and all, so I just removed the block entirely. I also factored out the Wayland detection code into a separate module so that xinit.py can use it without worrying about some other X11 related library getting imported in before it has a chance to run.

friday:
My time is a bit limited now but I trust you to handle this :+1:

friday (approved):
Looks good. Ready for merge?

troycurtisjr:
Yep, I'm all done with changes, seems to be working well for me.

Issue #803, opened 2021-09-04
Extension icons should support system icons


friday:
Currently extensions are able to use custom icon images supplied by the extensions, but not system icons that respect the user themes.

https://github.com/friday/ulauncher-clipboard implemented a hacky way to use the system icon for edit-paste, which I think ulauncher should support natively. If we can determine the `icon` argument is not a path, we can assume it's a system icon (ex `edit-paste`).

friday:
Added in v6 branch now

PR #817, opened 2021-09-16
Allow overriding extension icon at the keyword level.

brpaz:
I was just experimenting with this and manage to implement the feature in 1h or so, that´s why I didn´t open an issue before.

This PR allows specifying an icon, associated with a keyword, that will override the global the extension icon defined in the extension manifest when the extension is triggered using that keyword.

The main use case for this feature, is for extensions that implement multiple keywords, with different goals, where the developer might want to use a distinctive icon for each.

For example, on my [docsearch](https://github.com/brpaz/ulauncher-docsearch) extension, which allows users to search on documentation sites that implement [Algolia Docsearch](https://docsearch.algolia.com/), the user can access, for example, the  Vue documentation, using  `docsearch vue <query>` but also I want to have direct keywords for each documentation site like  `vue <query>`. In the latter, I would like to display the specific documentation icon instead of the extension icon.

With the changes in this PR, the only thing I need to do is on the on the `manifest.json` of my extension, define an "icon" property in the keyword preference object like so:

```json
 {
    "id": "kw_vue",
    "type": "keyword",
    "name": "Vue Documentation",
    "default_value": "vuedocs",
    "icon": "images/docs/vue.png"
},
```

When matching the input keyword with the extension preferences, Ulauncher will check if the "icon" property is defined and if yes, use that instead of the extension icon.

The new field is completely optional and should not break anything.

**Notes**

- The VS code configuration change was necessary so I can run unit tests directly from the editor. I can remove it from the commit but since Ulauncher already includes a `.vscode/settings.json` I think it might make sense to be versioned.
- I will update the documentation if you are ok with this feature.

### Checklist

### DIFF
--- a/ulauncher/api/server/ExtensionSearchMode.py
+++ b/ulauncher/api/server/ExtensionSearchMode.py
@@ -55,12 +55,7 @@ class ExtensionSearchMode(BaseSearchMode):
                 if not pref['value']:
                     continue
 
-                if pref['icon'] is not None:
-                    icon = c.manifest.load_icon_from_path(pref['icon'], ExtensionKeywordResultItem.get_icon_size())
-                else:
-                    icon = c.manifest.load_icon(
-                        ExtensionKeywordResultItem.get_icon_size())
-
+                icon = c.manifest.load_icon(ExtensionKeywordResultItem.get_icon_size(), path=pref['icon'])
                 items.append(ExtensionKeywordResultItem(name=html.escape(pref['name']),
                                                         description=html.escape(pref['description']),
                                                         keyword=pref['value'],
--- a/ulauncher/api/server/ExtensionSearchMode.py
+++ b/ulauncher/api/server/ExtensionSearchMode.py
@@ -55,7 +55,12 @@ class ExtensionSearchMode(BaseSearchMode):
                 if not pref['value']:
                     continue
 
-                icon = c.manifest.load_icon(ExtensionKeywordResultItem.get_icon_size())
+                if pref['icon'] is not None:
+                    icon = c.manifest.load_icon_from_path(pref['icon'], ExtensionKeywordResultItem.get_icon_size())
+                else:
+                    icon = c.manifest.load_icon(
+                        ExtensionKeywordResultItem.get_icon_size())
+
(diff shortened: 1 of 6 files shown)


ulauncher/api/server/ExtensionSearchMode.py
```diff
@@ -55,7 +55,12 @@ def get_searchable_items(self):
                 if not pref['value']:
                     continue
 
-                icon = c.manifest.load_icon(ExtensionKeywordResultItem.get_icon_size())
+                if pref['icon'] is not None:
+                    icon = c.manifest.load_icon_from_path(pref['icon'], ExtensionKeywordResultItem.get_icon_size())
+                else:
+                    icon = c.manifest.load_icon(
+                        ExtensionKeywordResultItem.get_icon_size())
```
troycurtisjr, line 62: Might move this back up to the same line as the assignment.

troycurtisjr (approved):
Everything looks good to me. I checked it out:
- [x] Passes all tests
- [x] All existing extensions are happy with the change (no effect)
- [x] Installed @brpaz extension and manually added an icon to the helm keyword. Confirmed it was properly displayed

ulauncher/api/server/ExtensionSearchMode.py
friday: Agreed. These two methods are nearly the same, and the old one is only used in a few places, so I think it's better to refactor it to work with an optional path if needed. Although I'm a bit confused because it seems all you're doing is passing `pref['icon']`. I though that's what the original method defaults to?

friday (changes requested):
I like it, but please merge the new method wit the old one if you can, and also be aware that I won't merge this for a while because it will mean a new API version bump, and I would prefer too keep those minimal, and coupled with major releases if possible.

Also somewhat related: #803

friday:
I had issues working with your docsearch extension so I forked your encoder extension and added two variants: https://github.com/friday/ulauncher-encoder



There's an issue with the icons being svg when rendering the list view, but it's a separate issue (that I would like to fix next).

Pushed some changes and changed the base branch to the v6 branch.

Issue #823, opened 2021-10-04
Switch to damerau levenshtein algorithm for fuzzy matching


friday:
Damerau–Levenshtein is a little bit smarter and probably a little bit slower than regular Levenshtein. Comparing for example "Hello world" with "Hello wolrd", Levenshtein says 2 of the characters need to chance, but Damerau–Levenshtein will say one change is needed, because the characters just need to flip.

https://github.com/jamesturk/jellyfish has this, and I think it's about as well supported on different distros as the Levenstein library. Both are implemented in C and fast.

https://packages.ubuntu.com/hirsute/python3-jellyfish
https://archlinux.org/packages/community/x86_64/python-jellyfish/

friday:
Unfortunately just started relying on functionality specific to Levenshtein ("matching_blocks", which are not available in jellyfish)

Issue #816, opened 2021-09-15
Some search results are not correct for the search query or at least not related to the name


justperfection:
After updating to v5.13.0, Ulauncher doesn't show the search result more related to the app names. Maybe the priority is exec instead of name for the new version.

### Expected behavior

When I hit "B", I expect to see **B**rave browser in the list. I see "Icon Browser", "Disk Usage Analyzer" but not Brave Browser.
Same thing happens for "G" and **G**Musicbrowser.

https://user-images.githubusercontent.com/66198705/133515496-e453e753-8de2-43fd-ab41-96c422e62a4a.mp4

It wasn't like this before the update.

### Environment
**Ulauncher version**: v5.13.0 
**Linux distro**: Fedora 34
**Desktop environment**: GNOME Shell 40.4

### Logs

https://gist.github.com/justperfection/f54c17a9ee31bd4f05d6e17c266ecc63

troycurtisjr:
Tuning these parameters are pretty challenging! I guess "Brave Browser" is a long enough name that the score for a single letter just can't make it to the threshold. Of course lowering the threshold might bring in increasingly unrelated matches. At least in your case though there is still room to put more matches before it reaches the max, so it definitely seems reasonable that it should at least be in the list.

Consistent behavior for single character query strings is probably going to be really challenging to do, especially as we try to score each entry independently. I know this is definitely an area we want to get better in.

friday:
Yeah, I don't think it's possible for the fuzzy matching to ever be "perfect", especially since it has conflicting goals.

The PR that introduced this was actually improving cases like this in general, but it removed a condition that we maybe want to revert, and it needs tuning. I started working on that, but the result actually wasn't satisfactory, so I shelved it.

friday:
This is fixed in the Ulauncher v6 branch and PR #827. It won't be released for a while, but we made a tracking issue if you want to be notified #869

PR #824, opened 2021-10-04
Improve fuzzy matching (again)

friday:
* Restored the condition to return 100% if the match starts with the query (#816)
* Rewrote the custom algorithm. The previous one was actually mis-implemented and didn't work as intended (although it worked well enough in most case).

Also see #823 for further improving this a bit.


friday:
Closing in favor of #827 to save time. Hopefully people can wait.

Issue #833, opened 2021-10-27
Rewrite logic handling applications


friday:
There is a very fast method in `Gio.AppInfo.get_all()` that instantly lists all applications, handling all the logic for us (including checking DESKTOP_DIRS and using the right order, filtering out only applications as defined by having `Type=Application` and ignoring "Hidden" apps, and caching.

We can use this, instead of the custom file watcher using inotify and a separate thread to watch files, and also the separate classes for AppDb, AppIconCache.

I started working on this already and the change will remove something like 500 lines of code without removing any functionality, or sacrifice performance (I think the previous method had more performance issues).

This will also solve #320

PR #835, opened 2021-10-28
Rewrite apps logic using Gio.DesktopAppInfo.get_all()

### LINKED ISSUE #310: App scanner doesn't follow symlinks
**Steps to reproduce:**
Place a shortcut to a folder that contains some desktop configuration files into ~/.local/share/applications/
These applications will immediately become available to default application finder but not to Ulauncher

**Ulauncher version**: v4.3.1r4
**Linux distro**: CentOS
**Desktop environment**: xfce

**Logs**:

### LINKED ISSUE #320: Custom desktop entries in home dir should persist if the global variant updates
**Steps to reproduce:**
1. Copy a desktop entry from `/usr/share/applications/` to `~/.local/share/applications`
2. Add "NoDisplay=true" to the new file.
3. Reinstall/rebuild or upgrade software that recreates the original desktop entry.

**Expected result:**
After step 2, the desktop entry should be excluded from the results, and as long as the custom entry in the home dir remains there it should stay hidden regardless of changes to the original file.

**Result:**
When I upgrade a package with a desktop entry in `/usr/share/applications/`, Ulauncher loads it, even if there's a version in my home dir with the same name. Restarting Ulauncher fixes it.
(I actually had to restart after step 2 as well, but this is a different issue).

It seems the `add_file`-methods in app_watcher.py needs a little more logic to handle this. Didn't add the log since this should happen to everyone.

**Linux distro**: Arch (Antergos)
**Ulauncher version**: 4.3.1.r4-1
**Desktop environment**: Gnome 
**DESKTOP_DIRS**: `['/home/albin/.local/share/applications', '/usr/share/applications']`

friday:
Work in progress implementation for #833

This removes the inotify dependency and file watching completely, as well as several other classes we don't really need (AppDb, AppIconCache). It's both due to using `Gio.DesktopAppInfo.get_all()` now to load the apps, but also for just a lot of refactoring in general

* There's no caching at all now in our logic, but the C-code for the `Gio.DesktopAppInfo.get_all()` API does cache.
* Most of the logic is now in AppResultItem.py, which I consider to be as close as it gets to the "app object". I want to move the LaunchAppAction there as well, but I'm not sure if this class is used by any extensions.
* The code to fetch the apps and filter them is now run for every keystroke, which might be something we want to avoid, but it was easier to implement it like this as there is no event for when the window is shown in the right place. And the performance is snappy as ever.
* There's an incompatibility with the app stat db, so I changed the name of the file to `app_stat_v3`. Previously it used the desktop file path as the identifier, but now it uses the desktop app id (which is what it should have used before, to avoid duplicating apps)
* This will fix #320 as we're no longer handling files directly, and should also fix #310, assuming `Gio.DesktopAppInfo.get_all()` supports symlinks (if it doesn't we should ask users to submit an issue to the maintainers of Gnome/Glib).
* SortedList and SortedCollection were used for the file browse mode as well, but I removed this dependency in a separate commit before this, so they can be removed completely. We really don't need them. It's easier to sort normal python lists in just a few lines of code, and performance isn't an issue here.


ulauncher/search/apps/AppResultItem.py
```diff
@@ -1,42 +1,87 @@
...
+    @staticmethod
+    def from_id(app_id):
+        try:
+            app_info = Gio.DesktopAppInfo.new(app_id)
+            return AppResultItem(app_info)
+        except TypeError:
+            return None
+
+    @staticmethod
+    def search(query, min_score=50, limit=9, apps=Gio.DesktopAppInfo.get_all()):
```
troycurtisjr, line 44: Using `get_all()` as a default argument means it is called at import time, and then never updated. better to set as `None` and then check in the body of the function.

ulauncher/search/apps/AppResultItem.py
```diff
@@ -1,42 +1,87 @@
...
 from ulauncher.api.shared.action.LaunchAppAction import LaunchAppAction
 from ulauncher.api.shared.item.ResultItem import ResultItem
 from ulauncher.search.QueryHistoryDb import QueryHistoryDb
 from ulauncher.search.apps.AppStatDb import AppStatDb
+from ulauncher.utils.image_loader import get_app_icon_pixbuf
+
+settings = Settings.get_instance()
+disable_desktop_filters = settings.get_property('disable-desktop-filters')
+blacklisted_dirs = settings.get_property('blacklisted-desktop-dirs')
+blacklisted_dirs_list = blacklisted_dirs.split(':') if blacklisted_dirs else []
```
troycurtisjr, line 17: The values should probably be read in the body of the class, otherwise they are only set at import time, and never refreshed.

friday: Thank you :+1: Such an obvious one 🤦‍♂️

ulauncher/search/apps/AppResultItem.py
friday: It was inside of the block before but I moved it to a default argument so I would be able to mock it. Was still working on the tests when I ran into other issues.

tests/search/apps/test_AppResultItem.py
```diff
@@ -29,23 +39,32 @@ def app_stat_db(self, mocker):
...
 
-    def test_get_description(self, item):
-        assert item.get_description(Query('q')) == 'Description of TestAppResultItem'
+    def test_search_score(self, app1):
+        assert app1.search_score("true") > app1.search_score("trivago")
 
-    def test_get_icon(self, item):
-        assert item.get_icon() == 'icon123'
+    def test_search(self, app1, app2):
+        searchresults = AppResultItem.search('false', min_score=0, apps=[app1._app_info, app2._app_info])
```
friday, line 56: I know this is kind of ugly (using the _app_info prop to get the appinfo object from the AppResultItem), but the only reason this `apps`parameter exists is so we can make a test like this. And I would prefer to keep the tests ugly rather than make the API harder to use, or adding methods we don't need.

tests/search/apps/mock_desktop_entries/falseapp.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Name=FalseApp
+X-GNOME-FullName=FalseApp - Full Name
+Comment=App 2 comment
+Exec=false
```
friday, line 5: `true` and `false` were picked because Exec needs to actually exist as a command, and they're harmless commands that does nothing, and that everyone has.

ulauncher/search/apps/AppStatDb.py
```diff
@@ -31,16 +28,5 @@ def get_most_frequent(self, limit=5):
...
-        app_db = AppDb.get_instance()
-
-        return [AppResultItem(i) for i in islice(
-            filter(None,
-                   map(lambda r: app_db.get_by_path(r[0]),
-                       sorted(self._records.items(),
-                              key=itemgetter(1),
-                              reverse=True))),
-            0, limit)]
+        top_results = sorted(self._records.items(), key=lambda rec: rec[1], reverse=True)[:limit]
```
troycurtisjr, line 31: The previous code had a `filter`, so we don't think there is the possibility of an empty result any longer?

ulauncher/ui/windows/UlauncherWindow.py
```diff
@@ -106,7 +106,6 @@ def finish_initializing(self, builder):
             # bind in the main thread
             GLib.idle_add(self.bind_hotkey, accel_name)
 
-        start_app_watcher()
```
troycurtisjr, line 109: :partying_face: :partying_face:

ulauncher/search/apps/AppStatDb.py
troycurtisjr: I see that it was moved into the caller, and there it can use `AppResultItem` without an in-line import, which does feel a bit cleaner. The downside is that the limit is applied earlier now, so it is possible that we would have had valid (non-empty results), but the limit might actually elide all the "good" results, leaving only empty results that end up getting trimmed out by the caller. I could see especially during changing install/uninstall and upgrading of systems we could definitely have cases where the stuff in the "front" no longer exists. Then the `get_most_frequent(5)` might give none, or 1 or 2 entries that will actually results in a valid result item, even though we should be able to fill out all 5 if the limit happened later. Ultimately I'm not sure the code cleanliness is worth the UX (potential) impact.

troycurtisjr (changes requested):

ulauncher/search/apps/AppStatDb.py
friday: You are absolutely right about filtering being applied too early. :+1: I changed these methods so much back and forth that I confused myself in the process it seems.

The easiest way to fix this is to remove the limit argument and move [:limit] to the UlauncherWindow class as well, but I think I will move the entire method as a static method in AppResultItem instead.

tests/search/apps/test_AppResultItem.py
```diff
@@ -29,23 +39,32 @@ def app_stat_db(self, mocker):
...
 
-    def test_on_enter(self, item, mocker, query_history, app_stat_db):
+    def test_on_enter(self, app1, mocker, query_history, app_stat_db):
         LaunchAppAction = mocker.patch('ulauncher.search.apps.AppResultItem.LaunchAppAction')
-        assert item.on_enter(Query('query')) is LaunchAppAction.return_value
-        LaunchAppAction.assert_called_with('path/to/desktop_file.desktop')
-        query_history.save_query.assert_called_with('query', 'TestAppResultItem')
-        app_stat_db.inc_count.assert_called_with('path/to/desktop_file.desktop')
+        assert app1.on_enter(Query('query')) is LaunchAppAction.return_value
+        LaunchAppAction.assert_called_with(f'{ENTRIES_DIR}/trueapp.desktop')
```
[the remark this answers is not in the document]
friday: Yeah. Already removed the Ubuntu Xenial build target in the v6 branch, and Bionic was removed accidentally before instead of Xenial, so we can support 3.8 even.

tests/search/apps/test_AppResultItem.py
friday: Pytest and mocking is an area of Python I'm not well acquainted with, but I'll try to do that instead :+1:

ulauncher/search/apps/AppStatDb.py
friday: I moved AppStatDB into AppResultItem. `get_most_frequent` was the only public method used outside of AppResultItem.

tests/search/apps/test_AppResultItem.py
friday: Mocking Gio.DesktopAppInfo.get_all instead in the latest push (only the last commit is new, although the previous ones were rebased over v6).

tests/search/apps/test_AppResultItem.py
```diff
@@ -1,21 +1,41 @@
...
+    @pytest.fixture(autouse=True)
+    def patch_DesktopAppInfo_new(self, mocker):
+        def mkappinfo(app_id):
+            return Gio.DesktopAppInfo.new_from_filename(f'{ENTRIES_DIR}/{app_id}')
+        return mocker.patch("ulauncher.search.apps.AppResultItem.Gio.DesktopAppInfo.new", new=mkappinfo)
+
+    @pytest.fixture(autouse=True)
+    def patch_DesktopAppInfo_get_all(self, mocker):
+        def get_all_appinfo():
+            return map(Gio.DesktopAppInfo.new, ['trueapp.desktop', 'falseapp.desktop'])
```
friday, line 29: I patched `Gio.DesktopAppInfo.new` above. That was `ulauncher.search.apps.AppResultItem.Gio.DesktopAppInfo.new`. However patching a class method just patches it all places the module is used, and the object path you use is just a way to access it? When I tested this it worked andyway, and it was shorter than using `AppResultItem.from_id(...)._app_info`

tests/search/apps/test_AppResultItem.py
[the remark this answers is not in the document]
friday: This code already ran in pytest, which requires 3.6 anyway :), but I actually want to break 3.5 compatibility immediately with this branch, so no one says it's a regression if we do it later.

PR #839, opened 2021-11-07
Added build support for Ubuntu Impish and Bionic

gornostal:
Added build support for Ubuntu Impish and Bionic

### DIFF
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -95,10 +95,10 @@ launchpad_upload() {
     else
         PPA="agornostal/ulauncher"
     fi
-    xenial="PPA=$PPA RELEASE=xenial ./ul build-deb $VERSION --upload"
-    hirsute="PPA=$PPA RELEASE=hirsute ./ul build-deb $VERSION --upload"
+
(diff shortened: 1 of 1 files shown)


gornostal:
I've merged this commit to master too

friday:
> I've merged this commit to master too

I stopped using master, because it sort of had no purpose, and we have a v6 branch for the most active development currently (removed Xenial build target there as well, but didn't add back Bionic or add Impish).

Bionic is already unsupported by Ubuntu, apart from security updates. Is the aim to support the same as the "Security Support" column here? https://endoflife.date/ubuntu

Regarding master, we could remove it, or use it to sync with releases. Initially I did the latter, but it didn't really serve any cause other than to have it updated.

gornostal:
Then we sold probably remove master branch.

Bionic's end of support is April 2023. https://wiki.ubuntu.com/Releases
Or am I missing something?

friday:
Must be endoflife.date that has the info wrong then. Strange that they can get it so wrong. I found that recently when searching for that Ubuntu page, and have been using it since since it's much easier to read.

Issue #834, opened 2021-10-27
Remove "Blacklisted App Dirs" option


friday:
"Blacklisted App Dirs" is a custom Ulauncher setting and not part of the desktop specification. The desktop specification allows many ways to hide apps already, and "Blacklisted App Dirs" was implemented before supporting some of these (#343).

In the original issue that lead to creating this setting (#136) it is described like this:
> In my case, on Linux Mint, I have 210 .desktop files in /usr/share/applications/screensavers. All these shortcuts appear in Ulauncher and I often launch some of them given their names looks a lot like apps I use most :sweat_smile:

Could also find another person having this issue https://askubuntu.com/q/1096973/420716 and it seems to be Xscreensaver.

Besides that path there are other paths in the default blacklisted dirs:

![2021-10-27_16-03](https://user-images.githubusercontent.com/515120/139081637-f81a2090-572c-4693-b64f-71d8907b3db1.png)

* /usr/share/app-install is an Ubuntu Software Center thing. Ubuntu Software Center doesn't exist any more
* /usr/share/mimelnk is for storing mimetypes. These are desktop entries but `Type` is set to "MimeType", so they will be filtered out by Ulauncher anyway.
* All the other ones seem to be related to performance in KDE, which loves to implement customization via desktop entries apparently #130. They do not have `Type` set to "Application". More info: https://askubuntu.com/questions/1179100/add-item-into-context-menu-in-dolphin

After implementing #833 this option won't make much sense any more for the performance cases. Only if it still happens with Xscreensaver. I think version 6.0 would be a good time to try to drop this setting and see if users complain. It's easy to restore, should we need to.

iago-lito:
> I think version 6.0 would be a good time to try to drop this setting and see if users complain.

Hi, maybe I'll be that user @friday ? ^ ^" I use ulauncher a lot with custom shortcuts, but not for scrolling the list of apps installed on my system. Now that the list of apps shows up, it clutters my search results with a lot of unwanted items.

I actually wasn't a user of the "blacklist" feature, so maybe I am not exactly asking to reintroduce it. What I'm after is rather some default box like:
- [x] Search system applications.

.. that I would personally untick. Or alternately a default box like:

- [ ] Only search among custom shortcuts.

.. that I would personally tick?

friday:
@iago-lito Your use case is indeed not what this blacklist was intended to solve, so it would be hack/bad/unintuitive to use it for that imo. I created #1246

PR #847, opened 2021-11-08
Remove Blacklisted Dirs settings and EditableTextList component it used

friday:
Implements #834


friday:
I'm going to merge this immediately, but I wanted it to be a PR for visibility. We can always restore this is if we need, but I really don't think we do. It would be if we wanted to add other lists of strings to the settings somewhere I think. If users get issues because of removing this setting and they have some kind of odd setup for themselves, then it's kind of on them. If multiple users get issues with the same offending directory we can hard code ignoring it, without needing the setting.

PR #850, opened 2021-11-10
Add fallback solution for searching if Levenshtein is missing or outdated

friday:
This PR adds a fallback for Levenshtein if it's missing or outdated. I added it because I knew about the other method (it's how I originally implemented `get_matching_blocks()`), and it doesn't add a lot of code.

I don't think we strictly need this PR now, but we would need it for Xenial support (which we are pulling anyway with 6.0.0, and we stopped making Xenial releases for 5.14.1 also). I do think it's good to have in though, as the Levenshtein library has been in a sorry state for many years, and I'm not sure if it will continue to be packaged for distros.

Not sure this means we should:
* Make the tests run for both implementations?
* Remove the [CI server pypi package installation](https://github.com/Ulauncher/Ulauncher/pull/822/commits/6dfd4bc1e131a28a9b09fb6ab47409c388c543cf)
* Mark the Levenstein library as an optional dependency?

### DIFF
--- a/ulauncher/utils/fuzzy_search.py
+++ b/ulauncher/utils/fuzzy_search.py
@@ -1,17 +1,31 @@
 from functools import lru_cache
-from Levenshtein import matching_blocks, editops
+from logging import getLogger
+from difflib import SequenceMatcher
+
+logger = getLogger(__name__)
+
+
+# Using Levenshtein is ~10x faster, but some older distro releases might not package Levenshtein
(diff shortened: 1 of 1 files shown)


troycurtisjr (approved):

troycurtisjr:
Ideally I'd think we would keep the install, and run with both for tests.

friday:
I'll merge it because I already have other code continuing from this state. I would have added tests for both if I understood how to, but I can add that to the backlog. I can mock it to behave any way, but never both, and can't seem to use parametrized tests for this.

Issue #842, opened 2021-11-07
ULauncher not working after latest upgrade


antonio-petricca:
Hi, I just applied latest ULuancher upgrade, it does not start any applications.

Here a log of the error:

> 2021-11-07 16:24:31,502 | INFO | ulauncher.api.shared.action.LaunchAppAction: run() | Run application Text Editor (/usr/share/applications/xed.desktop) Exec xed %U
Job for run-r5d4df67bf6a24568a6960ad89d9b1105.scope failed.
See "systemctl status run-r5d4df67bf6a24568a6960ad89d9b1105.scope" and "journalctl -xe" for det

troycurtisjr:
Can you follow the issue template so that we have all the necessary information to try to help you troubleshoot?

antonio-petricca:
Sorry, here are all the details:

### Steps to reproduce

1. Upgrade to latest release.
2. try to start any application

### Current behavior

It starts no more applications.

### Expected behavior

It must run the selection application.

### Environment
**Ulauncher version**: 5.14.0-0ubuntu1ppa1~bionic 
**Linux distro**: Linux Mint 19.3 (Ubuntu 18.04)
**Desktop environment**: MATE

### Logs

> 2021-11-07 16:24:31,502 | INFO | ulauncher.api.shared.action.LaunchAppAction: run() | Run application Text Editor (/usr/share/applications/xed.desktop) Exec xed %U
Job for run-r5d4df67bf6a24568a6960ad89d9b1105.scope failed.
See "systemctl status run-r5d4df67bf6a24568a6960ad89d9b1105.scope" and "journalctl -xe" for det

antonio-petricca:
Some other logs:

> 2021-11-07 23:11:58,261 | INFO | ulauncher.api.shared.action.LaunchAppAction: run() | Run application Vivaldi (/usr/share/applications/vivaldi-stable.desktop) Exec /usr/bin/vivaldi-stable %U
Job for run-r98cfa415db314d6787bd1602aa523a70.scope failed.

troycurtisjr:
We saw something similar with someone running an alternate arch Kali distro (running on Android). Are you running on standard x86 hardware? I will try to reproduce in a VM.

antonio-petricca:
I runs ULauncher since 2 or 3 years on Delle E5495 equipped with an Amd Ryzen 7 PRO, 16Gb RAM, 512Gb SSD.

[...]

PR #852, opened 2021-11-11
Pass environment to systemd-run with multiple --setenv arguments

friday:
Addresses #780

I think this is a bit ugly, but it works. It feels redundant to pass the environment twice, but I think we need to if systemd-run is not picking it up. I don't know strictly why we need to pass it to GLib.spawn_async, but this was added by @troycurtisjr adressing #811

(Hoping to be able to remove this whole convoluted solution at some point, but if the app runs as a systemd service we need to keep it)

### DIFF
--- a/ulauncher/api/shared/action/LaunchAppAction.py
+++ b/ulauncher/api/shared/action/LaunchAppAction.py
@@ -42,6 +42,10 @@ class LaunchAppAction(BaseAction):
             # strip field codes %f, %F, %u, %U, etc
             sanitized_exec = re.sub(r'\%[uUfFdDnNickvm]', '', exec).rstrip()
             terminal_exec = shlex.split(settings.get_property('terminal-command'))
+            env = dict(os.environ.items())
+            # Make sure GDK apps aren't forced to use x11 on wayland due to ulauncher's need to run
+            # under X11 for proper centering.
+            env.pop("GDK_BACKEND", None)
             if app.get_boolean('Terminal'):
                 if terminal_exec:
                     logger.info('Will run command in preferred terminal (%s)', terminal_exec)
@@ -51,16 +55,12 @@ class LaunchAppAction(BaseAction):
             else:
                 sanitized_exec = shlex.split(sanitized_exec)
             if hasSystemdRun and not app.get_boolean('X-Ulauncher-Inherit-Scope'):
+                setenv = ["--setenv={}={}".format(k, v) for k, v in env.items()]
                 sanitized_exec = [
                     'systemd-run',
                     '--user',
-                    '--remain-after-exit',
-                ] + sanitized_exec
-
-            env = dict(os.environ.items())
-            # Make sure GDK apps aren't forced to use x11 on wayland due to ulauncher's need to run
-            # under X11 for proper centering.
-            env.pop("GDK_BACKEND", None)
+                    '--remain-after-exit'
+                ] + setenv + sanitized_exec
 
             try:
                 logger.info('Run application %s (%s) Exec %s', app.get_name(), self.filename, exec)


ulauncher/api/shared/action/LaunchAppAction.py
```diff
@@ -51,16 +55,12 @@ def run(self):
...
                     '--user',
-                    '--remain-after-exit',
-                ] + sanitized_exec
-
-            env = dict(os.environ.items())
-            # Make sure GDK apps aren't forced to use x11 on wayland due to ulauncher's need to run
-            # under X11 for proper centering.
-            env.pop("GDK_BACKEND", None)
+                    '--remain-after-exit'
+                ] + setenv + sanitized_exec
```
troycurtisjr, line 63: It does seems like we are just piling on the callers, and each tweak seems to break a different mode. I wish there was an easy way to write some integration tests for these issues!

Looking at the systemd-run doc, I see the running as a transient scope (using `--scope`) inherits the execution environment, and systemd-run doesn't exit until the command exits. This explains why it worked to pass the env vars through `spawn_async`. However, without the `--scope` it executes in a transient "service", in a fresh environment, which is why the env vars are busted now. So the spawn_async call doesn't need to have the env vars if we use `systemd-run` and if `-E`, but of course they *are* required if a system is configured to use the terminal or doesn't have the systemd-run executable.

Dropping `--scope` means that new transient services are created for each execution, and as far as I can tell, never cleaned up:
```
  run-r34df63b471104419bb191529b3160d6b.service       loaded active exited  /usr/bin/alacritty
  run-r3f7592536bdb48c58c57ba1078ee0317.service       loaded active exited  /usr/bin/gpxsee
  run-r4949f4505d24432294fdfce423283f83.service       loaded active running /usr/bin/alacritty
● run-r4f80efc235104561889d0dfaa59457fb.service       loaded failed failed  /usr/bin/git-dag --prompt
  run-r9e503ce5f19442b38715618f9308d75e.service       loaded active running /usr/bin/alacritty
```

This seems like for a long running system this could get pretty big. Sure we are doing something wrong is there is the "right" way to do this?

darkfishy: These were my thoughts exactly, when I was going through the docs. However, what if we use `--scope`, remove `--remain-after-exit` (it doesn't work with `--scope`) and continue using `GLib.spawn_async`. In my limited testing, apps start with the inherited environment as expected and continue to execute even after stopping `ulauncher` itself.

friday: @darkfishy Wow. I'm surprised that works! systemd-run is needed specifically for launching apps detached from the main process group when Ulauncher is running as a service with systemd. When I kill ulauncher the apps remain though, so it works I think, and it's less of a convoluted weird solution.

Could you test this also @Jaakkonen? Replacing `--remain-after-exit` with `--scope`? It should work right if the app launches. And I would love some input from @troycurtisjr also before we go with this and push another patch release?

troycurtisjr: Isn't using scope what we changed to address some of the errors people had when not running ulauncher with systemd? https://github.com/Ulauncher/Ulauncher/commit/93c68de2b4c994840e577bb5a3aa5a5659da287b

Is there a way to tell when we are running in our own systemd scope/service? Perhaps we only need to do the systemd-run truck in that case?

darkfishy: I reckon the issue #842, which commit 93c68de2b4c994840e577bb5a3aa5a5659da287b addresses, could have been because of resource exhaustion and eventual starvation for new processes, but not necessarily due to the way `ulauncher` was calling `systemd-run`. More specifically, looking at the https://github.com/Ulauncher/Ulauncher/issues/842#issuecomment-963927467 in that issue, one of the errors is that `No PIDs left to attach`.  This could have been due to either system level `pid.max` or some stringent limit on `pid.max` up the `slice` hierarchy getting exhausted.

Without slicing, the new scope/service is placed under the default `app.slice` which may have different resource limits (generally max).

It also seems like `systemd` [encourages](https://systemd.io/DESKTOP_ENVIRONMENTS/#xdg-standardization-for-applications) use of `.service` over `.scope`, though then it means using the `--setenv` hack above.

On a different note, was there a reason behind using `--remain-after-exit`? As I understand, this keeps the transient process loaded and active even after the underlying process has exited.

friday: > Isn't using scope what we changed to address some of the errors people had when not running ulauncher with systemd? [93c68de](https://github.com/Ulauncher/Ulauncher/commit/93c68de2b4c994840e577bb5a3aa5a5659da287b)

--scope was only removed because it's incompatible with --remain-after-exit

> Is there a way to tell when we are running in our own systemd scope/service? Perhaps we only need to do the systemd-run truck in that case?

No. Removing --scope is what caused the issue this PR is trying to solve.

friday:
I made a new PR at #853. I think it's the best we can do for now. If using only `--scope` works for all, then we won't need the warning, but it's not like anyone will actually see it unless they are experiencing issues.

I like systemd personally, but I think we can't recommend it to Ubuntu users in particular.

PR #853, opened 2021-11-12
Restore --scope argument but only use systemd-run if running from within a systemd unit

friday:
Also, prints a warning each time recommending people not to use systemd if it break for them (assuming they read the log with journalctl)


troycurtisjr (approved):
Looks good to me. I tried both with and without running under systemd.

PR #870, opened 2021-12-02
Add event.get_keyword_id()

friday:
This is exactly the same solution as before, but fixing the tests and documentation.

I was too confused about the existing tests to understand where to add tests for this.

It can be tested using this forked extension that implements it:
https://github.com/friday/ulauncher-encoder

### Checklist


friday:
Going to just merge this and if we need to change it before releasing v6 that's fine, but we need this in the API. The current way is broken and this method is the solution.

PR #916, opened 2021-12-06
Update Google and Stackoverflow icons

cycool29:
The icons of Google and Stackoverflow are very old, so I replaced them with the official one.

### Checklist


friday:
Thanks.
Probably won't be released for a while. Targeting a new major release at this point, but we will work this change into that branch before releasing it.

PR #917, opened 2021-12-07
Move "ulauncher.api.server" to "modes.extensions"

friday:
`ulauncher.api.server` is actually the server "mode", that is imported by Ulauncher and runs extensions

It is true that this includes the extension "server", but it's all part of the ExtensionMode anyway, and it's internal code which extensions should not import or see. So it's bad to have it under the `ulauncher.api` namespace, because that's what we encourage extensions to import.

I have already ran into issues with extensions importing internal parts of our code, which is the first reason for this, but I actually never understood why this directory wasn't placed consistently with the other modes to begin with.

### Checklist


friday:
@troycurtisjr I'm going to merge this straight away (ran the tests and the app and it all works). I hope you are not working on something in these parts that will conflict. But wanted to make you aware in that case.

Issue #920, opened 2021-12-07
Make a clear distinction between Ulauncher API and internals


friday:
### Description

Currently extensions can import any part of Ulauncher. This means **any time we make any internal changes to our code it has the potential to break extensions**. Even if we only change internals not meant to be used by extensions. It also hogs up ram for no good reason, but this is secondary.

1. The Ulauncher API should be limited to `ulauncher.api`, with fatal errors if importing from another Ulauncher module [like this](https://github.com/imsakg/ulauncher-translator/blob/824037cd06fb56162c0d629020bdc4b70ce3d77e/main.py#L15).
2. `ulauncher.api` should be flattened into one directory, merging `ulauncher.api.client` and `ulauncher.api.shared`, but keeping backwards compatibility (`ulauncher.api.server` has been moves as of #917).
3. Any other part of `ulauncher.api` which is not meant to be consumed by extensions should be moved to `ulauncher.modes.extensions`
4. `ulauncher.api` should be self-contained if possible and not import Ulauncher internals or Glib (like `RenderResultListAction` in particular, but also other "actions" like `SetUserQueryAction` and `CopyToClipboardAction`) This is a big part of #285. Maybe it can import from `ulauncher.utils` though. If `ulauncher.api` is not self-contained` and has to import internal Ulauncher modules like now, __this means users have to import the ulauncher internals also to use types__, and the API documentation will "leak" Ulauncher internals, which is confusing for users. __We can't blame devs for using modules not meant for API usage__ until we fix this, because it's really our fault.
5. Actions and ResultItems don't need to be extended and have all these variants. All we need from them is a way for the extension to tell the app what to do (ex `pass this data to the "Copy" action`), but instead we now import the whole copy action, and all it's dependencies in the extension runtime, attach the data to be copied, serialize it with pickle, stream it with sockets to the server, deserialize it and run it on the other side.

### Enhancement isn't already requested

- [X] I have checked the existing Enhancement requests before suggesting this

gornostal:
The first point is very interesting and will definitely be helpful to make sure only `api` module is used. I see how this can be implemented, although the approach will look more like a workaround. Basically, all non-`ulauncher.api` imports should register some global variable that will be read from `api.client.Extension.run()`. If the variable is set, the runner logs an error and does `sys.exit(1)`.

Issue #864, opened 2021-11-21
Deprecate support for bad extension APIs and behavior


friday:
My three main concerns are
1. The query listener should only render a result list. It can trigger any action now, rather than just RenderResultListAction. I think instead of letting users import and return a `RenderResultListAction`, I think it's better they return a `List[ResultItem]`, which is the input argument RenderResultListAction takes, so less code for the extensions.
2. `event.get_keyword()` is misimplemented and unsafe, see #284
3. Alt + enter is also misimplemented imo. It should either always show a list, or execute the first action from a list, as per #326

I think it's best introduce alternatives ways, then show a deprecation warning for these, to give extension devs an upgrade window.

friday:
All of these are resolved now. The last part was https://github.com/Ulauncher/Ulauncher/pull/1774

Issue #285, opened 2018-09-07
Extension idle ram usage


friday:
I may be measuring this wrong, but it seems that every extension uses at least 16mb of ram while idle (maybe this is the Python runtime and the websocket client). This builds up.

**Ulauncher version**: 4.1.1.r1
**Linux distro**: Arch
**Desktop environment**: Gnome

![screenshot from 2018-09-07 18-16-32](https://user-images.githubusercontent.com/515120/45230767-3ceb6d80-b2ca-11e8-9c0b-20af6f48e53e.png)

goodwillcoding:
@friday right, it's the python interpreter run per plugin ... using ram if has to. completely normal usage. it's is not leaking memory as far as I can tell. how is this an issue?

friday:
> No memory leaks as far as I can tell.

I never said it was. I said "This builds up". This is referring to extensions using 16mb ram each, so if you have a lot of extensions this builds up. Over 200mb with 10 extensions, and close to 400mb with 20, 700mb with 40 etc (doubt anyone has this many but my point is this doesn't scale well). 

> It's the python interpreter run per plugin ... using ram if has to. completely normal usage.

> how is this an issue? 

> still not an issue, just standard python process running.

I don't understand what point you are trying to make. Allocating ram comes with a cost. There are multiple possible solutions to allocate less ram. How is "this is normal" or "this is not an issue" relevant or constructive arguments? How much does it need to allocate before it becomes an issue?

* Aprox 11/16mb (see footnote 1) allocated per extension seem to be due to the Python Gnome API's (GLib, Gtk and Gdk). Architecturally, they don't need to be loaded into each extension. Any point where extensions need to call them this should optimally be done by sending a message to the server. Duplicating the API is inefficient. It's also fair to assume it's a blocker for #283 if that issue is under consideration.
* I have yet to see any extension that actually runs background jobs, so I don't understand why extensions are kept running after the Ulauncher UI is closed. Extensions may depend on other API's and services doing the heavy lifting, but rarely does much more than wrap their APIs. It could be just a missed opportunity rather than intentional design, or it's a performance benefit. At least on my system running a small Python script with a basic websocket implementation (see footnote 2) takes less than 50ms. So I don't really see the need. Maybe I'm missing something?

This issue may not be a low-hanging fruit, but well worth considering for long term API-breaking improvements imo. If you disagree, you have to argue better.

Footnotes:
[...]

PR #923, opened 2021-12-09
Support list as a shortcut/alias to RenderResultListAction

friday:
Instead of this

```py
from ulauncher.api.shared.action.RenderResultListAction import RenderResultListAction
```
...and returning `RenderResultListAction(list_of_items)`, you can now just return `list_of_items`.

This applies to all places RenderResultListAction was previously used, including the modes `handle_query` method, the extension query listener and and the extensions `on_enter` and `on_alt_enter` method.

For the extension methods this change is backward compatible, but as step two to this (in a future major release) we should remove `RenderResultListAction` completely as an action and only support returning lists. Primarily because this API imports a lot of internals and it's problematic to expose it as part of the API #920 #285, and it's also better to restrict the extension query listener to _only_ accept this instead of any action, so that users get a more consistent and reliable experience (not something opening or the Ulauncher window hiding when they're typing the query) and extension developers get better guidelines to provide this. #864

### Checklist

### DIFF
--- a/docs/diagram-source.txt
+++ b/docs/diagram-source.txt
@@ -7,7 +7,7 @@ note over Ulauncher: extracts keyword\nand checks if any extension\ncan handle i
 Ulauncher->Extension: sends KeywordQueryEvent
 note over Extension: on_event method of \nKeywordQueryEventListener\nis triggered
 note over Extension: handles event by preparing\na list of result items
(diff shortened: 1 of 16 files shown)


gornostal:
I agree with your point about the consistency of user experience!
And I like that you updating tests too :slightly_smiling_face:  :+1:

PR #938, opened 2021-12-15
Search now also searches app descriptions, albeit with lower priority

### LINKED ISSUE #613: Can I recognize and search even "comment" in ~/.local/share/applications/*.Desktop File?

BharatKalluri:
Partly solves #270 . Also closes #613. 

Previously discussed on #652 

### Checklist
- [X] Update the documentation according to your changes (when applicable)

### DIFF
--- a/ulauncher/modes/apps/AppResult.py
+++ b/ulauncher/modes/apps/AppResult.py
@@ -75,10 +75,11 @@ class AppResult(Result):
             return 0
 
         # Also use the executable name, such as "baobab" or "nautilus", but score that lower
-        return max(
+        return max([
             get_score(query, self._name),
-            get_score(query, self._executable) * .8
-        )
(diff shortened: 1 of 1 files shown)


friday (approved):
It works good in my brief tests. I think we can just add it like this and if turns out we need to lower or raise the weight value we will probably find that out.

[`max()`](https://docs.python.org/3/library/functions.html#max) works with either a list or just an unlimited list of arguments as long as it's at least two. So you didn't need to add the brackets, but it also doesn't hurt.

Issue #947, opened 2021-12-30
Error building docker image in v6 branch


the-allanc:
### Environment

Ubuntu 20.04

### Bug description

Ran `docker build` on my machine so that I could then make code changes, but it fails on step 8 of the build. The output ends with
```
yarn install v1.22.17
info No lockfile found.
[1/5] Validating package.json...
[2/5] Resolving packages...
warning bootstrap-vue > popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1
warning webpack-hot-middleware > querystring@0.2.1: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
[3/5] Fetching packages...
error @npmcli/fs@1.1.0: The engine "node" is incompatible with this module. Expected version "^12.13.0 || ^14.15.0 || >=16". Got "10.24.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command '/bin/sh -c pip3 install -r requirements.txt &&     pip3 install -r docs/requirements.txt &&     yarn &&     mv node_modules /var' returned a non-zero code: 1
```

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines.

### Not an extension issue

- [X] This issue applies to the Ulauncher application, not a specific extension.

### Latest version

- [X] I am using the latest Ulauncher release.

friday:
You don't need Docker for v6, but if you want to build it, you can check out the dev branch, build and check out v6 again.

the-allanc:
> You don't need Docker for v6, but if you want to build it, you can check out the dev branch, build and check out v6 again.
I did try that, but Docker builds are deterministic, so that didn't really make a difference.

Regarding the motivation of why I'm trying it - when I was making other changes, the guidelines for pull requests mentioned that tests would be passing, and then when I tried to run tests locally, I got the following notice:

```WARNING: It's recommended to run tests in a docker container to be sure they will also pass in CI```

I don't mind either way whether I should be using Docker or not, but if we're encouraged to write tests, and the test command recommends us to use Docker, then it probably should work. :grinning:

friday:
Oh yeah. I forgot about that. Read and ignored it so many times.

friday:
[...]

Issue #721, opened 2021-03-28
Isolation for extensions


gaborbernat:
Currently, extensions require the user to install their own dependency via ``pip install --user``, added via https://github.com/Ulauncher/Ulauncher/issues/541. This is a bad idea. Because installing dependencies not managed by host OS can mean you're installing conflicting dependencies and break your host OS python applications. Also, extension A and B might have conflicting dependencies. Can we instead make sure every dependency has its own virtual environment, and dependencies are installed into that virtual environment?

friday:
This was never meant to be the final solution I think. See the discussion in #273.

friday:
We have now implemented automatic dependency installation (#1461). We didn't go the virtual environment route, but if we find this approach lacking without it, we will consider it. If anyone wants to make a case for that they are welcome to open a new issue for it.

In the meanwhile I'm closing this. 

And thanks for all your awesome contributions to the Python community Bernát 🙏

PR #948, opened 2021-12-30
Add pip-run support for extensions

the-allanc:
After discovering ulauncher a couple of days ago, and experimenting with plugins, I've created a small change which would allow extensions to indicate what third-party dependencies they would need using [pip-run](https://github.com/jaraco/pip-run#script-runner).

By adding a `__requires__` definition within `main.py`, a script can indicate which dependencies it needs to be installed. These dependencies are then installed in a transient virtualenv from which the script can run from.

My changes assumes that `pip-run` is available somewhere on the system path, and only attempts to run a script using `pip-run` if indicated in the manifest.

One slight issue is that this adds some overhead (even if the libraries are cached locally), and there is enough of a delay (at least with testing locally) between Ulauncher executing the extension and the extension being ready that you often won't get any options being presented during the first invocation.

I've had to make these changes against the dev branch rather than the v6 branch, as I've been struggling to run the v6 branch locally (I've filed #947 in relation to this).

Anyway, interested in what your thoughts are as a proof-of-concept.

### DIFF
--- a/ulauncher/modes/extensions/ExtensionManifest.py
+++ b/ulauncher/modes/extensions/ExtensionManifest.py
@@ -37,6 +37,7 @@ ManifestJson = TypedDict('ManifestJson', {
     'description': str,
     'developer_name': str,
     'icon': str,
+    'use_pip_run': bool,
     'options': Optional[Options],
     'preferences': List[ManifestPreferenceItem]
 })
@@ -81,6 +82,9 @@ class ExtensionManifest:
 
     def get_developer_name(self) -> str:
         return self.manifest['developer_name']
+
+    def get_use_pip_run(self) -> bool:
+        return self.manifest.get('use_pip_run', False)
 
     def get_preferences(self) -> List[ManifestPreferenceItem]:
         return self.manifest.get('preferences', [])
--- a/ulauncher/modes/extensions/ExtensionRunner.py
+++ b/ulauncher/modes/extensions/ExtensionRunner.py
@@ -85,7 +85,9 @@ class ExtensionRunner:
         manifest.validate()
         manifest.check_compatibility()
 
-        cmd = [sys.executable, os.path.join(self.extensions_dir, extension_id, 'main.py')]
+        cmd = ['pip-run', '--'] if manifest.get_use_pip_run() else [sys.executable]
+        cmd.extend([os.path.join(self.extensions_dir, extension_id, 'main.py')])
+
         env = {}
         env['PYTHONPATH'] = ':'.join(filter(bool, [ULAUNCHER_APP_DIR, os.getenv('PYTHONPATH')]))
 


friday:
This relates to #273 and #721. We haven't decided on how to solve them yet, but I'm not as knowledgeable about venvs and similar techniques as Troy so I think it's better he has a look at it. We have discussed namespace packages as the solution ifnot using the proposal in #721.

As mentioned in the other issue, you don't need Docker any more to contribute. It's easier now. In fact I'm not sure where you found the references to Docker installation as they were not in the v6 branch and removed recently from the dev branch also.

the-allanc:
Finally got the v6 branch running locally, and so I've updated my changes against that branch.

the-allanc:
> This relates to #273 and #721. We haven't decided on how to solve them yet, but I'm not as knowledgeable about venvs and similar techniques as Troy so I think it's better he has a look at it. We have discussed namespace packages as the solution ifnot using the proposal in #721.

Just some reasons why I advocate for `pip-run` as a solution:
1. Having a virtualenv created on demand (and deleted afterwards) saves any hassle of trying to maintain that environment.
2. Being able to write single file scripts which can also state their dependencies makes the script much more convenient - although extensions require multiple files anyway, so that's not as much of a selling point here.

troycurtisjr:
I'll take a close look this weekend, it seems really interesting, I've not come across pip-run before.

friday:
One big issue I can see with this is that pip-run isn't supported as a dependency in distros, while virtualenv is. pip-run is just a convenient wrapper for virtualenv, so I think we can do the same thing as pip-run does without needing pip-run.

If I understand things right it also seem that every time Ulauncher would launch, it would download and install the same extension dependencies, install the extensions to a temporary location and then afterwards delete them? I think this would slow down the initial start of Ulauncher by a lot if this is the case.

the-allanc:
pip-run would need to be installed by a user manually, yes - but then it would be the only dependency that a user would need to installed.

Ulauncher could be in charge of creating environments, but pip-run tears them down after use, so you would have to work out either a similar strategy or if they were persisted, then you would have to determine how they are maintained if dependencies change.

Regarding how the code currently works - the overhead of initialising the environment is in the spawned process for the plugin. So currently in this branch, Ulauncher starts up just fine, but when the plugin is initialised, it usually isn't ready quickly enough to respond to the first keyword query event.

To counter that, it might be suitable (either as a preference or just default behaviour by Ulauncher) to start the plugin before the user attempts to use it; that way, a user shouldn't notice the overhead.

troycurtisjr:
I haven't had the chance yet to really dig into the PR itself, but from the
description of the pip-run implementation I doubt that this specific
approach will be the right way to go. The overall idea is on the right
track, but I don't think it's reasonable to have users install, and
presumably need to keep updated, pip-run when ulauncher itself is available
in a distro's package manager. Individual extensions may require manual
installation now, which is not ideal obviously, but at least if they
install using their package manager it will then be kept up-to-date
automatically. Additionally for me personally having the churn of each
environment created and torn down every invocation is pretty much a
deal-breaker as a user, which means I wouldn't want to pass it on to
general users. It would also make extension development pretty painful. I
presume this also needs an Internet connection at runtime to work? I see
that it relies on the pip cache to not require the download of
individual packages each time, but I think it still needs the connection to
decide what to install, and what is needed, right?

friday:
Yeah. It sounds like you found a great tool that you enjoy to use. But it doesn't seem to fit with our needs. Appreciate you taking the time to read our guidelines, contributing and rebasing over v6 and all.

the-allanc:
Let me reply to those comments in brief:

> Individual extensions may require manual
> installation now, which is not ideal obviously, but at least if they
> install using their package manager it will then be kept up-to-date
> automatically.

Maybe there is a difference between how you and I would be imagining how users would be managing third-party dependencies required by extensions; rather than installing a package via `apt`, I imagined that the user would have to use `pip` to install their dependencies. In that case, `pip-run` is actually a better alternative when it comes to using up-to-date dependencies.

> Additionally for me personally having the churn of each
> environment created and torn down every invocation is pretty much a
> deal-breaker as a user

I wrote this pull request as I'm writing my own plugin which requires external libraries. The overhead of the initial initialisation is annoying, but once the extension is running, then performance isn't an issue. Maybe you still find that unacceptable (which is why I suggested in my previous comment to allow a plugin to be run as soon as Ulauncher is launched), but I do want to make it clear that the overhead only applies on the _first_ invocation, not every time you send a new keyword query event.

> It would also make extension development pretty painful.

Using this branch, I define a single `use_pip_run` option in the manifest, and add a single `__requires__ = ['thedependency']` line in `main.py`. That was all I needed to be able to start writing my extension.

> I presume this also needs an Internet connection at runtime to work?

Yes. Annoyingly, there doesn't seem to be a good way to utilise the local pip cache without a call to Pypi first. The best _offline_ solution is to [run a Devpi mirror](https://blog.jaraco.com/devpi-as-offline-pypi-cache/), which is the recommendation from the pip-run author himself.

Making changes to work around that give a stronger case to instead managing environments directly within Ulauncher - ideally, you want environments to be persisted (which pip-run doesn't do), and it's easy to read dependencies from the manifest, so the advantage of being able to mention the dependencies in the main script file isn't compelling here.

Anyway, as a proof-of-concept, this was pretty straight-forward to implement. If adding virtualenv support for extensions takes a particularly long time to achieve, then you might want to consider pip-run as a stop-gap solution.

I'll let you guys decide whether you want to close the MR or if you think it might have some merit at some point. Thanks for taking a look.

troycurtisjr:
@the-allanc You may be interested in my latest comment I just added to #273 where I outline an approach using virtualenv that I think could be workable for ulauncher's usage. My plan was to take a stab at actually implementing that idea, however if you have interest in giving it a shot you are more than welcome to. I just wanted to check in to determine if this is an area that you have a strong interest in contributing in as I don't want to take away the opportunity if you do. Just let me know!

friday:
I'm going to close this now. It's been clear we don't want to go this way since the PR was created. But thanks again for the effort!

PR #973, opened 2022-02-13
Don't rely on Gdk.Display.get_primary_monitor().

troycurtisjr:
This is a newer Gdk method, and will return None if the user has not configured a primary
monitor (or has only a single monitor).

### Checklist
- [N/A] Update the documentation according to your changes (when applicable)

### DIFF
--- a/ulauncher/utils/display.py
+++ b/ulauncher/utils/display.py
@@ -10,20 +10,31 @@ from gi.repository import Gdk, GdkX11, Gio
 
 logger = logging.getLogger(__name__)
 
+def get_screen_and_monitor(use_mouse_position):
+    """
+    Returns the default screen and an appropriate monitor.
+    """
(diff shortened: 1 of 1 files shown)


troycurtisjr:
I ran into this exception when I pulled down the latest v6. I think this eliminates the exception when a primary monitory isn't defined, and doesn't use a method that requires 3.22 or greater of Gtk. I wasn't sure if we wanted to require that version or not. If we do, then there are some other deprecated methods that we should switch over as well.

friday:
3.22 came out more than 5 years ago, so I removed fallbacks and workarounds for older versions. But this comment was misleading as it was apparently more than a fallback: https://github.com/Ulauncher/Ulauncher/commit/ad2145e6dac2563427035a490baa8d9689eaed9b.

PR #974, opened 2022-02-13
Simplify Display utilities and switch from deprecated APIs (GTK 3.22+)

friday:
Does this look good and work for you @troycurtisjr? I can probably test this myself later, but I at least wasn't initially able to get the same issue as you did.

### Checklist


troycurtisjr:
Looks good and seems to work fine for me. I also checked in on RHEL8, in relation to #971, and it seems 3.22 is available there. So this seems all around like a good change! :+1:

troycurtisjr (approved):

PR #975, opened 2022-02-13
Small v6 bugfixes

troycurtisjr:
I fixed a few small issues I ran into today:
- I was getting spurious results show up when I first started Ulauncher, specifically a few of the fallback searches, but I had no query. Though interestingly it didn't happen every time. It definitely happened every time when I just typed spaces, but of course that wasn't a meaningful query because we `lstrip()` anyway. I also made it so that you can't even type the leading spaces, so that the user experience matches what is actually used.
- Looks like the extra space below the query box when there are no results was just due to a minimum size for the scrolling container. I added to code to just hide the container when there are no results.
- I had an issue where launching apps wouldn't close ulauncher. I found what appeared to be an inverted conditional.

### Checklist
- [] Update the documentation according to your changes (when applicable)

### DIFF
--- a/tests/ui/test_ItemNavigation.py
+++ b/tests/ui/test_ItemNavigation.py
@@ -56,11 +56,11 @@ class TestItemNavigation:
     def test_enter_no_index(self, nav, items):
         nav.select(2)
         selected_result = items[2].result
-        assert nav.enter('test') is selected_result.on_enter.return_value.keep_app_open.return_value
+        assert nav.enter('test') is (not selected_result.on_enter.return_value.keep_app_open.return_value)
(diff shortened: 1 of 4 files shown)


friday (approved):
Thank you! :+1:

friday:
1. I think this happened because I applied strip() a bit too aggressively and reverted it. I did know about it and had this on a mental note to fix it.
2. Perfect :+1: 
3. Nice! I have no idea how this could have worked in v5, considering how the conditions have been like this forever, but I did break this somehow in v6, without changing the conditions (perhaps two issues cancelled out each other before).

Issue #476, opened 2019-12-18
Support "desktop entries" application actions


endorama:
Add support for Desktop Entries Additional applications actions.

[Additional application actions](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions) are part of the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html).

Are extremely useful to trigger custom behaviours present in `.desktop` files.

Example `.desktop` file: `firefox.desktop`

<details>
<summary>Details</summary>

<pre>
[Desktop Entry]
Version=1.0
Name=Firefox Web Browser
Comment=Browse the World Wide Web
GenericName=Web Browser
Keywords=Internet;WWW;Browser;Web;Explorer
Exec=firefox %u
Terminal=false
X-MultipleArgs=false
Type=Application
Icon=firefox
Categories=GNOME;GTK;Network;WebBrowser;
MimeType=[...]
StartupNotify=true
Actions=new-window;new-private-window;

[Desktop Action new-window]
Name=Open a New Window
Exec=firefox -new-window

[Desktop Action new-private-window]
Name=Open a New Private Window
Exec=firefox -private-window
</pre>

</details>

I believe this feature would be very useful to support out of the box a wide array of advanced functionalities. 

Inspiration has been taken from [this tweet](https://twitter.com/elementary/status/1207394390550994946) about Elementary OS Application menu

<blockquote class="twitter-tweet"><p lang="en" dir="ltr">elementary OS <a href="https://twitter.com/hashtag/protip?src=hash&amp;ref_src=twsrc%5Etfw">#protip</a>: search for app actions in the Applications Menu! For example, quickly open straight to an incognito browser session by searching for “incognito”. Create a new Calendar event with “event”. Check for updates in AppCenter with “updates”. And more! <a href="https://t.co/2FrorTSOpD">pic.twitter.com/2FrorTSOpD</a></p>&mdash; elementary (@elementary) <a href="https://twitter.com/elementary/status/1207394390550994946?ref_src=twsrc%5Etfw">December 18, 2019</a></blockquote> 

What do you think?

gornostal:
I'll keep this in the backlog for now.

friday:
[...]

PR #979, opened 2022-02-18
Try to activate app before launching

### LINKED ISSUE #162: Switch to application if it's already running
When i've opened terminal in ubuntu and i couldn't switch to that terminal again using albert, it just reopen a new terminal.

Any plans to add this feature?

friday:
This fixes #162 for "pure" X11 (not Wayland or XWayland).

It's added as a non-default setting (because we can't support it in all environments).

It uses the command name or the "WM Class" in the .desktop entry (most apps don't specify the wm class) as the way to identify apps, and it only works on 100% matches (case-insensitive though). I think it's very important to avoid false positives, and better to start this way and deal with any false negatives we discover later (one seemingly safe way could be to strip non-alphanumerics before the comparison).

It's working well in my tests, and since swapping to using Wnck for this we can also add exceptions (like for [Thunar](https://gitlab.xfce.org/xfce/thunar/-/issues/731) that has two different apps identifying the same way and breaks in [jumpapp](https://github.com/mkropat/jumpapp) and [raise-or-run](https://github.com/bulletmark/raise-or-run) and even the default Gnome app launcher (but not in this PR).

### Checklist


friday:
There is a new standard [`SingleMainWindow`](https://nicolasfella.de/posts/taskbar-single-app/) which replaces the previous one `X-GNOME-SingleWindow`. If these pick up usage, then in the future it would make sense to always try to raise and only launch if that fails, then if not (and assuming we have implemented support for desktop actions), add the regular launch action as "open new window" if there isn't already an action for `new-window`.

troycurtisjr:
Code is looking good and I'll test it tonight on my system. I'm really glad you already added it as an option, otherwise that would have been my next PR 😁. Even though I don't personally want this setting enabled, I think making it the default makes total sense. It does seem to be the dominant mode among other launchers and as a new seeing those that would want it, might not realize it was available if it defaulted to off. For those of us that don't want it, when the behavior changes, that will prompt digging into the settings to see what changed and disabling it.

friday:
> Code is looking good and I'll test it tonight on my system. I'm really glad you already added it as an option, otherwise that would have been my next PR grin. Even though I don't personally want this setting enabled, I think making it the default makes total sense. It does seem to be the dominant mode among other launchers and as a new seeing those that would want it, might not realize it was available if it defaulted to off. For those of us that don't want it, when the behavior changes, that will prompt digging into the settings to see what changed and disabling it.

I always intended to support the behavior to launch a new window via #476 and maybe adding a manual entry when there's no "new-window" action. And maybe adding a new hotkey (like shift+enter). I think this is a lot better than a setting for it, but we would have to implement #476 first for that, and that's harder. And seeing as there's no reliable way to solve this issue currently, then I prefer a setting so people can opt out if it causes issues for them (as with the Thunar example).

friday:
Oh, that sucks. I did test it in X11 mostly, but also Wayland (actually still using Wayland now because of it) and it worked for the apps I tested. Maybe because I have Gnome set up to remember my session 🤦, and it started the apps with xwayland  because I was using X in the previous session.

It should "degrade gracefully" when it doesn't work, but making it default when it's not working with pure Wayland feels ugly.

Also, this means that `ulauncher-toggle` will only work if ulauncher is launcher with X11 or XWayland, although it's not the only thing that depends on this (same for centering and moving the window with the cursor).

I did find this, regarding your previous suggestion to use the Gnome APIs. It seems they removed that option unless you use a shell extension: https://www.reddit.com/r/gnome/comments/pneza1/gdbus_call_for_moving_windows_not_working_in/

Alternative I think [this](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1496) is what wmctrl would need for this to work in Wayland, and it doesn't look like this will be supported any time soon.

friday:
* Rebased with v6
* Made it non-default
* Added a warning in the Preferences if using Wayland (like the hotkey warning). I think it's better to move both settings to a "X11" section instead though.
* Added checks if using Wayland and if wmctrl is missing (since it's not a dependency now since e5df3b6 (we can revert add it back of course, but I think we shouldn't do that for just this feature)

I don't know if we should merge this, but it's the best I could do with this PR for now. I would like to find a better way, but I think there aren't any :(

friday:
Found a GTK API to use instead of `wmctrl`, so we don't need to introduce that dependency again, and we can control the implementation more to add exceptions etc.

Haven't tested in Wayland yet, but I know it won't work since it's x11 only.

Issue #986, opened 2022-03-02
Use GtkApplication


friday:
### Description

Ulauncher is currently not using any [GtkApplication](https://wiki.gnome.org/HowDoI/GtkApplication). Just a window. That window is wrapped in a singleton and all the stuff Ulauncher needs to remember between toggling the window is stored on the window object.

What we probably should do instead is add a GtkApplication to be the new toplevel object.

Every time we show the window we can create a new window, and instead of hiding it we close it for real. This way we get around the focus stealing issue with toggling the same window in Wayland (I have tested this and it did work to steal the focus in Wayland without using the dbus hack that introduces race conditions).

In addition, by using an application with application ID, we should automatically get this behavior:
> GtkApplication defaults to applications being single-instance. If the user attempts to start a second instance of a single-instance application then GtkApplication will signal the first instance and you will receive additional activate or open signals. In this case, the second instance will exit immediately, without calling startup or shutdown.

This should remove the need for the separate ulauncher-toggle command as well as the python-dbus dependency.

The problem of course is that this is a huge change for the architecture, so it would take a lot of time, and I'm not up for the job myself.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

friday:
Here is an example solving the same problem as #987, but using `Gtk.Application` instead of `python-dbus` and `wmctrl` (so we could remove those two dependencies apart from `wmctrl` would still be needed for #979 if implemented)

It also solves the current race condition with setting the focus in `ulauncher-toggle`, but the problem is it is very different from Ulauncher's architecture, and it also seems to reintroduce #508, but maybe there's a way around this.

```py
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
import sys
import gi
gi.require_versions({
    "Gdk": "3.0",
    "Gtk": "3.0"
})
# pylint: disable=wrong-import-position
from gi.repository import Gdk, Gtk


class Application(Gtk.Application):
    held = False

    # pylint: disable=arguments-differ
    def do_activate(self):
        # If you run the command when it's already started the second process will send the activate signal to the first
        # process and then exit. So nothing will run in the second process and this code will run in the first process
        # for every time you run the command (which is what we want)
        # Only issue I found with this is that the generic behavior for picking which screen to launch new windows picks
[...]

PR #987, opened 2022-03-06
Make ulauncher "toggle" by default if it's already running

friday:
The title is a bit misleading, because ulauncher already does this, but it doesn't do it well.

This PR does four things:
1. Moves the part of the main scripts that checks for already running instances to the top, and avoid loading all the heavy dependencies until after this check. This makes the command `ulauncher` about as fast as `ulauncher-toggle`.
2. Check if it was invoked with `--no-window` (if it was, don't show the window of the running Ulauncher instance).
3. Removes the previous warning "DBus name already taken. Ulauncher is probably backgrounded. Did you mean `ulauncher-toggle`".
4. Remove `--no-window` from the desktop entry. I think it's better if the desktop entry follows the experience of a regular "app", that you can "launch", while the systemd service file has `--no-window`, because it should run in the background.

What's missing for this to be a perfect solution is the Wayland focus stealing fix from ulauncher-toggle, but that fix is unreliable anyway and I would want to replace it with something better (#986). I think we should keep `ulauncher-toggle` around for a while anyway because people have that in their DE configs.

### Checklist

### DIFF
--- a/ulauncher.desktop
+++ b/ulauncher.desktop
@@ -4,7 +4,7 @@ Comment=Application launcher for Linux
 GenericName=Launcher
 Categories=GNOME;GTK;Utility;
 TryExec=ulauncher
-Exec=env GDK_BACKEND=x11 ulauncher --no-window
+Exec=env GDK_BACKEND=x11 ulauncher
 Icon=ulauncher
 SingleMainWindow=true
 Terminal=false
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -2,51 +2,24 @@ import sys
 import signal
 import logging
 from functools import partial
-import dbus
-import dbus.service
-from dbus.mainloop.glib import DBusGMainLoop
-from ulauncher.config import API_VERSION, VERSION, get_options
-
-# Start DBus loop
-options = get_options()
-logger = logging.getLogger('ulauncher')
-DBUS_SERVICE = 'net.launchpad.ulauncher'
-DBUS_PATH = '/net/launchpad/ulauncher'
-DBusGMainLoop(set_as_default=True)
-bus = dbus.SessionBus()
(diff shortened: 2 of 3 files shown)


troycurtisjr (commented):
Certainly a whole lot faster! 

Perhaps the dbus stuff could go into a dedicated module, and the actual logic could be in a function which could get called at the bottom of the module automatically at import time. That way at the top of this file it can be just an import, still needing to be called before the heavy weight modules of course. Having a separate module which does a bunch of actual work at import time, including potentially exiting, can be a little odd for sure, but in this case it seems like a reasonable trade off for a cleaner header and nicer division of responsibility. Since we don't need the bus or instance objects later, if they get put into a function then they'd go away automatically as well.

friday:
> Certainly a whole lot faster!

:+1: 

> Perhaps the dbus stuff could go into a dedicated module, and the actual logic could be in a function which could get called at the bottom of the module automatically at import time. That way at the top of this file it can be just an import, still needing to be called before the heavy weight modules of course. Having a separate module which does a bunch of actual work at import time, including potentially exiting, can be a little odd for sure, but in this case it seems like a reasonable trade off for a cleaner header and nicer division of responsibility. Since we don't need the bus or instance objects later, if they get put into a function then they'd go away automatically as well.

Good idea. I didn't think about this. It sort of hides some important implementation details, but I think it's better also. Especially since we already do a similar thing with ulauncher.utils.xinit.

I didn't implement it exactly like you suggested, but moved the UlauncherDbusService there too (kind of ugly considering how deeply it integrates with the window, but it shares the same dependencies as the other dbus code, so I think this looks cleaner.

friday:
Closing in favor of #988 (unless that one is rejected for some reason, then we can reevaluate this solution)

PR #988, opened 2022-03-09
Switch to Gtk.Application and make `ulauncher` command "toggle" by default

friday:
Alternative implementation of #987, using Gtk.Application for the main loop. This handles stuff we had to do manually before with dbus, so we can remove the `python-dbus` dependency.

I also removed `gi.require_version('GLib', '2.0')` because I recently learned it's [not needed](https://gitlab.gnome.org/GNOME/pygobject/-/blob/master/gi/importer.py#L64)

As a nice side effect, this seems to get rid of the race condition issue with ulaucher-toggle. Previously I ran into that a lot on Wayland, but now I removed the `sleep 0.03` bit completely and it works every time (still need wmctrl though).

I kept things in the same file mostly because I wanted to keep the changes minimal. This PR is just the first part and there' s so much more this opens up for:
* Move the `UlauncherApp` to it's own singleton module
* Moving logic from the window to the app so we can create a new window for each time (this way we won't need wmctrl in `ulauncher-toggle`)
* Use the application to send desktop notifications (removing the libnotify dependency).
* Move the argparse logic to the UlauncherApp.setup method instead of handling it in two places.

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ For the v6 branch you need the the following to setup the local build environmen
 
   ```sh
   sudo apt-get install \
-    gobject-introspection libgtk-3-0 libkeybinder-3.0-0 wmctrl \
+    gobject-introspection libgtk-3-0 libkeybinder-3.0-0 \
     gir1.2-{glib-2.0,gtk-3.0,gdkpixbuf-2.0,notify-0.7,webkit2-4.0,keybinder-3.0,ayatanaappindicator3-0.1} \
     python3-{all,gi,levenshtein}
   ```
@@ -60,8 +60,7 @@ For the v6 branch you need the the following to setup the local build environmen
 
   ```sh
   sudo pacman -Syu --needed \
-    gtk3 webkit2gtk libappindicator-gtk3 libnotify libkeybinder3 wmctrl \
-    python-{gobject,levenshtein}
+    gtk3 webkit2gtk libappindicator-gtk3 libnotify libkeybinder3 python-{gobject,levenshtein}
   ```
 </details>
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ For the v6 branch you need the the following to setup the local build environmen
   sudo apt-get install \
     gobject-introspection libgtk-3-0 libkeybinder-3.0-0 wmctrl \
     gir1.2-{glib-2.0,gtk-3.0,gdkpixbuf-2.0,notify-0.7,webkit2-4.0,keybinder-3.0,ayatanaappindicator3-0.1} \
-    python3-{all,gi,dbus,levenshtein}
+    python3-{all,gi,levenshtein}
   ```
 
 </details>
@@ -61,7 +61,7 @@ For the v6 branch you need the the following to setup the local build environmen
   ```sh
   sudo pacman -Syu --needed \
     gtk3 webkit2gtk libappindicator-gtk3 libnotify libkeybinder3 wmctrl \
-    python-{gobject,dbus,levenshtein}
+    python-{gobject,levenshtein}
   ```
 </details>
 
--- a/bin/ulauncher-toggle
+++ b/bin/ulauncher-toggle
@@ -1,13 +1,4 @@
 #!/bin/sh
-
-dbus-send \
-    --session \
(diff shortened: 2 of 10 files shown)


ulauncher/main.py
```diff
@@ -1,54 +1,68 @@
...
+        if Settings.get_instance().get_property('show-indicator-icon'):
+            AppIndicator.get_instance().show()
 
-    @dbus.service.method(DBUS_SERVICE)
-    def toggle_window(self):
-        self.window.toggle_window()
+    def do_activate(self, *args, **kwargs):
+        self.window.show_window()
 
+    def do_command_line(self, *args, **kwargs):
```
troycurtisjr, line 47: Should this message signature just be `def do_command_line(self, appcl):` or similar? Looks like according to doc that is all that will be passed in.

CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ For the v6 branch you need the the following to setup the local build environmen
   sudo apt-get install \
     gobject-introspection libgtk-3-0 libkeybinder-3.0-0 wmctrl \
     gir1.2-{glib-2.0,gtk-3.0,gdkpixbuf-2.0,notify-0.7,webkit2-4.0,keybinder-3.0,ayatanaappindicator3-0.1} \
-    python3-{all,gi,dbus,levenshtein}
```
troycurtisjr, line 39: Excellent removal of a dependency!

bin/ulauncher-toggle
```diff
@@ -1,13 +1,2 @@
...
-    --print-reply \
-    --dest=net.launchpad.ulauncher \
-    /net/launchpad/ulauncher \
-    net.launchpad.ulauncher.toggle_window
-
-# dbus-send is asynchronous and sometimes it takes a bit of time for Ulauncher window to be presented
-sleep 0.03
-
-wmctrl -a "Ulauncher - Application Launcher"
+ulauncher
```
troycurtisjr, line 2: You might `exec` here on the off chance ulauncher isn't actually running and this starts it.

troycurtisjr (commented):
I'm really liking it, and it didn't really require as much changes as you might have thought! Seems to work great for me (once I updated my workaround shortcut to point to the right place)

ulauncher/main.py
friday: That's what I did originally, but then I get these linting errors locally (my pylint is way more recent than the docker image one though):
>ulauncher/main.py:47:4: W0221: Number of parameters was 3 in 'Application.do_command_line' and is now 2 in overridden 'UlauncherApp.do_command_line' method (arguments-differ)
>ulauncher/main.py:47:4: W0221: Variadics removed in overridden 'UlauncherApp.do_command_line' method (arguments-differ)

Same for `do_activate()`.

bin/ulauncher-toggle
friday: Ok

friday:
Merging. We can always tweak the details later.

DADA30000:
> Pushed some more fixes. I was surprised after doing some tests that the wmctrl workaround just wasn't needed any more at all. When I implemented it it wasn't working to set the focus without it, but now it is (and yes, I'm still on wayland).
> 
> So we can remove it after all, and no need to create a new window each time fore that 🎊
> 
> Also removed the toggle behavior in favor of always showing. It never really made sense to me have the the hotkey hide the window if it's showing. At least I think it's better to have the behavior consistent, and the new behavior is much better.
> 
> Might want to add wmctrl back as an optional dependency for #979 though (if we decide we want to pick up from that PR). I also considered other reasons to use wmctrl more. But in this PR it's not used, so it could be removed.

can I somehow redo the toggle behavior like it was before? I just have issues with v5, and v6 is working fine, but not having ability to toggle normally is just annoying and I have to workaround that, and also there why launcher is not disappearing after I open application? It;'s just strange and weird for me, other launchers close by default...

friday:
> but not having ability to toggle normally is just annoying and I have to workaround that

It's definitely not "normal", and it's not coming back. Just press Esc to close. It's standard for windows like this.

> and also there why launcher is not disappearing after I open application? It;'s just strange and weird for me, other launchers close by default...

This was fixed and released two days ago: https://github.com/Ulauncher/Ulauncher/releases/tag/v6.0.0-beta10

PR #1003, opened 2022-03-29
Update `ulauncher-toggle` to use gdbus

Jaakkonen:
This sends a direct `org.gtk.Application.Activate` message to the window reducing latency of toggling from 810ms to 9ms on a test machine.

### DIFF
--- a/bin/ulauncher-toggle
+++ b/bin/ulauncher-toggle
@@ -1,2 +1,6 @@
 #!/bin/sh
-exec ulauncher
+gdbus call --session \
+  --dest net.launchpad.ulauncher \
+  --object-path /net/launchpad/ulauncher \
+  --method org.gtk.Application.Activate \
+  {}


friday:
Your numbers are much higher than mine, which is good I guess (need to test of different hardware).

Are you getting these numbers when running `ulauncher` from a built system install (not the dev environment) and there's already a running process?

If I run `ulauncher` when it's already running it shows instantly, but `./bin/ulauncher` from the dev environment is much slower because it also checks and conditionally builds the preferences. I think you must have tested this?

`time ulauncher-toggle` with your modifications runs very very fast for me (`0,01s user 0,01s system 55% cpu 0,029 total`), but it already ran very fast before:
`time ulauncher`: `0,10s user 0,01s system 95% cpu 0,122 total`
`time ./bin/ulauncher` from the dev directory though outputs `0,46s user 0,08s system 95% cpu 0,561 total`, which is why you shouldn't use that for performance tests as the end-user won't be running that.

Also, these measure time until the command finishes. Not until the window shows. I don't know if the `gdbus call` command waits for the method to complete for example (I think it doesn't), so measuring the time is complicated.

Still an improvement though, so I think I will merge it. It seems to work with Wayland and focus too.

friday (approved):

friday:
Oh, you switched from `dbus-send` instead of `gdbus call` which we used before v6. I don't know the implications of this, but maybe `gdbus call` actually waits for the method to complete.

Jaakkonen:
`gdbus` was chosen as that supports sending empty dicts whereas `dbus-send` doesn't support that. 

The numbers are in development environment testing with `./bin/ulauncher-toggle` which does a `sh` jump, some checks in the `./bin/ulauncher` after eventually going to the `Gtk.Application.__init__` which tries to acquire the dbus name and activates it & exists if it's already present.

In fact my benchmark numbers are quite a bit off due to the checks in `./bin/ulauncher` testing debug script and those account for most of the time.

I repeated the tests by running Ulauncher directly with `python -c 'from ulauncher.main import main; main()'` and adding a `os._exit(1)` call just before going to `Gtk.Application.__init__`. That yielded a lower bound for the time and I got 128.3 ms ±   8.9 ms execution time for that case. Without the premature exit (letting Gtk recognize the existing window and sending Activate message to it) yielded 172.3 ms ±  11.3 ms execution time.
So this suggests that the improvement to writing anything to DBus to make the window be toggled is anywhere from 120ms to somewhere around 160ms.

Thanks for noting the issue in my benchmarking style, I didn't take a closer look into what I was saying. It's always good to say what you exactly mean when throwing around numbers so that those stay meaningful and I didn't do that in this case.

friday:
Thanks for explaining :+1: It's faster and works well for me, so I see no reason not to merge :)

PR #1004, opened 2022-03-30
Make the icon error handling a little more complete.

troycurtisjr:
I noticed some strange behavior where some characters I would type would have results, then the next character wouldn't and then next would. I found it was due to an uncaught exception loading an icon:
```
2022-03-29 22:20:27,557 | ERROR | ulauncher: except_hook() | Uncaught exception
Traceback (most recent call last):
  File "/home/troycurtisjr/working/oss/Ulauncher/v6/ulauncher/ui/windows/UlauncherWindow.py", line 369, in show_results
    results = self.create_item_widgets(results, self._get_user_query())
  File "/home/troycurtisjr/working/oss/Ulauncher/v6/ulauncher/ui/windows/UlauncherWindow.py", line 406, in create_item_widgets
    item_frame.initialize(builder, result, index, query)
  File "/home/troycurtisjr/working/oss/Ulauncher/v6/ulauncher/ui/ResultWidget.py", line 55, in initialize
    self.set_icon(load_icon(result.icon, result.ICON_SIZE * self.scaling_factor))
  File "/home/troycurtisjr/working/oss/Ulauncher/v6/ulauncher/utils/icon.py", line 47, in load_icon
    return GdkPixbuf.Pixbuf.new_from_file_at_size(icon_path, size, size)
gi.repository.GLib.GError: g-file-error-quark: Failed to open file “/usr/share/ulauncher/media/stackoverflow-icon.svg”: No such file or directory (4)
```
There was already some handling for not finding the file path, but there could also be errors loading the file itself (like this case), so I added in some additional handling.

### Checklist

### DIFF
--- a/ulauncher/utils/icon.py
+++ b/ulauncher/utils/icon.py
@@ -15,6 +15,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_EXE_ICON = "icons/executable.png"
 
+
 def get_icon_path(icon, size=32, base_path=""):
     """
     :param str icon:
--- a/ulauncher/utils/icon.py
+++ b/ulauncher/utils/icon.py
@@ -13,6 +13,7 @@ from ulauncher.config import get_asset
 icon_theme = Gtk.IconTheme.get_default()
 logger = logging.getLogger(__name__)
 
+DEFAULT_EXE_ICON = "icons/executable.png"
 
 def get_icon_path(icon, size=32, base_path=""):
     """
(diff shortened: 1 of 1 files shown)


friday (approved):
Looks good.

I will probably try to restructure it a bit more because my programming OCD really doesn't like the duplicated `GdkPixbuf.Pixbuf.new_from_file_at_size(icon_path, size, size)`, but it's hard to avoid it without having to actually add more code to compensate.

friday:
Made some small changes to this in 1b66c76, but it's functionally the same. Most notably made the exception block in load_icon call itself with the default icon unless it was already called with it. And if it was called with it, it throws a more specific error instead to avoid recursion (and maybe help if someone has a really broken environment missing the fallback).

PR #1012, opened 2022-04-16
Include keywords when searching for application.

loops:
Some .desktop files contain an entry such as:

    Keywords=key1;key2;key3

Which associates each of the semicolon separated keys with the
application.  These can be helpful when searching for the application.

### Checklist


friday (approved):
I've been reluctant to add this because of the reasons https://github.com/Ulauncher/Ulauncher/issues/270#issuecomment-1080166394, but this implementation is limited to only match one keyword at a time, which I think is a reasonable middle ground. And it's barely adding any complexity to the code, which is great.

So I will merge it. If someone discovers it affects performance or makes the results too noisy we might pull it or weight it down further compared to the other things the search matches on.

PR #1015, opened 2022-04-17
Calculator: Add trigonometric methods and constants

friday:
All the actual functionality is exactly as implemented by @Henje in #1009, but this is rebased over #1014 which solved some of the issues with #1009. `is_enabled()` has also been completely rewritten to use ast (just checking the root node) instead of regex, which made it easier to prevent false positives.

Also added more tests.

It seems pretty solid to me after these changes.

### Checklist

### DIFF
--- a/tests/modes/calc/test_CalcMode.py
+++ b/tests/modes/calc/test_CalcMode.py
@@ -11,21 +11,24 @@ class TestCalcMode:
 
     def test_is_enabled(self, mode):
         assert mode.is_enabled('5')
+        assert mode.is_enabled('+2')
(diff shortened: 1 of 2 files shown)


Henje:
This looks great. I was not sure how to integrate my extension well, so this seems like the better solution. Thanks for your help.

friday:
> This looks great. I was not sure how to integrate my extension well, so this seems like the better solution. Thanks for your help.

You did awesome work on your PR, but the calculator's was just not in a good state to extend with this type of functionality, which lead to the strange results.

PR #1019, opened 2022-04-25
Add build support for Ubuntu 22.04 "jammy"

basert:


friday:
Looks good. Thank you :+1:

This is for our Launchpad builds and won't be triggered until the v6 release though. Would need to add it to the v5 branch and somehow trigger a rebuild of the last release or make a new release to actually get the package up for Jammy though.

PR #1020, opened 2022-04-27
Run extensions on demand

friday:
See #1063

Update 2026: This is now a clean and fully working implementation. #1752 already moved a lot of things in the right direction.

This PR will wait to load the extension until the keyword is called, and then unloads it when the window is hidden. 

This avoid hogging ram and cpu needlessly. It also makes `ExtensionController` a lot less complex, because all methods are sync now.

The graceful timeout before killing is gone, but `UnloadEvent` is still present (it probably shouldn't be).

Removing reliance on the extension "running" state is also good, because that only works in the app runtime.

The log is now too noisy when closing an extension after closing the window:

```
2026-07-06 18:46:14,448 ℹ️  INFO modes.extensions.extension_runtime: Terminating extension "home.albin.src.personal.ulauncher-clipboard" stop:103
2026-07-06 18:46:14,449 ℹ️  INFO modes.extensions.extension_runtime: Extension home.albin.src.personal.ulauncher-clipboard connection closed socket_cleanup:65
2026-07-06 18:46:14,450 ℹ️  INFO modes.extensions.extension_runtime: Sending SIGKILL to extension home.albin.src.personal.ulauncher-clipboard _kill:111
2026-07-06 18:46:14,461 ℹ️  INFO modes.extensions.extension_runtime: Extension "home.albin.src.personal.ulauncher-clipboard" was stopped by the user handle_exit:149
```

### DIFF
--- a/tests/modes/extensions/test_extension_mode.py
+++ b/tests/modes/extensions/test_extension_mode.py
@@ -30,6 +30,7 @@ def _mock_service(mocker: MockerFixture) -> MagicMock:
     mocker.patch.object(ext_service, "get", mocked.get)
     mocker.patch.object(ext_service, "iterate", mocked.iterate)
     mocker.patch.object(ext_service, "is_running", mocked.is_running)
+    mocker.patch.object(ext_service, "start_extension", mocked.start_extension)
     return mocked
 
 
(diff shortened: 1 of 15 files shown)


friday:
> ## ℹ️ Startup-perf improvement no longer present
> The earlier improvement is no longer measurable as of [ac5a629](https://github.com/Ulauncher/Ulauncher/commit/ac5a629fbebd78ff5a2b7f8dd36faa5ecb5b84fe). `make perf-compare` against `main` now produces |delta| < 5%.

Likely because main already got extension related perf improvements in #1772.

PR #1039, opened 2022-05-24
Use monospaced fonts in shortcut query field

wafarm:
### Checklist


friday:
Could you please attach a screenshot? It's a very straight forward UI change if we can just see it first there's no need to  check out the PR and do it :)

friday:
Oh, it's for the code input textarea and the query on the list page. I'll approve it. I think the textarea at least is a nice improvement. The list page I'm less sure about, but I also don't think it's worse.

PR #1041, opened 2022-05-26
Add ulauncher-toggle query param to set up initial query.

ajboni:
This PR makes possible to pre set the query when calling `ulauncher-toggle` 
usage: 

`ulauncher-toggle` => Same  default behavior
`ulauncher-toggle -q some_query`
`ulauncher-toggle -q "some long query"`
`ulauncher-toggle --query=some_query`
`ulauncher-toggle --query="some long query"`

### DIFF
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -39,10 +39,10 @@ class UlauncherApp(Gtk.Application):
         self.window.set_application(self)
         self.window.finish_initializing()
 
-    def do_before_emit(self, data ):
+    def do_before_emit(self, data):
         query = data.lookup_value("query",  GLib.VariantType("s"))
         if query:      
-           self.window.initial_query = query.unpack() + " "
+           self.window.initial_query = query.unpack()
 
     def do_activate(self, *args, **kwargs):
         self.window.show_window()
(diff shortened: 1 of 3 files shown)


ulauncher/main.py
```diff
@@ -39,6 +39,11 @@ def setup(self, _):
         self.window.set_application(self)
         self.window.finish_initializing()
 
+    def do_before_emit(self, data ):
+        query = data.lookup_value("query",  GLib.VariantType("s"))
+        if query:      
+           self.window.initial_query = query.unpack() + " "
```
friday, line 45: I think we shouldn't add the space. If people want the space they can send ulauncher-toggle -q "keyword ". If people don't want a space they can't send negative one character.

friday (changes requested):
Looks good and simple.
1. I don't think we should pad with the space.
2. I don't know if I think we *need* the getopts logic rather than just `$1`, but it's more future-proof if we do eventually want more arguments, so I'm good with that.

ulauncher/main.py
ajboni: Yes, makes sense!

friday (approved):

friday:
Thanks again :+1:

Issue #1068, opened 2022-06-19
Open Ulauncher from tray icon?


friday:
### Discussed in https://github.com/Ulauncher/Ulauncher/discussions/1067

<div type='discussions-op-text'>

<sup>Originally posted by **seidler2547** June 19, 2022</sup>
Hi, I would like to use Ulauncher on a tablet device and so I was lookinng for a way to open it with a single tap. I found that tapping/clicking the tray icon only opens the menu, can it be configured to show the menu on right click and open the input field on left click?</div>

friday:
From #1069:

It is technically possible to distinguish right and left clicks. I for example use Enpassed, that does this. But it seems using the app indicator API we do now, it's not possible, and the PR above uses middle button click instead.

https://stackoverflow.com/a/66270962/633921

XApp seems to be supported by all distros we package for (Ubuntu Bionic is the oldest one), so we could possibly switch to that: The dependency exists for the oldest Ubuntu version we support: [gir1.2-xapp-1.0](https://packages.ubuntu.com/bionic/gir1.2-xapp-1.0).

PR #1069, opened 2022-06-19
add "Show Ulauncher" menu item to tray menu and set it as secondary target

seidler2547:
Implements https://github.com/Ulauncher/Ulauncher/issues/1068

It's not possible to distinguish left or right click with libindicator, so the next best option is to add a menu item and also set it as secondary target, which means that it is called when a middle click is received

### Checklist

### DIFF
--- a/ulauncher/ui/AppIndicator.py
+++ b/ulauncher/ui/AppIndicator.py
@@ -30,6 +30,8 @@ class AppIndicator:
 
     def __init__(self, app):
         if AppIndicator3:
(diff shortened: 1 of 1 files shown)


friday:
Thank you :+1:  It is technically possible to distinguish right and left clicks. I for example use Enpassed, that does this. But it seems using the app indicator API, it's indeed not possible.

https://stackoverflow.com/a/66270962/633921

XApp seems to be supported by all distros we package for, so we could possibly switch to that: https://packages.ubuntu.com/bionic/libdevel/libxapp-dev

This PR is an improvement either way.

PR #1075, opened 2022-06-22
AppIndicator: Support separating between right and left clicks

### LINKED ISSUE #1068: Open Ulauncher from tray icon?

friday:
This changes the documented optional dependency to Xapp, which is a lesser known, but better library for us.

It tries to import and use Xapp and show Ulauncher on left click on the icon and show the menu on right click. If that fails it uses  the old way with (Ayatana)AppIndicator

Fixes #1068

### Checklist


friday:
@seidler2547 Maybe you can try this? While it wasn't a prio to add the functionality, the dependency change (even if it's an optional dependency) made it more important to including this in the coming major release rather than having to introduce it later.

See the new dependency:
https://github.com/Ulauncher/Ulauncher/pull/1075/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055

PR #1079, opened 2022-06-27
Remove FileQueries and use file system atime

friday:
FileQueries was a class used to store and update the last time each file was accessed via ulauncher's file browser mode.

There was never any reason to store this imo. We can get the result sorted by `atime` (access timestamp), which is a lot more useful as that's the last time the file was accessed by any means. Not just by Ulauncher.

It was implemented with `os.scandir()` which is faster than `os.listdir()` and `Path.iterdir()`

Also made the name sorting case-insensitive.

### Checklist


troycurtisjr:
Of course, `noatime` is a very common mount option used as an optimization, so I suspect the value isn't meaningful in a non-trivial number of setups. However, it seems it is really only a sorting issue so probably isn't a huge deal breaker.

friday:
> Of course, `noatime` is a very common mount option used as an optimization, so I suspect the value isn't meaningful in a non-trivial number of setups. However, it seems it is really only a sorting issue so probably isn't a huge deal breaker.

:+1: 

I never really cared about the sorting. Could be alphabetical, atime, mtime or ctime as long as it makes sense, is consistent and more useful than not sorting. I went with atime since it was the equivalent of what the previous solution used. What I care about is the maintenance costs and performance overhead of inventing our own ways to deal with things instead of relying on standards. For this example I think there is also a privacy concern with storing our our own atime instead of respecting the mount option (although there are other similar concerns with storing the other usage data).

I use noatime boot option myself, but forgot. I'm still getting the files sorted correctly by either mtime or ctime. Maybe atime falls back on mtime when using noatime? Either way this PR works better for me both as a user and maintainer.

Issue #1074, opened 2022-06-21
New "Triggers" API to replace keywords


friday:
The current way to define a trigger is in the extension manifest as a [`preference`](http://docs.ulauncher.io/en/latest/extensions/tutorial.html#manifest-json) with `type` "keyword".

There is no way to define a trigger without a keyword, and in its place there's only a [hacky workaround](https://github.com/friday/ulauncher-gnome-settings) (discovered by me before I was the maintainer). This workaround [no longer works in v6](https://github.com/Ulauncher/Ulauncher/issues/1072#issuecomment-1161896020). And rather than trying to "fix the workaround" it would be better to officially support this type of behavior. Essentially there are three scenarios:

1. ✅ Keyword and argument (supported now for extensions and shortcuts)
2. ❌ Application-like triggers without arguments
3. ❌ Arguments without shortcuts (like the calculator)

This proposal is an API change to support 2 and eventually 3, by moving the shortcuts from preferences to a new place "triggers" (preferences will remain as an optional list for actual preferences).

```diff
{
  "required_api_version": "3",
  "name": "My extension",
  "description": "Description",
  "developer_name": "John Doe",
  "icon": "images/icon.png",
- "preferences": [
+ "triggers": [
    {
      "id": "demo_kw",
-     "type": "keyword",
      "name": "Demo",
      "description": "Demo extension",
-     "default_value": "dm"
+     "keyword": "dm",
+     "icon": "images/optional-alternative-icon.png"
    }
  ]
}
```

#1056 already laid the ground work to keywords no longer being considered as "preferences", and #817 fits in better with this as well (confusing to have optional icon arguments for "preferences" that only applies to keywords).

Note:
* The `keyword` argument is optional. If unspecified, the trigger would work as an application-like trigger instead.
* Note to self: The old `get_searchable_items()` in the modes should be renamed to `get_triggers()`

To support case 3 we would have to add another argument (validation pattern/regex). and also make a user interface in the preferences to let users opt in for this for the extensions they want to work without the keyword (because it shouldn't be up to the extension devs). This will require more work.

Related to #306, and #1072

friday:
Fixed in #1097

Issue #338, opened 2019-01-23
feature suggestion: clear list of frequently used apps


lebedov:
It would be nice to include a button in the preferences to clear the list of frequently used apps.

lebedov:
Thanks - just what I was looking for!

humanplayer2:
Should others find this as I did, the file to remove is now `~/.local/share/ulauncher/app_stat_v2.db`

PR #1097, opened 2022-07-26
Implement triggers

friday:
Implements #1074

#### New terminology:
* "input triggers" (`on_input`) for the established trigger type (previously referred to as "keywords"). They are for getting user input after all.
* "launch triggers" (`on_launch`) for the new type of triggers that trigger instantly when you activate them. They do work similar to apps (which are "launched"). "launch" isn't 100% spot on, since the action might not actually be to launch something. It could be to refresh/reload a file for example (like this: https://github.com/Ulauncher/Ulauncher/issues/338). But it's the best I could think of which works well for the method name too.

Can be tested with [this extension](https://github.com/friday/ulauncher-gnome-settings/tree/apiv3) (v6 should support the gnome settings panels by default on most systems, so if this "works" you get duplicates):

It's as simple as this:

#### manifest.json:
```json
{
  "api_version": "3",
  "authors": "friday",
  "name": "Gnome Settings",
  "icon": "icon.svg",
  "triggers": {
    "background": {
      "name": "Background Settings",
      "description": "Change your background image to a wallpaper or photo"
    },
    "bluetooth": {
      "name": "Bluetooth Settings",
      "description": "Turn Bluetooth on and off and connect your devices" 
    },
    "color": {
      "name": "Color Settings",
      "description": "Calibrate the colour of your devices, such as displays, cameras or printers" 
    }
  }
}
```

#### main.py:
```py
import subprocess
import shutil
from ulauncher.api.shared.action.HideWindowAction import HideWindowAction
from ulauncher.api import Extension

class GnomeSettings(Extension):
    def on_launch(self, trigger_id):
        subprocess.Popen(['gnome-control-center', trigger_id])
        return HideWindowAction()

if __name__ == '__main__':
    GnomeSettings().run()
```

Hence, the `on_launch` method is only for the new trigger type. The old trigger type (now called "input triggers") need a "keyword" property in the manifest and then you will be able to use this method:

```py
def on_input(self, input_text: str, trigger_id: str):
    ...
```

`input_text` is a string with the user input after the keyword. So if the user enters "find pdf" you only get "pdf", but you get the trigger_id corresponding to "find" as the second argument. This is better since the actual keyword can be customized by the user and hence can't be used safely to determine which trigger. It's also easier for people who want typing information to not have to import our `Query` class.

The documentation still doesn't include `on_launch` yet, but it was always kind of basic and doesn't include some of the other ones either.

### Checklist


friday:
I'm not super happy with the method name on_static_trigger for triggers with no keywords or arguments. I kind of just went with something here. The other method on_query_change existed before in v6 (but not in v5). So I just changed the arguments for that. But maybe it's better if it's on_query_trigger. Or just on_query and on_static, or some variant of `on_(query|input|keyword|prompt)_(change|trigger)` and whatever name you can find for the other one that fits to pair with that `on_(static|nullary|keywordless|void|direct)_(change|trigger)`.

My first attempt used the same event and method for both, but I think that was worse, because the main argument for the query one is the query, which the other one doesn't have.

Also think what we call them matters a lot for the documentation and how people think of them. The new type is far less important than the old keyword one, so I'm fine with it taking a back seat though.

Other than that I think this PR is spot on myself. I think we should move [this extension](https://github.com/friday/ulauncher-hash2) to the Ulauncher repos and also use it for the documentation since it's extremely simple, yet demonstrating a valid use case of multiple keyword arguments.

I was able to make it even easier by adding support for `yield` (this is not in the PR, but will be in the v6 branch soon):

```py
import hashlib
from ulauncher.api import Extension, ExtensionResult
from ulauncher.api.shared.action.CopyToClipboardAction import CopyToClipboardAction

class Hash2(Extension):
    def on_query_change(self, argument, trigger_id):
        # Show the algorithm specified as keyword, or all if the keyword was "hash"
        algorithms = hashlib.algorithms_guaranteed if trigger_id == 'hash' else [trigger_id]

        for algorithm in algorithms:
            try:
                seed = hashlib.new(algorithm)
                seed.update(argument.encode())
                hash = seed.hexdigest()
                yield ExtensionResult(
                    name=hash,
                    description=algorithm,
                    icon='icon.svg',
                    on_enter=CopyToClipboardAction(hash),
                )
            except:
                pass

if __name__ == '__main__':
    Hash2().run()
```

gornostal (approved):

gornostal:
> I think we should move [this extension](https://github.com/friday/ulauncher-hash2) to the Ulauncher repos and also use it for the documentation since it's extremely simple

Good idea!

friday:
@brpaz Would be interested in hearing if you have thoughts on the API changes. You have written more code against it than anyone.

The PR redefines preferences with the type "keyword" as "input triggers" and the query argument to "input". And also adds a new type of trigger "launch trigger" which is a simpler type of trigger that doesn't take user input but "launches" (calls the extension listener) directly.

The idea is that this will enable future changes so the other type of "keywordless" (like our calculator) will be possible also down the road. But it's more complex and won't be part of v6.0.0.

brpaz:
@friday hey. just saw this now. I think it looks good.

I have been using more and more this pattern of having multiple keywords/triggers in an extension, inside of having a single entry point and then parsing the entire input text, to know which sub-action to execute.

I think this "trigger" approach simplifies a lot this process. The new `on_input` function signature, will reduce a lot of boilerplate code that I need to do frequently, to get the trigger, parse the arguments, etc.

friday:
> I have been using more and more this pattern of having multiple keywords/triggers in an extension, inside of having a single entry point and then parsing the entire input text, to know which sub-action to execute.
> 
> I think this "trigger" approach simplifies a lot this process. The new `on_input` function signature, will reduce a lot of boilerplate code that I need to do frequently, to get the trigger, parse the arguments, etc.

Great :) Separating preferences and triggers was definitely the right move. The differ quite a lot (one having different properties and uses). Triggers custom keywords are not a concern the extension, and removing it completely would be best (which is *almost* the case with this PR, but they will still be in `self.preferences` for backwards compatibility. Changing the terminology is tricky though, so I wanted confirmation that that part felt natural and made sense :)

Issue #1110, opened 2022-08-17
Ulauncher causes clion to crash


otisdog8:
### Environment

Arch Linux, KDE, Nvidia GPU (both wayland and X)

### Bug description

I tried opening clion on ulauncher and as soon as it loaded a project it crashed. What's notable is that it only crashes when using ulauncher. If I run clion from the terminal or using systemd-run (also from the terminal), or using alt+f2 on KDE it doesn't have any issues. 

This happens on both my laptop (intel+nvidia, wayland) and desktop (nvidia, wayland). I've tested and it occurs on v6 (built from source), ulauncher package in aur, and the ulauncher-git package in aur.

### Log output

```shell
Gtk-Message: 10:05:40.961: Failed to load module "appmenu-gtk-module"
2022-08-17 10:05:41,192 | INFO | ulauncher: main() | Ulauncher version 5.14.7
2022-08-17 10:05:41,192 | INFO | ulauncher: main() | Extension API version 2.0.0
2022-08-17 10:05:41,192 | INFO | ulauncher: main() | GTK+ 3.24.34
2022-08-17 10:05:41,192 | INFO | ulauncher: main() | Is Wayland: False
2022-08-17 10:05:41,192 | INFO | ulauncher: main() | Wayland compatibility: off
2022-08-17 10:05:41,210 | DEBUG | ulauncher.ui.windows.Builder: __init__() | consider using a pythonic name instead of design name 'input-box'
2022-08-17 10:05:41,211 | DEBUG | ulauncher.ui.windows.Builder: do_connect() | connect builder by name 'ulauncher_window','destroy', 'on_destroy'
2022-08-17 10:05:41,211 | DEBUG | ulauncher.ui.windows.Builder: do_connect() | connect builder by name 'ulauncher_window','focus_in_event', 'on_focus_in_event'
2022-08-17 10:05:41,211 | DEBUG | ulauncher.ui.windows.Builder: do_connect() | connect builder by name 'ulauncher_window','focus_out_event', 'on_focus_out_event'
2022-08-17 10:05:41,212 | DEBUG | ulauncher.ui.windows.Builder: do_connect() | connect builder by name 'input','key_press_event', 'on_input_key_press_event'
2022-08-17 10:05:41,212 | DEBUG | ulauncher.ui.windows.Builder: log_unconnected_functions() | Not connected to builder 'on_input_changed'
[...]

PR #1111, opened 2022-08-18
Fix clion crash (#1110)

otisdog8:
This PR fixes issue #1110 which is caused by the presence of os.setsid. I assume this could be useful on systems missing systemd-run so this PR keeps that functionality intact in that case while disabling it when systemd-run is present (where os.setsid shouldn't have an impact). I did test this and apps started by ulauncher don't stop when ulauncher is stopped. 

### Checklist


friday:
Looks good. Thank you 👍

PR #1120, opened 2022-09-08
Merge Result classes (and glade ui files)

### LINKED ISSUE #1118: Merge Result classes (Result, SearchableResult, SmallResult, ExtensionResult, ExtensionSmallResult)
### Description

These don't have to be separate classes. They can be implemented using:
1. Properties to decide if they're big or small
2. Properties to decide if they're searchable
3. Check if running from an extension, so the regular Result class can work there also.

In the case of the small/normal variants the UI files can also be merged.

### Communication guidelines

friday:
Fixes #1118

### Checklist


ulauncher/api/__init__.py
```diff
@@ -1,8 +1,12 @@
...
 from ulauncher.api.extension import Extension
-from ulauncher.api.result.result import Result
-from ulauncher.api.result.searchable_result import SearchableResult
-from ulauncher.api.result.small_result import SmallResult
-from ulauncher.api.result.extension_result import ExtensionResult
-from ulauncher.api.result.extension_small_result import ExtensionSmallResult
+from ulauncher.api.result import Result
+
+
+class ExtensionResult(Result):
```
friday, line 7: This is only temporary because I ported extensions to v6 using ExtensionResult and ExtensionSmallResult. Before release these should be removed.

PR #1129, opened 2022-09-22
Addes zstd to deal with control.tar.zst

eXtrem0us:
### Checklist

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -27,7 +27,8 @@ Depends: ${misc:Depends},
  gir1.2-gdkpixbuf-2.0,
  python3-gi,
  python3-gi-cairo,
- python3-levenshtein
+ python3-levenshtein,
+ zstd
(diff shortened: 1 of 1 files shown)


friday:
zstd is not a dependency for Ulauncher. It's only the Jammy release that uses it because the packaging command defaults to that for Jammy. So I guess it should be a build dependency specifically for Jammy maybe, but all recent Ubuntu releases (at least since 18.04) comes with zstd, and our instructions for debian users is to use the Focal package that doesn't require zstd.

friday:
I decided to merge it anyway. Better to have too many dependenccies than too few I think. I'll move it to dev-dependency though.

eXtrem0us:
You where right @friday 
It wasn't necessary to add zstd to dependencies. But yeah. I encountered the issue for installing deb package on Debian and even after installing zstd on it, the issue where persisted.

Issue #701, opened 2021-03-05
Wayland Fullscreen behaviour


Iss-in:
AFAIK for now ulauncher behaves as a toolkit, ( same as other floating windows) . But in given wayland DEs are implementing their specific [layer_shell_protocol](https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-layer-shell-unstable-v1.xml),  ( implemented in sway, and likely to be merged soon in plasma [too](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/28) ), launcher will always be hidden behind a fullscreen window.

Will it be possible to make use of this protocol and make ulauncher always use the top layer to be shown above fullscreen ?

hhurma:
I am using a manjaro gnome. Ulauncher does not appear on the desktop. But if I open a fullscreen app it shows up.

Jaakkonen:
https://github.com/wmww/gtk-layer-shell has a python example https://github.com/wmww/gtk-layer-shell/blob/master/examples/simple-example.py of using `wlr_layer_shell_unstable_v1` for PyGTK windows.

PR #1172, opened 2022-12-29
feat: Try to use wlr-layer-shell protocol on wayland

Maetveis:
Uses https://github.com/wmww/gtk-layer-shell to set up the Ulauncher window as a layer window, which drops decorations, allows positioning and displaying over fullscreen windows.
Fails gracefully if the module cannot be loaded.
The extension is currently supported by wlroots based desktops (e.g Sway) and KDE.
Moving the window to a different position is not supported in this mode yet, but possibly could be implemented by changing the margin and or anchor edges.

Requires Ulauncher to be using wayland i.e. explicitly setting `GDK_BACKEND=wayland`.
Q: Should I add some notes about this "experimental" option and how to change e.g. the systemd user service file?

Fixes: https://github.com/Ulauncher/Ulauncher/issues/701

### Checklist

### DIFF
--- a/ulauncher/ui/LayerShell.py
+++ b/ulauncher/ui/LayerShell.py
@@ -1,0 +1,58 @@
+import logging
+import gi
+
+from gi.repository import Gtk
+
+logger = logging.getLogger()
+
+try:
+    gi.require_version('GtkLayerShell', '0.1')
+    # pylint: disable=ungrouped-imports
+    from gi.repository import GtkLayerShell
+except (ValueError, ImportError):
+    logger.warning("Failed to load GtkLayerShell, disabling positioning on wayland!")
+    GtkLayerShell = None
+
+
+class LayerShellOverlay(Gtk.Window):
+    '''
+    Allows for a window to opt in to the wayland layer shell protocol.
+
+    This Disables decorations and displays the window on top of other applications (even if fullscreen).
+    Uses the wlr-layer-shell protocol [1]
+
+    [1]: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml
+    '''
+
+    @classmethod
+    def is_supported(cls, *, quiet=False):
+        '''Check if running under a wayland compositor that supports the layer shell extension'''
+        supported = GtkLayerShell is not None and GtkLayerShell.is_supported()
+        if not quiet:
+            logger.info("Wayland layer shell extension: %s", ("supported" if supported else "not supported"))
+        return supported
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._use_layer_shell = False
+
+    def enable_layer_shell(self):
+        assert __class__.is_supported(quiet=True), "Should be supported to enable"
+        self._use_layer_shell = True
+
+        GtkLayerShell.init_for_window(self)
+        GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.EXCLUSIVE)
+        GtkLayerShell.set_layer(self, GtkLayerShell.Layer.OVERLAY)
+        # Ask to be moved when over some other shell component
+        GtkLayerShell.set_exclusive_zone(self, 0)
+
+    @property
+    def layer_shell_enabled(self):
+        return self._use_layer_shell
+
+    # pylint: disable=unused-argument
+    def set_vertical_position(self, pos_y):
+        # Set vertical position and anchor to the top edge, will be centered horizontally
+        # by default
+        GtkLayerShell.set_anchor(self, GtkLayerShell.Edge.TOP, True)
+        GtkLayerShell.set_margin(self, GtkLayerShell.Edge.TOP, pos_y)
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -1,11 +1,11 @@
 # -*- Mode: Python; coding: utf-8; indent-tabs-mode: nil; tab-width: 4 -*-
 import logging
-import gi
 from gi.repository import Gtk, Gdk, Keybinder  # type: ignore[attr-defined]
 
 # pylint: disable=unused-import
 # these imports are needed for Gtk to find widget classes
 from ulauncher.ui.ResultWidget import ResultWidget  # noqa: F401
+from ulauncher.ui.LayerShell import LayerShellOverlay
 from ulauncher.config import PATHS
 from ulauncher.ui.ItemNavigation import ItemNavigation
 from ulauncher.modes.ModeHandler import ModeHandler
@@ -18,19 +18,9 @@ from ulauncher.utils.Theme import Theme
 
 logger = logging.getLogger()
 
-try:
-    gi.require_version('GtkLayerShell', '0.1')
-    # pylint: disable=ungrouped-imports
-    from gi.repository import GtkLayerShell
-except (ValueError, ImportError):
-    logger.warning("Failed to load GtkLayerShell, disabling positioning on wayland!")
-    GtkLayerShell = None
-
-
-class UlauncherWindow(Gtk.ApplicationWindow):
+class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
     _css_provider = None
     _drag_start_coords = None
-    _layer_shell_enabled = False
     results_nav = None
     settings = Settings.load()
 
@@ -48,15 +38,8 @@ class UlauncherWindow(Gtk.ApplicationWindow):
             window_position="center",
         )
 
-        # Check if running under a wayland compositor that supports the layer shell extension
-        if GtkLayerShell is not None and GtkLayerShell.is_supported():
-            logger.info("Layer shell extension is supported, using to position window")
-            self._layer_shell_enabled = True
-            GtkLayerShell.init_for_window(self)
-            GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.EXCLUSIVE)
-            GtkLayerShell.set_layer(self, GtkLayerShell.Layer.OVERLAY)
-            # Ask to be moved when over some other shell component
-            GtkLayerShell.set_exclusive_zone(self, 0)
+        if LayerShellOverlay.is_supported():
+            self.enable_layer_shell()
 
         # Try setting a transparent background
         screen = self.get_screen()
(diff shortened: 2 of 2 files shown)


friday:
Very much appreciated. I didn't know that you could do that.

I'm on Gnome so I can't test it myself without virtualization or installing another desktop environment, and also don't know this API at all, so I might just have to take your word for that it works.

Could you make a new abstraction for this in a separate module, similar to AppIndicator? I'm good with the code inside `position_window`, but the stuff above that I think should all be in a separate module.

Maetveis:
> I'm on Gnome so I can't test it myself without virtualization or installing another desktop environment, and also don't know this API at all, so I might just have to take your word for that it works.

I'll get some screenshots later with it enabled / disabled on Sway. However if you do decide to try it out you don't need full virtualization to do it: [The wayland socket can be forwarded to docker containers](https://github.com/mviereck/x11docker/wiki/How-to-provide-Wayland-socket-to-docker-container). Then you can launch another compositor nested inside the original (i.e. it will display as a window in Gnome).

> Could you make a new abstraction for this in a separate module, similar to AppIndicator? I'm good with the code inside `position_window`, but the stuff above that I think should all be in a separate module.

ulauncher/ui/LayerShell.py
```diff
@@ -0,0 +1,58 @@
...
+    This Disables decorations and displays the window on top of other applications (even if fullscreen).
+    Uses the wlr-layer-shell protocol [1]
+
+    [1]: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml
+    '''
+
+    @classmethod
+    def is_supported(cls, *, quiet=False):
+        '''Check if running under a wayland compositor that supports the layer shell extension'''
+        supported = GtkLayerShell is not None and GtkLayerShell.is_supported()
```
friday, line 30: Nice :+1:  Had exactly something like this in mind for the supported method to work regardless :)

ulauncher/ui/windows/UlauncherWindow.py
```diff
@@ -17,8 +18,7 @@
 
 logger = logging.getLogger()
 
-
-class UlauncherWindow(Gtk.ApplicationWindow):
+class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
```
friday, line 21: I didn't think of that, but like it :+1:

friday:
Looks great now :+1: I would take this PR as is, but I have two thoughts:

1. Ulauncher v6 supports dragging the windows to move it
https://github.com/Ulauncher/Ulauncher/blob/HEAD/ulauncher/ui/windows/UlauncherWindow.py#L209-L218 Can this be supported in wl-roots also?

2. If 1. can be supported I think the wl-roots experience will be able to handle everything Xwayland can do except "Switch to application if it's already running", which only works for apps running in Xwayland (and we don't want to force that). Then I think we can default to pure Wayland for wlroots, and Xwayland for others?

Maetveis:
Draft as the reposition experience on kwin wayland is pretty bad with this.

Maetveis:
> 1. Ulauncher v6 supports dragging the windows to move it
>    https://github.com/Ulauncher/Ulauncher/blob/HEAD/ulauncher/ui/windows/UlauncherWindow.py#L209-L218 Can this be supported in wl-roots also?

I tried implementing this and altough it works on Sway with kwin it bounces between two positions when dragged around.
The final position is ok, so I think it has to do with kwin updating the window position during the drag (to which the mouse move coordinates are relative to).
The whole reposition thing is rather hacky IMO, and I would much rather see a proper configuration workflow for this like:

1. An option in preferences to set the window position e.g. default / custom
2. For custom a button that draws the launcher window with its maximum size and can be dragged around to position it.
3. After confirming the new position it gets saved and the window now always pops up at that position.

I don't think something like this is in scope for this PR tough, so I'd much rather just drop positioning from it for now.

> 2. If 1. can be supported I think the wl-roots experience will be able to handle everything Xwayland can do? Then I think we can default to pure Wayland for wlroots, and Xwayland for others?

Idk, are there other window related features that I should test? I've been using Ulauncher with these changes over the holidays and have not seen anything missing (altough I don't use many extensions for example).

> One exception to 2 is "Switch to application if it's already running" only works for apps running in X/Xwayland (and we don't want to force that), so I think this is fundamentally broken for Wayland anyway, even with Xwayland.

This could be supported using the [wlr-foreign-toplevel-management extension] e.g. via `pywayland`

[wlr-foreign-toplevel-management extension]: https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1

friday:
> The whole reposition thing is rather hacky IMO

It's a bit hacky yeah. I don't remember exactly why now, but I originally tried to use [`Gtk.Window.begin_move_drag()`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Window.html#Gtk.Window.begin_move_drag) when I implemented it, but this worked much worse and didn't give enough low level control or something. Both APIs are X11 only anyway. There's a [GTK4 API](https://docs.gtk.org/gdk4/method.Toplevel.begin_move.html) one that might work, but  we use GTK3. However it might be worth looking into trying to use GTK4 conditionally.

> The whole reposition thing is rather hacky IMO, and I would much rather see a proper configuration workflow for this like:
> 
> 1. An option in preferences to set the window position e.g. default / custom
> 2. For custom a button that draws the launcher window with its maximum size and can be dragged around to position it.
> 3. After confirming the new position it gets saved and the window now always pops up at that position.

I disagree. Just being able to drag the window without persisting it is great if you for example want to manually copy something from the window below Ulauncher (ex: using the calculator and entering data from the screen). You can't do that by moving the window below (that closes Ulauncher).

Being able to persist that would be great, but it shouldn't be automatic, and you shouldn't have to go to the settings to move it.

> I don't think something like this is in scope for this PR tough, so I'd much rather just drop positioning from it for now.

That's very reasonable :+1: Thank you for trying it. I wasn't sure how big it would be when I asked, and it was mostly to make it easier to decude whether to default to pure Wayland or Xwayland for wl-roots.

Why do you not want to use Xwayland anyway? Is it for security/performance or rendering blurry? I've never been able to reproduce the latter myself.

Maetveis:
> > The whole reposition thing is rather hacky IMO
> 
> It's a bit hacky yeah. I don't remember exactly why now, but I originally tried to use [`Gtk.Window.begin_move_drag()`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Window.html#Gtk.Window.begin_move_drag) when I implemented it, but this worked much worse and didn't give enough low level control or something. Both APIs are X11 only anyway.

Interesting, because `begin_move_drag` works for me on Sway, but `move` just does nothing with both native wayland and Xwayland.

> There's a [GTK4 API](https://docs.gtk.org/gdk4/method.Toplevel.begin_move.html) one that might work, but we use GTK3. However it might be worth looking into trying to use GTK4 conditionally.
 
That api looks the exact same as `begin_move_drag` just on a different object (toplevel vs window) so I wouldn't expect it to give finer control.
 
> I disagree. Just being able to drag the window without persisting it is great if you for example want to manually copy something from the window below Ulauncher (ex: using the calculator and entering data from the screen). You can't do that by moving the window below (that closes Ulauncher).

That makes sense, I wasn't thinking about this.

> Being able to persist that would be great, but it shouldn't be automatic, and you shouldn't have to go to the settings to move it.

Yep makes sense with the usecase in mind.

> Why do you not want to use Xwayland anyway? Is it for security/performance or rendering blurry? I've never been able to reproduce the latter myself.

The window pops up at an awkard position in the dead centre of the screen in Sway, and the result list extends outside of it. I know Sway is special with window positioning and I can probably set up rules for Ulauncher in my configuration to make it wok, but wanted to help by making it work "out of the box".

I am not a wayland elitist by far, but there's also the angle that new (or actively maintained) software shouldn't keep relying on deprecated stuff, because things are bound to break.

TBH it was also a good excuse to learn a bit about how things work in wayland while working on this.

> > I don't think something like this is in scope for this PR tough, so I'd much rather just drop positioning from it for now.
>
> That's very reasonable 👍 Thank you for trying it. I wasn't sure how big it would be when I asked, and it was mostly to make it easier to decude whether to default to pure Wayland or Xwayland for wl-roots.

How do we proceed then, should I drop the two new latest commits or just the positioning and leave the `on_mouse_down` fix? (The drag won't work on Sway anyway, but at least it doesn't show the drag icon when clicking on results.)

Also maybe just because the extension is available it probably shouldn't be enabled by default (for users that alredy opt in to native wayland) as it breaks moving the window around.

If it had an option to enable it, then maybe the current positioning could also be kept, it does work on Sway at least.

friday:
> Interesting, because `begin_move_drag` works for me on Sway, but `move` just does nothing with both native wayland and Xwayland.
> 
> Could it have been to do with the window margin? When checking out my changes in weston it wouldn't let me drag to the edge of the screen because of the margin drawn for the shadow. If thats the case then maybe [`set_shadow_width`](https://docs.gtk.org/gdk3/method.Window.set_shadow_width.html) could work, or drawing the shadow in a different way.

The shadow had nothing to do with it, but I don't remember exactly. It was [a year](https://github.com/Ulauncher/Ulauncher/pull/957) ago. Let's explore this later. If you want to look at switching to `begin_move_drag` that would be welcome, but that's even more so outside the scope of this PR.

About my other question. It's not about X11 vs Wayland. I just want to understand a non-purist argument for what to default to on wl-roots now (XWayland vs pure Wayland). From my obserations before this PR Ulauncher in pure Wayland had 3 issues that Ulauncher in XWayland didn't have: positioning the window (which this PR implements for wl-roots), dragging the window (which we might be able to solve) and displaying the window as "always on top" aka ["keep above"](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Window.html#Gtk.Window.set_keep_above), which I think is not nearly as important as the other two.

If we can solve all but "always on top", and pure Wayland has other benefits that could be argued to be more important, I would be open to changing the defaults. I know of this: https://www.reddit.com/r/gnome/comments/x1qbd2/has_anyone_solved_blurry_xwayland_apps_with/ but I think that is something that would happen for Ulauncher or just apps without HiDPI support and/or if it can be worked around.

> How do we proceed then, should I drop the two new latest commits

Yes. Let's go with that :+1: I think the `on_mouse_down` fix is outside the scope of this PR, and it breaks because the window positioning includes the shadow (I assume you disabled that). So when you get the eventbox positioning and apply that to the window it'll move it down and right, just for grabbing it and moving a pixel.

I would welcome a separate PR for that. Preferable with a fix for the shadow, but if not, I could have a look at that part since it's easy for me to reproduce.

> Also maybe just because the extension is available it probably shouldn't be enabled by default (for users that already opt in to native wayland) as it breaks moving the window around.

Yeah, that's what I was wondering about before, but let's keep the defaults and try `begin_move_drag` instead, and if that works we can revisit the defaults.

friday:
Btw, I considered redesigning Ulauncher so the top part with the entry field and settings button would be a GTK header bar. I never really explored that, but that would give us the "drag to move" behavior for free by the window manager, and the window manager would also handle the shadows (for better and worse as we would not be able to customize the shadow appearance, but it would be degrading gracefully as that's also handled by the window manager.

This might not be possible to style the way we would want though and/or might break current themes.

friday (approved):
Looks great. High quality PR :+1: :pray:

PR #1173, opened 2023-01-01
ci: check workflow

friday:
Adds a github action/workflow for checking/validating commits.

This integrates better with GitHub than the travis checks, so those (and the `run-ci` command) was removed. It also checks multiple versions of Python, but I only added 3.8 and 3.11. Can't go older than that because you need [Ubuntu 18.04 😵 (unsupported)](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/) for 3.6.

### Checklist


friday:
Merged, but somehow this PR wasn't closed, so closing manually

PR #1178, opened 2023-01-06
fix: Fix dragging when clicking on results

Maetveis:
The event coordinates are relative to the widget being clicked on instead
of the window. Therefore it was possible to activate dragging by clicking
on the results.
Add an EventBox widget around the draggable area and subscribe to the
click event only on that instead of the whole window in order to fix it.

Taken from #1172 + fixed the coordinate relative to be relative to the window. 

### Checklist


friday (approved):
Works perfectly on my end. Thank you :+1:

Issue #1151, opened 2022-11-06
Open a new window every time


friday:
### Description

Why:
1. If the theme is loaded every time the window opens, it means we don't have to reload the window after changing the preferences, and you can edit the theme and without the [NOHUP-hack to force Ulauncher to reload](https://docs.ulauncher.io/en/stable/themes/themes.html#creating-custom-color-themes).
2. We need it for #1304
3. It would be a nice clean state to run `Gio.DesktopAppInfo.get_all()` from only once, then free the memory after closing (now we run it on every keypress).

I have already tested this (via an ugly workaround), and I couldn't see any performance differences.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

friday:
Reopening because of #1338

PR #1179, opened 2023-01-07
feat: use begin_move_drag for moving window

friday:
As per discussion in https://github.com/Ulauncher/Ulauncher/pull/1172#issuecomment-1371922928

I don't remember why I couldn't use this before. It may have been related to other changes we have made since.

### Checklist

### DIFF
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -69,9 +69,9 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
         )
         window_frame.pack_start(window_container, True, True, 0)
 
-        self.event_box = Gtk.EventBox()
+        event_box = Gtk.EventBox()
         input_box = Gtk.Box()
-        self.event_box.add(input_box)
+        event_box.add(input_box)
 
         self.input = Gtk.Entry(
             can_default=True,
@@ -109,7 +109,7 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
         self.result_box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
         self.scroll_container.add(self.result_box)
 
-        window_container.pack_start(self.event_box, True, True, 0)
+        window_container.pack_start(event_box, True, True, 0)
         window_container.pack_end(self.scroll_container, True, True, 0)
 
         window_container.get_style_context().add_class("app")
@@ -123,7 +123,7 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
 
         self.connect("focus-in-event", self.on_focus_in)
         self.connect("focus-out-event", self.on_focus_out)
-        self.event_box.connect("button-press-event", self.on_mouse_down)
+        event_box.connect("button-press-event", self.on_mouse_down)
         self.input.connect("changed", self.on_input_changed)
         self.input.connect("key-press-event", self.on_input_key_press)
         prefs_btn.connect("clicked", lambda *_: self.app.show_preferences())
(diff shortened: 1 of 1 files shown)


friday:
Does this work for you @Maetveis?

Maetveis:
> Does this work for you @Maetveis?

| Compositor                       | Status                                      |
|-------------------------------------|----------------------------------------|
|Sway XWayland                  | fixed (did not work before)|
|Sway layer shell                 | doesn't work                         |
|Sway layer shell disabled | fixed                                       |
|KWin Xwayland                  | works (didn't break)            |
|Kwin layer shell                  | doesn't work                        |
|Kwin layer shell disabled | fixed                                       |

Layer shell "disabled" means I commented out the code for it, to use GTK without the layer shell extension. Overall this fixes some paltforms and simplifies the code so +1 from me, but it doesn't help to achive feature parity on wlroots which is what I guess you'd hoped for.

ulauncher/ui/windows/UlauncherWindow.py
```diff
@@ -189,27 +187,10 @@ def on_input_key_press(self, widget, event):
 
     def on_mouse_down(self, _, event):
         """
-        Prepare moving the window if the user drags
+        Move the window on drag
         """
         if event.button == 1:
-            self.set_cursor("grab")
-            x, y = self.event_box.translate_coordinates(self, event.x, event.y)
```
Maetveis, line 196: The `self.event_box` attribute could be dropped I guess it has no uses anymore.

friday:
> Layer shell "disabled" means I commented out the code for it, to use GTK without the layer shell extension. Overall this fixes some paltforms and simplifies the code so +1 from me, but it doesn't help to achive feature parity on wlroots which is what I guess you'd hoped for.

So if I understand this correctly then the support should be something like this now:

|                                     | X11 | XWayland (GDK_BACKEND=x11) | Wayland (GDK_BACKEND=wayland) | Layer Shell (GDK_BACKEND=wayland) |
|-------------------------------------|-----|----------|----------------------------------|-----------------------------------|
| Global hotkey                       | ✔️   | ❌        | ❌                                | ❌                                 |
| Center top positioning              | ✔️   | ✔️        | ❌                                | ✔️                                 |
| Movable window                      | ✔️   | ✔️        | ✔️                                | ❌                                 |
| "Always on top"                     | ✔️   | ✔️        | ❌                                | ✔️                                 |
| Launch on the display with your cursor | ✔️   | ❌        | ❌                                | ❌                                 |

The method we use for global hotkeys can't be supported in Wayland (for very good reasons), although the new global shortcut portal should let us do that eventually, once implemented in more desktop environments.

For the last one Ulauncher has an option if you want to launch on the default display or the one with the cursor. The Ulauncher window is never closed, so this is handled by using an X11-only method to get the cursor position, use that to see which screen the cursor is at and move the window. While XWayland can handle the positioning, it can't get the cursor position unless the cursor is over an active XWayland window. I think the way to fix this is #1151, and let the window manager decide which display to launch on. In Gnome at least that's the one with the cursor.

I think that should fix it for all environments, unless they default to something else, or let the users decide. We can still support the "default display" option for X11 and XWayland.

Maetveis:
> So if I understand this correctly then the support should be something like this now:

Yes pretty much, I can't test the last one now, but it should behave as you say with the layer surface as well based on the 
[`get_layer_surface` description]:
> You may pass NULL for output to allow the compositor to decide which output to use. Generally this will be the one that the user most recently interacted with.

[`get_layer_surface` description]: https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_shell_v1:request:get_layer_surface

Also there was the reactivate window if already open, which is only partially supported on XWayland (only X11 windows).

friday:
> Also there was the reactivate window if already open, which is only partially supported on XWayland (only X11 windows).

Yeah, but that one is rather complex to fit in such a table. Would require different columns. And it can be solved by disabling focus-stealing prevention in the DE/WM.

Issue #1177, opened 2023-01-02
Rewrite search highlighting so it can use css


friday:
### Description

Currently themes use a css file an a manifest json file. The latter is almost completely useless.

```json
{
    "manifest_version": "1",
    "name": "dark",
    "display_name": "Elementary Dark",
    "extend_theme": "light",
    "css_file": "theme.css",
    "matched_text_hl_colors": {
        "when_selected": "#fff",
        "when_not_selected": "#fff"
    }
}
```

It contains information we don't need like "name" (just use the css file name), but it has two things we do need: `extend_theme` for inheritance and `matched_text_hl_colors` for highlighting.

The former can be solved in other ways (including reset.css for all themes so they don't have to inherit and/or adding a special css import syntax)

The `matched_text_hl_colors` are needed because of an implementation detail. The result uses a regular css label with "pango html" support, and we're adding spans. This can't be styled with css. We should instead switch to using a horizontal Gtk Box with multiple labels as children.

This way themes could use pure css and be easier to write.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

Laurens256:
I *almost* got this working.The only part of this I don't see being possible is how to ellipsize the text if it gets too long. Since there would be multiple labels, and there's no ellipsize attribute for `Gtk.box`, the text doesn't ellipsize correctly and the container could potentially get super wide for long titles

https://github.com/Laurens256/Ulauncher/tree/highlighting-rewrite

friday:
You can make a draft PR if you want @Laurens256. That way makes it easier to share, test and collaborate :).

I did test it though and it worked very well.

I didn't consider the ellipsize problem when I created the issue. It can't be solved in the same way as before, but I think if the unhighlighted labels ellipsize, then it's an improvement from before. I haven't tested that though. The second (description) row is always longer for me, so usually that's the only one that gets shortened.

If you want to change the abstraction a bit more and have an idea of how you want to do that, also feel free to do that. The new behavior warrants a different abstraction and terminology in the methods imo. It doesn't have to be part of this PR, but if you already have ideas for that definitely don't feel like you have to respect the old method names.

[...]

Issue #1148, opened 2022-11-05
Convert ResultWidget to not use Glade


friday:
### Description

After merging #1147 this is now the last/only glade ui/xml file (or actually two files). Would be nice to get rid of it as well and simplify the result widget generation by making this code fully dynamic without the glade file. We're already overriding the margins a lot, to get a nicer adaptive appearance, so it would make more sense from that perspective too.

This is not important, but should be a low hanging fruit by now.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

Issue #931, opened 2021-12-14
Add generic extension method to handle activating results instead of actions


friday:
### Description

Problems now:
1. The actions abstraction is not very straight forward. You need to always send an action (even to do nothing). You need a wrapper action to send/run multiple actions, and while you can create a custom callback with `ExtensionCustomAction` you can't run the other actions from within it. You can only do something and return the action at the end (which makes sense for closing the window but not a lot else).
2. Why limit to only "enter" and "alt+enter"? What if you want to have a "Delete" action when you press Delete or backspace (ex: https://github.com/friday/ulauncher-clipboard/issues/5)? Or just multiple options: #326 #476

### Plan
- [x] C. Improve ExtensionCustomAction and `on_item_enter` handler
    * Make sure you can extend the result with custom data you need for the handler (reserved custom "data" property or adding any properties? Both have pros and cons).
    * It should take two argument: the `Result` object and an "action" type/name/id.
    * Add a new list property "actions" to the Result class which should contain a list of 1-2 named actions, ex `["Open", "Open in File Manager", "Copy path to clipboard"]`. When this is present, we should ignore `on_enter` and `on_alt_enter` and trigger the new method (implemented in #1613)
- [x] D. Add alternatives to CopyToClipboardAction and OpenAction that are functions that can be called from `on_activate` or `on_item_enter` and handles everything for you (sending the message via dbus to Ulauncher) and waiting for a response. Not done, but moved to #1622 with slightly spec
- [x] E. Mark `ExtensionCustomAction` for deprecation (don't want to turn on all the annoying Deprecation warnings in the first stable release) Not done, but moved to #1582
- [ ] F. Update the action docs
- [ ] G. Extend step C with the ability to set custom keybindings? But how... This step is problematic if we allow extension devs to override/hijack keybindings and/or set inconsistent ones, making it hard to learn. Also, consider the option for end-users to override these (this adds another level of complexity): Suggestion: Just let the first one in the list be the default ("enter" key) and if someone makes one action named "Delete", we can use the delete key.
- [ ] H. Consider how to avoid new API additions with hard coded English language that can't be translated (like avoiding the hard coded array of actions, maybe these should be ids instead, with the titles in the manifest). 

Also related to: https://github.com/Ulauncher/Ulauncher/issues/662#issuecomment-994773059 and https://github.com/Ulauncher/Ulauncher/discussions/966

friday:
[...]

Issue #869, opened 2021-12-02
Ulauncher v6


friday:
### Description

This is a tracking issue for the v6 release that we have been working hard on. V6 is still in development and will not be released for quite a while due to containing major changes (and major improvements) and more changes planned.

The [6.0.0](https://github.com/Ulauncher/Ulauncher/milestone/7) milestone has a lot more information. It lists a lot of breaking changes, but they are mostly limited in scope such as moving niche CLI options, files for storing users settings and removing compatibility with very old distro releases.

This issue was created so that people interested in the progress can subscribe to the issue to be notified, since I will be closing issues that are fixed in the v6 branch and redirecting people here if they want a notification when we release or something similar.

### Feature isn't already requested

- [X] I have checked the existing feature requests before suggesting this

joseph-zhong:
Huge shoutout and thanks to the team for working on this! Real excited for the v6 release; been using v5.x for only a few days and been a huge game-changer; particularly excited for the improvements in this release and also happy to beta-test -- is there suggested documentation for building/running from source?

friday:
Thank you @joseph-zhong 

> Is there suggested documentation for building/running from source?

Yes: https://github.com/Ulauncher/Ulauncher/blob/v6/CONTRIBUTING.md

friday:
You should be able to build a Debian package now with `./ul build-deb`, assuming you are on Debian and have the packages needed (see above for the general build dependencies and add `apt install debhelper dh-python rsync` for the debian package build.

LinuxOnTheDesktop:
_Blacklisted dirs_ ('If you still feel like you need this setting in Ulauncher v6, let us know'). I think I do need this. For, there are directories that contain executables (or links to them) that do not work on my system or that I do not wish to use.

Also: could we have _dark mode in the configuration windows_, please?

friday:
> _Blacklisted dirs_ ('If you still feel like you need this setting in Ulauncher v6, let us know'). I think I do need this. For, there are directories that contain executables (or links to them) that do not work on my system or that I do not wish to use.

Then why do you have them in the dedicated directory for desktop entries to use for launchers and menu apps if I may ask?

> Also: could we have _dark mode in the configuration windows_, please?

That's not planned, and is not easy to implement, so very unlikely for v6. Hopefully later at some point.

LinuxOnTheDesktop:
[...]

PR #1219, opened 2023-04-08
Add Info Window for displaying additional information from extensions

DraggonFantasy:
This pull request introduces a new feature: the Info Window. The Info Window is a non-focusable, borderless window that can be displayed alongside search results. It is designed to provide additional information for selected search results, such as previews, descriptions, or other relevant details.

The Info Window can be utilized by extensions to display custom HTML content related to the selected search result. As a demonstration of this new functionality, I have implemented an extension that searches Wikipedia. When a user selects a Wikipedia search result, a preview of the Wikipedia page is displayed in the Info Window, allowing the user to quickly review the content before opening the full page in a web browser.

Here's a screenshot showcasing the Info Window with the Wikipedia search extension:

This feature will enhance the user experience by allowing them to access more information without needing to open a separate application.

Please review the changes and let me know if you have any suggestions or concerns.

### DIFF
--- a/docs/extensions/actions.rst
+++ b/docs/extensions/actions.rst
@@ -56,6 +56,11 @@ SetUserQueryAction
 .. autoclass:: ulauncher.api.shared.action.SetUserQueryAction.SetUserQueryAction
 
 
+UpdateInfoAction
+----------------
+
+.. autoclass:: ulauncher.api.shared.action.UpdateInfoAction.UpdateInfoAction
+
 
 .. NOTE::
   Please take `a short survey <https://goo.gl/forms/wcIRCTjQXnO0M8Lw2>`_ to help us build greater API and documentation
--- a/docs/extensions/tutorial.rst
+++ b/docs/extensions/tutorial.rst
@@ -229,6 +229,48 @@ Custom Action on Item Enter
   Now this will be rendered when you click on any item
 
 
+Using Information Window
+------------------------
+The Information Window Feature provides the ability to display HTML content in an information window within Ulauncher.
+This allows you to provide a richer and more detailed context for your Ulauncher extension.
+
+.. figure:: https://i.imgur.com/fVGkSj5.png
+  :align: center
+
+  Information Window example
+
+To display HTML content in the information window, you can either use the **UpdateInfoAction** or add two
+parameters ``info`` and ``base_url`` after the result list when creating the **RenderResultListAction**.
+
+The UpdateInfoAction accepts the following parameters:
+
+UpdateInfoAction(info, base_url="file:///", keep_app_open=False):
+ - info (required): The HTML content you want to display in the information window.
+ - base_url (optional): The base URL for resolving relative URLs within the HTML content. Defaults to "file:///".
+ - keep_app_open (optional): Determines if the Ulauncher app window should stay open when the action is executed. Defaults to False.
+
+  ::
+
+    def on_event(self, event, ext):
+        query = event.get_argument()
+        if not query:
+            return
+
+        search_results = self.search_wikipedia(query)
+        if not search_results:
+            return
+
+        items = [ExtensionResultItem(icon="images/icon.svg",
+                                     name=result["title"],
+                                     description=result["snippet"],
+                                     on_enter=OpenUrlAction(result["url"]),
+                                     on_alt_enter=CopyToClipboardAction(result["url"]),
+                                     on_select=UpdateInfoAction(self.fetch_wikipedia_html(result["url"]),
+                                                                "https://en.wikipedia.org")) for result in search_results]
+
+        return items
+  Example of using UpdateInfoAction to show a preview of wikipedia page when an item is selected
+
 
 .. NOTE::
   Please take `a short survey <https://goo.gl/forms/wcIRCTjQXnO0M8Lw2>`_ to help us build greater API and documentation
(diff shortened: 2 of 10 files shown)


friday:
Thank you for the PR, but you really should have checked first before making this effort as this is not something we are ready for at the moment.

To implement this well you need knowledge about the current priorities  (see #869), platform compatibility etc, and a lot of time. I don't want to add something that will not last, or that will create maintenance issues, or that won't work for all users. I want to replace the Extension Actions API completely, starting with v6 (#931). It is my top priority right now when I get time again. It should be backwards compatible for extensions, but I don't want to add anything new for the old API. I'm working on creating a new simplified and more consistent and opinionated extension API layer.

You also added  `on_select`. So extensions would be able to do anything when the user select an item. But do we want them to be able to do anything or just this? What if some extension goes super creative and decide that instead of pressing enter to trigger the action, they run the main action when you select, causing an inconsistent user experience (see #326 for what I think about this). Also, if the first row has `on_select`, how do you select it? Do you have to go down and up again with the arrow key or does it pop up without the user actually selecting it? In that case won't it also fire up that window lots of times when you type the query?

There are other issues with it being a window also. It's not really safe to assume you have space on the right side of the Ulauncher window, or that that's where you want it opened. In Wayland when you can't position windows, this would position itself on top of the regular Ulauncher window. Also, if you can't focus on the new window, can you select and copy text from it? You can't do that now either (only if the extension uses the copy action), but for "more details" I can certainly see why people would want to.

We also want to make Ulauncher use less ram, and WebKit2GTK is a resource hog, doesn't work for all users (#114) and keeps itself running the background even after you close the window (although possible to force-kill it: #1064).

I prefer the design in #893 more, or to add it below the list item. But either way whatever solution we go with should work well with the rest of the API. In the meanwhile I think this is not something critical. You can open a URL in a browser

I won't close this though. It can serve as a reference/example implementation for the future or lead to more discussion.

My take on this:
1. It has to wait until after the extension API changes are completed for v6.
2. It should be triggered by an intentional user interaction, not happen unexpected. For example: All extension items with "more info" the result could be marked in some way (ex: 🛈 o "ℹ️") so the user knows it has extra info, then they can click that or press a special dedicated keyboard combination to show it (ex "alt+i").
3. It should show in the existing window.
4. It's probably impossible not to use WebKit2GTK for this, but at least it should close down the process after (Ulauncher already suffers from performance issues that I'm trying to improve).

DraggonFantasy:
Thank you for such a detailed response. I didn't check the current priorities for this PR, because I've made this change primarily for myself as I'm going to use it with my own extensions made for my personal use, but also decided to share it with you or anyone who may want such function. I expected that you're probably not going to merge it now, because when I was creating the PR, I saw that you're not adding new features until v6 is ready. 

Maybe I'd take some of the issues you've mentioned and try to help with them if I have time, because I really like Ulauncher and want to make it better.

Regarding `on_select` - initially I thought to make it only capable of showing this extra-information window, but then I decided to make it more general and similar to `on_enter` or `on_alt_enter` to allow more "creative use", e.g. dynamic pagination of results or some fancy UX. I agree with you that it may be "too much", but it's just my opinion that it's nice when API allows more freedom, especially for such a software that is focused on personal performance/productivity. 

The first item's `on_select` is triggered automatically (no need to go down and again up) - that's what I experience with my example wikipedia extension. When I finish writing my query and get list of wikipedia articles - I immediately get the information window with the preview of the first article. It feels like a natural UX, because I immediately see that the items have extra information and feel that if I select other items I'll get the extra info for them as well.

Agree on your points about a window. 

1. Sure
2. I think it can be a setting like "show extra info on select", because I think some of the users would like to have the information shown immediately without need of extra steps to show it. If this setting is on - the info is shown on select, otherwise - it is shown on click/key press/whatever action from user side
3. Agree
4. Agree

friday:
I think about 3 we can probably reach some kind of agreement. I definitely wouldn't want to load a browser engine making a web request every time I press up/down to highlight results, for performance reasons. I also don't want to completely opt out of showing extra details though, in case I want to do it occasionally.

I think extra steps are great UX. If you make the steps easy and performant people will learn them in their muscle memory.

> Regarding on_select - initially I thought to make it only capable of showing this extra-information window, but then I decided to make it more general and similar to on_enter or on_alt_enter to allow more "creative use"

Ulauncher's extension API was originally designed like this, but that's what I want to get away from: https://blog.codinghorror.com/falling-into-the-pit-of-success/

Imo, when a user press enter, they should know what that does first. Same with alt-enter or any other action they can perform.

My vision for this:
* on_input should only yield results. Nothing else.
* on_item_enter is meant to be backwards compatible, but I want to add a replacement ex `on_item_activate` that should do the same, but have a stricter and simpler form:
1. If it returns a list of results, render that list.
2. If it returns a string, set the query to the string.
3. If it returns neither list or string it closes the app. To override this return True.
4. `OpenAction`, and `CopyToClipboardAction` should be regular functions that you can import and run inside of the method, instead of returning them. This way we no longer need `ActionList`
5. ExtensionCustomAction, DoNothingAction and HideWindowAction would also not be needed as a consequence of this, so it makes the API much easier.

> Maybe I'd take some of the issues you've mentioned and try to help with them if I have time, because I really like Ulauncher and want to make it better.

That would be very much appreciated. I think you did a very thorough job on this PR, so your help would be appreciated. The areas that I think is most in need of help though is stuff that is kind of hard to explain, but I went through the existing issues and marked a few of them with contributor friendly.

#273 is something I intentional have been leaving for others as I'm not a Python dev.
#1148 and #1177 are sort of isolated from the API stuff, but they still aren't easy.

PR #1230, opened 2023-05-29
refactor!: make results json serializable

friday:
Makes the Result instances JSON serializable.

It's also possible to set any property this way, which is something I plan to utilize in further extension API rework. However the current implementation allows overriding the methods I think. Not sure how to handle that

This is a breaking change since the current `Result` (before the PR) allows passing positional arguments. The extension documentation never mentioned that, and I haven't found a single extension written like that when checking the [github code search](https://github.com/search?q=ExtensionResultItem%28+language%3APython&type=code&l=Python). So I think it's fine as long as it's documented as a breaking change.

This is the first a step towards getting rid of pickle for extension communication. Everything we send as of this PR is JSON serializable except some of the actions: `OpenAction`, `CopyToClipboardAction`, `ActionList` and `ExtensionCustomAction` (the others actions have recently been made serializable already).

After that I plan to do a bigger rewrite of the extension communication, by using sending JSON messages over the existing application dbus connection instead of sockets (sorry @troycurtisjr). I tested it and it's much simpler and allows architectural improvements like #1020 that just I can't implement properly now because of the complex communication flow.

I will test the app for a while with this PR applied, so don't need help with that. Would like some input on the result class though. It's like a simplified dataclass, without the need for a decorator. I haven't noticed any change for the performance. I'm using `__getattribute__` instead of `__getattr__`, which is "dangerous", but I'm handling that by being stricter in the setter. I'm also iterating over the class (default) properties in a bit of a hacky way to make sure they are included in the data.

### DIFF
--- a/tests/modes/apps/extensions/test_DeferredResultRenderer.py
+++ b/tests/modes/apps/extensions/test_DeferredResultRenderer.py
@@ -1,3 +1,4 @@
+from types import MethodType
 from unittest import mock
 
 import pytest
@@ -31,7 +32,9 @@ class TestDeferredResultRenderer:
 
     @pytest.fixture
     def controller(self):
-        return mock.create_autospec(ExtensionController)
+        ctrl = mock.create_autospec(ExtensionController)
+        ctrl.get_normalized_icon_path = MethodType(lambda _: "/path/to/asdf.png", ctrl)
+        return ctrl
 
     @pytest.fixture
     def renderer(self):
--- a/ulauncher/api/result.py
+++ b/ulauncher/api/result.py
@@ -20,7 +20,9 @@ class Result(dict):
         for cls in reversed(self.__class__.__mro__):
             if cls in dict.__mro__:
                 continue
-            defaults = {k: v for k, v in vars(cls).items() if (not k.startswith("__") and not callable(v))}
+            defaults = {
+                k: v for k, v in vars(cls).items() if (not k.startswith("__") and not callable(getattr(cls, k)))
+            }
             self.update(defaults)
 
         # set values
--- a/ulauncher/api/result.py
+++ b/ulauncher/api/result.py
@@ -1,59 +1,30 @@
 import os
-from typing import Callable, Optional
 
 from ulauncher.api.shared.action.BaseAction import BaseAction
 from ulauncher.api.shared.query import Query
 from ulauncher.utils.fuzzy_search import get_score
 
-OnEnterCallback = Optional[Callable[[Query], Optional[BaseAction]]]
 
-
-# pylint: disable=too-many-instance-attributes
-class Result:
+class Result(dict):
     compact = False
     highlightable = False
     searchable = False
-    name: str = ""
-    description: str = ""
-    keyword: str = ""
-    icon: Optional[str] = None
-    _on_enter: Optional[OnEnterCallback] = None
-    _on_alt_enter: Optional[OnEnterCallback] = None
+    name = ""
+    description = ""
+    keyword = ""
+    icon = ""
 
-    # pylint: disable=too-many-arguments
-    def __init__(
-        self,
-        name: str = "",
-        description: str = "",
-        keyword: str = "",
-        icon: str = "",
-        highlightable: bool = False,
-        on_enter: OnEnterCallback = None,
-        on_alt_enter: OnEnterCallback = None,
-        searchable: Optional[bool] = None,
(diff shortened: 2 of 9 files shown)


troycurtisjr:
I'll definitely prioritize taking a look this week. Replacing pickle with JSON is a good thing overall. 

I'll be interested in seeing if dbus works well for our usage. I evaluated it very closely at one point for a work project (OMG, that was like 15 years ago at this point...👴), And there were a lot of throughout and latency issues. Basically the advice I came away with was that it was good for best effort signalling, but for any meaningful messaging volume it wasn't a great fit. Also since it is brokered, it is prone to delays on a busy system. I went looking for new info, and found some resources saying similar things, but so far they've been from the 2015 timeframe, so maybe things have improved a lot since then.

friday:
> I'll definitely prioritize taking a look this week. Replacing pickle with JSON is a good thing overall.
> 
> I'll be interested in seeing if dbus works well for our usage. I evaluated it very closely at one point for a work project (OMG, that was like 15 years ago at this point...older_man), And there were a lot of throughout and latency issues. Basically the advice I came away with was that it was good for best effort signalling, but for any meaningful messaging volume it wasn't a great fit. Also since it is brokered, it is prone to delays on a busy system. I went looking for new info, and found some resources saying similar things, but so far they've been from the 2015 timeframe, so maybe things have improved a lot since then.

dbus is slower for sure, but I don't think the performance will be noticeable for Ulauncher. It will remain to be seen though. Although I have tested it for more limited use with just the copy and open actions in the dbus branch, and that was not noticeably slower on my system at least: https://github.com/Ulauncher/Ulauncher/tree/dbus

Also overall we haven't tried to improve performance much. We still call the method to get all the desktop entries for each keystroke for example.

After writing this PR I got the idea that we might even use dicts directly. The Result class for extensions could just be replaced with a function that return a dict. I still like the Result class for our internal use though, and we still do need methods there. And I like this version of the class more than the earlier one.

troycurtisjr:
Overall it seems good as far as I can tell, but I was curious why not a few options:

dataclasses directly. Using a decorator seems preferable to adding a fair bit of code. In one of my work projects we've been using frozen dataclasses a lot and it is really nice. A quick look around seems to indicate the results are all created and used already, without being changed, so that would lend well to frozen dataclasses. I think there is a couple of `__init__` usages which would have to be adjusted for that to work directly, but it should be a small lift. Then Ulauncher doesn't need to maintain the low-level bits, it can be in the standard library.

Alternately, why not use the existing JsonData class and associated decorator. It looks like most of the low level function are nearly exact copies, so reusing the existing capabilities seems like it would be a good approach.

Any particular reason to not use one of the established functionalities?

friday:
> Overall it seems good as far as I can tell, but I was curious why not a few options:
> 
> dataclasses directly. Using a decorator seems preferable to adding a fair bit of code. In one of my work projects we've been using frozen dataclasses a lot and it is really nice. A quick look around seems to indicate the results are all created and used already, without being changed, so that would lend well to frozen dataclasses. I think there is a couple of `__init__` usages which would have to be adjusted for that to work directly, but it should be a small lift. Then Ulauncher doesn't need to maintain the low-level bits, it can be in the standard library.
> 
> Alternately, why not use the existing JsonData class and associated decorator. It looks like most of the low level function are nearly exact copies, so reusing the existing capabilities seems like it would be a good approach.
> 
> Any particular reason to not use one of the established functionalities?

Dataclasses doesn't exist in py36 (which we don't officially support, but still work with), and doesn't support `kw_only` until py310, which is problematic for inheritance: https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses. And even in py310+ you need to both declare the inheritance and use the decorator, and you can't add a custom property in the constructor without declaring it in the class, so it means if API devs need to make custom classes if they want to add custom data to the result instances (which is another key point of this change to enable).

JsonData solves most of these problems, but it has the same issues with inheritance and requiring a decorator, and it has methods we don't need, which I think would be confusing to extension devs. I think it's better to change JsonData to work more like this, and then make a lighter base class that both can inherit from. However, these shared methods don't add a lot of weight to the code base, so I didn't feel like it was a problem to copy them.

I think a better option than these though (for extension at least) might be to just use dicts directly, and replace the `Result` class with a dict factory method.

friday:
I separated the data class parts out to a new base data class and commented on the differences. Will merge and hopefully some time in the future get time to rewrite JsonData to be based on it.

Issue #283, opened 2018-09-07
Extensions in other languages


diogox:
I've vocalized this in another issue as a sort of off-topic, but figured it would be best to make it its own issue. 

**The idea**
Changing the architecture so that virtually any language could be used to make extensions, provided that someone creates a lib to communicate via websockets with the server. 

**How it could be done**
As far as I could tell, and I'm no python programmer (part of the reason I want this "feature"), all it would take is trading pickle for JSON serialization/deserialization. And changing how the server starts the extensions so that it is not exclusive to python. Possibly also the github download so that it compiles the code in case of compiles languages (that part could be specified in the manifest). 

I think this would really open a lot of possibilities and, in some cases, make the extensions faster. I'm available to make the Rust lib for this if it gets implemented.

gornostal:
That's a nice feature to have.
Before I look into that, I need to address a few more important things, like extension API stability, Python 3 support, file and network permissions for extensions.

Jaakkonen:
This would also possibly make it easier to use  `mypyc` and `cython` for compiling the extensions to Python C extensions which allows them to run faster and with a smaller memory footprint.

PR #1240, opened 2023-06-15
refactor: switch extension message format from pickle to json

friday:
I've been pushing underlying PRs and commits to enable this change recently: #1221, #1230, 6706467, ee73262 #1237 d8f65e2 a23ff49

It's working well for me, although pickle is apparently more compact, so the result is the messages are actually much longer now (sometimes more than twice as big).

I tested with ulauncher-clipboard and sending 200 items (not that my window is big enough to render them all). The result size is 54196 bytes instead of 22773 before this PR.

But that still wasn't noticably slow. JSON is much faster to serialize and parse (more than 10x) and it is generally better and a step towards supporting extensions in other languages #283

I'm not sure how much the PickleFramer class (now renamed) still makes sense though. It looks to me like it's more about the socket communication and switching from pickle to json doesn't really change anything there?

### DIFF
--- a/tests/api/client/test_Client.py
+++ b/tests/api/client/test_Client.py
@@ -17,7 +17,7 @@ class TestClient:
 
     @pytest.fixture(autouse=True)
     def framer(self, mocker):
-        return mocker.patch("ulauncher.api.client.Client.PickleFramer")
+        return mocker.patch("ulauncher.api.client.Client.JSONFramer")
 
     @pytest.fixture(autouse=True)
     def timer(self, mocker):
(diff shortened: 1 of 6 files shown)


troycurtisjr:
It looks good to me and seems to run fine on my machine. I'll keep running with it to see if I notice any issues.

Yes, the `PickleFramer` (now `JsonFramer`) is just responsible for taking the object and putting it into a form that can be sent over a byte stream and reconstructed on the other side. The only thing that makes it Json or Pickle specific is the encode/decode functionality which was correctly updated in the PR. If a message based system is used in the future, like what I think dbus is, then message framing is no longer needed, since the system will keep track of which bytes go to which messages natively.

troycurtisjr (approved):

friday:
Merged so I can get started on the DBUS POC. We can always revert changes if they lead to regression.

Issue #204, opened 2018-01-23
Vim like navigation


CaptainQuirk:
Hi there !

Could it be possible to navigate through the list with something else than the arrow keys ? Like `Ctrl-j/Ctrl-k` or `Ctrl-n/Ctrl-p` ?

Thanks

gornostal:
I'm not going to do this. Sorry.

I want to keep amount of code to the minimum, so I can still maintain the app on my own without getting lost in all of the features.
I'm planning to add only features that are either absolutely necessary or affect majority of users.

CaptainQuirk:
Thanks for your response

PR #1252, opened 2023-07-27
feat(#204): added ctrl-j/k navigation hotkeys

AlexNabokikh:
## Summary
The purpose of this change is to assist the Vim community by enabling the use of ctrl-j/k for navigating through Ulauncher results. This combination is widely recognized as a navigation hotkey and is currently supported by most Linux, Mac, and Windows launchers.

## Local testing
The change has been tested successfully on Ubuntu 23.04.

### DIFF
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -195,7 +195,7 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
             # input_changed can trigger when hiding window
             self.handle_event(ModeHandler.get_instance().on_query_change(self.app.query))
 
-    def on_input_key_press(self, widget, event) -> bool:  # noqa: PLR0911
+    def on_input_key_press(self, widget, event) -> bool:  # noqa: PLR0911, PLR0912
         """
         Triggered by user key press
         Return True to stop other handlers from being invoked for the event
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -242,21 +242,23 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
                 return True
 
         if self.results_nav:
-            if keyname in ("Up", "ISO_Left_Tab") or (ctrl and keyname == "p") or (ctrl and keyname == "k"):
+            if keyname in ("Up", "ISO_Left_Tab") or (ctrl and keyname == "p"):
                 self.results_nav.go_up()
                 return True
 
+            if keyname in ("Down", "Tab") or (ctrl and keyname == "n"):
+                self.results_nav.go_down()
+                return True
+
             if not self.emacs_bindings_active():
-                if (ctrl and keyname == "k"):
+                if keyname in (ctrl and keyname == "k"):
                     self.results_nav.go_up()
                     return True
-                if (ctrl and keyname == "j"):
+
+                if keyname in (ctrl and keyname == "j"):
                     self.results_nav.go_down()
                     return True
 
-            if keyname in ("Down", "Tab") or (ctrl and keyname == "n") or (ctrl and keyname == "j"):
-                self.results_nav.go_down()
-                return True
             if keyname in ("Return", "KP_Enter"):
                 result = self.results_nav.activate(self.app.query, alt=alt)
                 self.handle_event(result)
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from typing import Any
 
-from gi.repository import Gdk, Gtk, Gio, Keybinder  # type: ignore[attr-defined]
+from gi.repository import Gdk, Gio, Gtk, Keybinder  # type: ignore[attr-defined]
 
 from ulauncher.api.result import Result
 from ulauncher.config import PATHS
@@ -195,9 +195,21 @@ class UlauncherWindow(Gtk.ApplicationWindow, LayerShellOverlay):
             # input_changed can trigger when hiding window
             self.handle_event(ModeHandler.get_instance().on_query_change(self.app.query))
 
-    def emacs_bindings_active(self) -> bool:
-        gsettings = Gio.Settings.new("org.gnome.desktop.interface")
-        return gsettings.get_string("gtk-key-theme") == "Emacs"
+    def is_emacs_bindings_active(self) -> bool:
+        """
+        Checks whether the Emacs bindings are active in the GNOME.
+
+        Returns:
+            bool: True if the Emacs bindings are active, False otherwise.
+        """
+        try:
+            gsettings = Gio.Settings.new("org.gnome.desktop.interface")
+            key_theme = gsettings.get_string("gtk-key-theme")
+        except Exception as e:
+            logger.info("Failed to get the gtk-key-theme: %s", e)
+            return False
+
+        return key_theme.lower() == "emacs"
 
     def on_input_key_press(self, widget, event) -> bool:  # noqa: PLR0911
         """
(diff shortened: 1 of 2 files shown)


friday:
Thank you for this contribution. ctrl+j/k isn't really vim mode of course, but maybe it's close enough? Also see #909

AlexNabokikh:
@friday I think It is close enough too. Lightweight Vim mode (ctrl-j/k) should suffice for the majority of users. To my knowledge, I haven't come across any launchers that support Vim's distinct modes (normal and insert) and functions. Instead, many popular launchers for Linux, Mac, Windows, and other applications simply incorporate ctrl-j/k/h/l.

friday (approved):

troycurtisjr:
It is a pretty common workaround instead of implementing full modal
editing, however the trouble with this approach is that it conflicts with
the Ctrl-k keybinding when using the emacs keymap in Gnome. Ctrl-j is
technically a conflict as well, but seems highly improbable anyone would be
using it on a single line widget like with Ulauncher. Honestly even Ctrl-k
seems very uncommon for such a single-line mode, but it can still be useful
at times judging from personal use/experience.

Due to the known conflict, I'd highly appreciate this set of keymappings to
be optional...probably safely defaulted to on. Though I know that @friday
isn't super keen on more options 😀️

AlexNabokikh:
@troycurtisjr Can you provide more details about the Emacs keymap in Gnome? Is it a non-standard mode? As a Gnome user, I have encountered several applications that use ctrl-k/j/h/l keys for navigation without any conflicts.

friday:
I'm good with options that makes sense, such as keybindings. Can we make this simplified somehow so it's not a full key for key map settings page?

AlexNabokikh:
@troycurtisjr Even after reading the article, I'm still confused about how to bump in keybinding overlap when the focused window is Ulauncher, but that may be just my lack of experience. 
Nevertheless, I believe that the Ulauncher community will benefit greatly by attracting Vim users. Many keyboard-centric users prefer launchers, and this is especially true for those who belong to the EMAC and VIM communities. I think, it would be fair to support such a large group such Vim users with this change.

troycurtisjr:
Oh I agree and that is a major reason I put in the Ctrl-n/p key bindings 😁

In this case I'm just saying that we need a way to turn it off per user configuration because it is known to conflict with at least one power user oriented prevalent configuration, which as you point out is a demographic that launchers are popular with. 

Ulauncher is a gnome application, and as such unless explicitly configured otherwise by the application developer will automatically get all the text editing capabilities in the text input. When configured with the emacs key map, this means most of the editing shortcuts are also available. I'll note that this key map is also default in all major command line shells. If this is not made optional, then everyone with emacs key map will hit ctrl-k expecting to "kill line", but it will instead be an "up motion". Considering all the other line editing keys work (ctrl-b/a/f, alt-b, alt-backspace, etc) this singular conflict will greatly interrupt this category of power user flow.

This isn't just theoretical, it would immediately negatively impact my own usage. Hence my request for an option to disable this particular key binding.

AlexNabokikh:
Ok. Any tips on how you want to make it optional? GUI settings checkmark, command line argument or any other options?

AlexNabokikh:
@friday @troycurtisjr I added a check for Emacs keybindings, but I'm not sure if this helper function should go somewhere else.
Let me know if you have other ideas on how to do it.

friday:
Looks like this solution could satisfy us all, assuming this works for Troy. Also I'm not sure ctrl+p/n supported is only needed for emacs mode @troycurtisjr? Should we disable that when we enable vim or is it good regardless?

Small nitpick. I would prefer something like this instead of duplicating the rest at the end:

```diff
+        up_aliases = ["p"]
+        down_aliases = ["n"]
+        if not self.is_emacs_bindings_active():
+            up_aliases = ["k", "p"]
+            down_aliases = ["j", "n"]

-            if keyname in ("Up", "ISO_Left_Tab") or (ctrl and keyname == "p"):
+            if keyname in ("Up", "ISO_Left_Tab") or (ctrl and keyname in up_aliases):
                self.results_nav.go_up()
                return True

-            if keyname in ("Down", "Tab") or (ctrl and keyname == "n"):
+            if keyname in ("Down", "Tab") or (ctrl and keyname in down_aliases):
                self.results_nav.go_down()
                return True
```

friday:
Actually, when rethinking this I think the only good solution would be to add a dedicated preferences tab for keyboard shortcuts. The annoying thing about that is the current solution for setting the shortcut for opening Ulauncher uses an ugly convoluted solution to bring up a new GTK prompt (since the settings are otherwise a web app).

Might be time to abandon that solution, but pure html input fields doesn't let you set the modifier.

troycurtisjr:
Yeah I agree that an explicit configuration is much better. I worry this is trying to be too clever and makes too many assumptions about what the user wants. I think we can just include an option in the file to disable these shortcuts. Just a pin of a discussion or similar on updating the setting is likely enough, as I do think it will be a pretty small minority case (even if it does include me). If we notice it being used more, perhaps for some other reason than the Gnome emacs keymap scenario, then we can add it to the UI, or we can add it as part of the more comprehensive keymap configuration @friday mentioned.

AlexNabokikh:
@troycurtisjrI modified the code to only verify the specific parameter in settings.json. I'm unsure of how to incorporate it into the GUI menu or if it's necessary. Perhaps it can be noted in the documentation instead, just like you mentioned.

friday:
That doesn't fully solve the problem imo. The problem is that all of these vim, emacs etc allow customization.

How about changing the setting name to `arrow_key_aliases` (string) and expect it to hold 4 characters (like `"hjkl"`). Then in the code, validate that the string is exactly 4 characters long, and if so, grab the aliases:

```py
arrow_key_aliases_valid = len(self.settings.arrow_key_aliases) == 4
arrow_key_aliases = [*self.settings.arrow_key_aliases] if arrow_key_aliases_valid else [None] * 4
left_alias, down_alias, up_alias, right_alias = arrow_key_aliases
```

Then you can compare directly with those (change `in up_aliases` to `= up_alias` and same for down)

And also add `and keyname` to the condition `if self.results_nav:` to be sure it's not `None` (I'm not sure if it can be None if you just press a modifier or the input is invalid, non-standard keyboard layout or something).

We don't actually use the left/right aliases for anything, but we really should. The Gtk (text) Entry takes care of left/right for us, but it's definitely possible to add aliases for it on top of those. If you think this is something you want to do, feel free, otherwise just leave that part (change the unused variable names to _left_alias, and _right_alias).

Basically this would be just like the like "jump keys" setting, and similar to them you can't set the modifier, just the key. Long term I think we want more advanced settings, but for now I think this is as flexible as we can possibly go.

I already had you revise this solution a lot, so I would recommend to wait for @troycurtisjr to have a say on this too in case he disagrees.

friday:
I don't know what the most logical order would be for the sting, so I just went with the vim "hjkl" order. If gamers want to use "wasd" they would have to put it in like "aswd" for example. In the preferences web view you could duplicate the jump_keys section and change the key and description, but again that's optional (easy for me to add afterwards).

AlexNabokikh:
@troycurtisjr Can you please provide your opinion on that?

troycurtisjr:
Sorry it took so long to get back around to this. 

I think this PR makes a good incremental positive progress. The more complete configuration which @friday described is certainly more capable, but it doesn't seem strictly necessary. If it is easy enough to add in, then it does give the maximum flexibility. Since the default should likely be "hjkl" that seems like a reasonable order to go with. We can always make the GUI more explicit with broken out fields so there is no confusion around the ordering. I also agree the GUI part is not required at this point.

So basically I think either way works, the PR as is, or the more complete option laid out by @friday .

friday:
@AlexNabokikh I pushed some changes to speed it up a little. Does this work for you?

AlexNabokikh:
@friday totally! Looks great!

friday:
I just merged #1256, so if you upgrade now the hotkey support will be using a completely different solution, with different support. If you are using Gnome or if you already set it up with your DE before (as per the recommendations to support Wayland) you should see no change. If you are on KDE or XFCE I added those too but haven't tested myself yet.

If not, then you will have to set it up with your DE settings. You should get a message prompt to help.

friday:
And @troycurtisjr you will have to change/add your settings.json arrow_key_aliases now to some four letter combination like "xnpz"

PR #1256, opened 2023-08-13
feat: replace keybinder with DE-wrappers

friday:
This replaces the keybinder dependency with a custom DE-wrapper to set the hotkey via DE-specific commands. Currently supports Gnome, KDE and XFC, but only the Gnome one has been tested.

The benefit is that this will work on both X11 and Wayland, but it won't work on unsupported DEs until users set it up themselves. There are system notifications to guide you though this though.

Rewrote much of the old code for setting the hotkey with a new better looking and more standardized dialog, and a better UI, better messages.

### First launch with this PR it will try to migrate your setting or add the default, then show this
If your DE is supported it will migrate the setting to the DE shortcut settings, and show this

Screenshot from 2023-08-13 18-43-28
If it's the first time you open Ulauncher, it will default to Ctrl+space (like before)

### If your DE is not unsupported, it will show this instead
Screenshot from 2023-08-13 18-27-32
And as it says, you will open the preferences by clicking it, with more details:
image

### When the DE is supported, the settings look like this
Screenshot from 2023-08-13 16-28-04
Screenshot from 2023-08-13 16-27-57

Screenshot from 2023-08-13 16-28-43

Some of the details might need polish, but that can be changed later.
* For example, when you press "Save" it doesn't close. It's because then you can try it out before you close.
* It also doesn't save when you press enter any more, because you might want enter to be the key in the shortcut (I would recommend against it, but technically you can set it to "Ctrl+Alt+Shift+Enter" or something like that which wouldn't be an existing shortcut).
* I don't know why the modal has so much padding. I guess it probably has an area reserved for the message label which I'm not using at all. I think that's unimportant though.


friday:
I'm going to merge this immediately. I just added you as reviewers @troycurtisjr @gornostal to let you know. This wasn't possible to do before because of the (now fixed) focus issue.

friday:
Did some testing on KDE, and found that this didn't actually work (based on an outdated article with fake date saying June this year).
Also, focus stealing by default is enabled (level 1 - low) and you have to disable it.
Screenshot_20230813_202445

It can be disabled programatically with
`kwriteconfig5 --file ~/.config/kwinrc --group Windows --key FocusStealingPreventionLevel 0`

Also, the empty space has an icon now, so I guess that's what it's supposed to have:
Screenshot_20230813_195735

gornostal:
Great stuff @friday!
And thanks for the very detailed PR description, btw.

friday:
Have tested and fixed the KDE and XFCE implementations now, and they work great (after pushing fixes). So we have great support coverage already. I am pretty sure Budgie, Panteon and Cinnamon uses the same way as Gnome, but with different gsettings paths, so maybe I will add those or some of them. I won't add support for more than those though. Should be easy for contributors who are on smaller desktops to contribute.

PR #1266, opened 2023-08-31
Improved debugging docs

janmechtel:


friday:
You are trying to update a document for v5 which doesn't apply for v6 because the problem has been solved there in a better way.

PR #1279, opened 2023-10-08
Fixed issue with the Ul. window not being hidden when selected app is…

gornostal:
… launched

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ yarn-debug.log*
 yarn-error.log*
 preferences-src/test/unit/coverage
 data/preferences
+.venv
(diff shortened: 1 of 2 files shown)


friday:
Strange that the behavior is different for me. I do recall something about this when testing on Kubuntu though, but I was too focused on fixing the actual KDE issue I was testing.

Also not sure atm why I needed it *not* to close when event is None, but I do think this will break some other case now.

gornostal:
I think the bug appeared only for me because my gnome desktop has a wired
bug itself. It doesn't bring newly opened windows to focus. Hence
ulauncher's on_focus_lost is never triggered.

The reason we need that conditional hiding is bc of the cases like when you
want to select actions from an alternate menu in file explorer by pressing
alt+enter.

PR #1291, opened 2023-10-26
Update ulauncher-toggle script to comply with POSIX sh

nathan818fr:
This pull request updates the ulauncher-toggle script to comply with POSIX sh.

Without this, ulauncher-toggle may not work, since `/bin/sh` only guarantees POSIX sh compatibility and may not support Bash-specific extensions.
For instance on Debian, where `/bin/sh` is linked to **d**ash, I get an error:
```
$ ulauncher-toggle 
/usr/bin/ulauncher-toggle: 6: Syntax error: Bad for loop variable
```

I used [shellcheck](https://www.shellcheck.net/) to validate that the script is now compliant.

### DIFF
--- a/bin/ulauncher-toggle
+++ b/bin/ulauncher-toggle
@@ -3,11 +3,12 @@
 # NOTE: ulauncher-toggle is no longer needed.
 # It's better to use these two gapplication commands directly.
 
-for ((i=1; i<=$#; i++))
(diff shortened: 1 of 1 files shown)


friday:
Thanks! I will test it this weekend.

friday (approved):

Issue #677, opened 2021-01-14
Feature: Disabling / enabling extensions from preferences


Sparky1492:
### Is your feature request related to a problem? Please describe

I thought of the idea while trying to solve an issue with two different extensions.
     _**(If needed I can create another ticket with the issue of the two different extensions in question, please let me know if I need to)_ 
The two plugins in question are [Ip Copy Extension](https://github.com/pwnyprod/ulauncher-ipcopy) and [IP Lookup Extension for ULauncher](https://github.com/munim/ulauncher-ip-lookup), both use the same keyword 'ip' to trigger their extensions.  If I changed extension 1's keyword, then extension 2's keyword would also changed to the same keyword.

### Describe the solution you'd like

This lead to the thought of having the ability to disable or enable plugins to either trace down an issue to a plugin or to keep a plugin that is not needed but wanted to be kept and can be re-enabled at a later date.  Otherwise you would have to keep a list of the extensions and re-pasting them.  If the ability to disable and enable is made available to scripting, then custom scripts could be made to disable or enable as the user would like.

### Additional context

Highlighted
![Feature-Reqeust_2021-01-14 14:27_highlited](https://user-images.githubusercontent.com/945772/104656859-a2637c00-5685-11eb-9814-4b51adb8cfc6.png)

Without Highlighting
![Feature-Reqeust_2021-01-14 14:27](https://user-images.githubusercontent.com/945772/104656916-b7d8a600-5685-11eb-8989-09a902477f42.png)

gornostal:
Yep! It's nice to have.

I'm not a big fan of the checkboxes on the side though.
If anyone is interested in implementing this, just make the extension name grayed out (`opacity: 0.5` or something like that)

freddii:
Just the ability do be able to disable an extension by the drop down menu would be ok for the beginning.

Issue #1257, opened 2023-08-13
Rewrite Ulauncher preferences to use pure GTK (not webkit)


friday:
### Description

Ulauncher's preferences is a separate web app with separate dependencies. It's slower to launch, has a ton of wrapper code to maintain to add a proxy server to communicate with the app, and it's bloating up the package (most of the package size is the preferences app). It also requires a build step in a separate language/environment and with separate packages to maintain, and [the Gtk4 version of the WebKit2Gtk dependency would also require migration work to make it work](https://github.com/Ulauncher/Ulauncher/blob/197e246983ee3779386e9de0bf046c9d24309c15/ulauncher/utils/WebKit2.py#L7).

I suggest rewriting it as a libadwaita/Gtk4 app, separate from the Ulauncher app, so we use Gtk3 and Gtk4, then the rest can be migrated to Gtk4 before we release v6.

I actually started prototyping it and I think it will definitely be less work rewriting it than all the work I already spent trying to improve it.

Alternatively we can use libhandy, which is what libadwaita was called in Gtk3, but I would rather not have to rewrite it to Gtk3, then to Gtk4...

(note that there are other issues in the way for Gtk4, like using X11-exclusive APIs that are now removed by Gtk, but I have a different plan for that).

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing feature requests](https://github.com/Ulauncher/Ulauncher/labels/feature) before suggesting this.

### Feature isn't possible to support as an extension.

- [X] This feature request cannot satisfactory be supported as a third party extension.

### Ulauncher v6

- [X] I have checked and couldn't find my suggestion in v6.

fgclue:
Slower to launch? It's been pretty fast for me.

friday:
> Slower to launch? It's been pretty fast for me.

That's because Ulauncher v5 never kills the preferences window. Just hides and unhides it. You're paying for it in ram instead.

Issue #1063, opened 2022-06-16
Don't run extensions in the background


friday:
### Description

Replaces #285, because it isn't productive or actionable. The majority of Ulauncher extensions don't need to run in the background, and the few ones that do take advantage of it wouldn't need to if we had:
1. An extension API to let extension schedule running itself at certain times later (ex [ulauncher-timer](https://github.com/Ulauncher/ulauncher-timer))
2. An API to call when the application starts (and must be re-run when the preferences are changed), for the few extensions that need this (ex [ulauncher-clipboard](https://github.com/friday/ulauncher-clipboard) to start the clipboard manager or [calculate-anything](https://github.com/tchar/ulauncher-albert-calculate-anything) and to sync currency rates at fixed intervals separate from the extension runtime, possibly also using the scheduling API ^)

Extensions would be started when the user activate them by keyword, and stopped when the Ulauncher window is closed/hidden. Multiple extensions could still run at the same time if the user runs queries in multiple extensions before closing the window. They would not need `PreferencesUpdateEvent`/ `on_preferences_update` any more. That should be handled by the startup API.

Extension preferences now shows errors and runtime status, which it couldn't do when this is implemented, but it could show an extension log, or just error log, or just the last error. And we could still start the extension after it's installed just to see that it *can* run and grab any errors.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

gornostal:
I think running extension in the background or not should be defined by developers. They know better if their extensions need that. 

Off by default.

PR #1292, opened 2023-10-27
Draft: Disable extension

Laurens256:
Hi, I created the functionality to disable extensions without removing them from the list. It works by adding a `is_enabled` key to the extension preferences, which defaults to true if not found.

There were some things I couldn't figure out on my own. in `_migrate_user_prefs` I added the `is_enabled` key, but when migrating with this code, preferences and triggers become empty objects. After manually setting these back, everything does work as expected again, just the migration itself is broken. I couldn't for the life of me figure out why (i tried a lot of stuff to fix this)

Also when running the tests I get the following error which I couldn't find where it was coming from
```
would reformat .../Ulauncher/ulauncher/modes/extensions/ExtensionManifest.py

Oh no! 💥 💔 💥
1 file would be reformatted, 147 files would be left unchanged.
```

Here's a video showing how it works
cinnamon-2023-10-27T104020+0200.webm

Resolves https://github.com/Ulauncher/Ulauncher/issues/677

### DIFF
--- a/ulauncher/modes/extensions/ExtensionManifest.py
+++ b/ulauncher/modes/extensions/ExtensionManifest.py
@@ -194,12 +194,11 @@ class ExtensionManifest(JsonConf):
     def get_is_enabled(self):
         return self.is_enabled
 
-    def toggle_is_enabled(self, enable = None):
+    def toggle_is_enabled(self, enable=None):
         self.is_enabled = enable or not self.is_enabled
 
     @classmethod
     def load_from_extension_id(cls, ext_id: str):
         manifest = super().load(f"{PATHS.EXTENSIONS}/{ext_id}/manifest.json")
         manifest.apply_user_preferences(JsonConf.load(f"{PATHS.EXTENSIONS_CONFIG}/{ext_id}.json"))
-        print(f"{PATHS.EXTENSIONS_CONFIG}/{ext_id}.json")
         return manifest
--- a/ulauncher/modes/extensions/ExtensionManifest.py
+++ b/ulauncher/modes/extensions/ExtensionManifest.py
@@ -50,7 +50,6 @@ class ExtensionManifest(JsonConf):
     input_debounce = 0.05
     triggers: dict[str, Trigger] = {}
     preferences: dict[str, Preference] = {}
-    is_enabled = True
 
     def __setitem__(self, key, value):
         # Rename "required_api_version" back to "api_version"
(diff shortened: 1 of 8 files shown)


ulauncher/modes/extensions/ExtensionManifest.py
```diff
@@ -188,8 +189,17 @@ def apply_user_preferences(self, user_prefs: dict):
...
+        return self.is_enabled
+
+    def toggle_is_enabled(self, enable = None):
+        self.is_enabled = enable or not self.is_enabled
+
     @classmethod
     def load_from_extension_id(cls, ext_id: str):
         manifest = super().load(f"{PATHS.EXTENSIONS}/{ext_id}/manifest.json")
         manifest.apply_user_preferences(JsonConf.load(f"{PATHS.EXTENSIONS_CONFIG}/{ext_id}.json"))
+        print(f"{PATHS.EXTENSIONS_CONFIG}/{ext_id}.json")
```
Laurens256, line 204: Left this in on accident

friday:
would reformat ... is https://github.com/psf/black and you can autoformat with that fast and easy.. I will check the rest tonight.

friday (commented):
ExtensionManifest is for the manifest only. Adding it there basically means that extension devs can specify whether their extensions will be enabled by default when they are installed.

So I reverted and rewrote the Python parts, because it was easier to just do it than trying to guide you though how to.

About the rest I think I would prefer it to change the toggle method to a setter, meaning you pass the boolean. That way the state can never get out of sync.

Also, I plan to rewrite the whole preferences window #1257 and also want to change it so extensions only run on demand #1063, so this feature isn't a prio for me that I would have introduced at this point, since we're gonna have to redo a lot/most of it. But it's also not a huge problem.

Laurens256:
haha sorry, I knew I didn't understand a lot of what was going on :) just wanted to see if I could get it working. Feel free to use / not use whatever you want obviously, hope I didn't cause you too much of a headache ;)

friday:
> haha sorry, I knew I didn't understand a lot of what was going on :) just wanted to see if I could get it working. Feel free to use / not use whatever you want obviously, hope I didn't cause you too much of a headache ;)

No, your contributions are very much appreciated. We don't get a lot of contributions, and if we do it's mostly smaller ones, trying to change a line in some isolated place. I can tell that you are making an effort to keep our standards. Some things are just too hard to understand without having the whole history though.

If you want to contribute more you can always ask us if we want the change, if it's a good time for it and if you need help with finding where to change something :)

I will merge this now. I pushed another change. It says it's failing, but that's actually the base branch.

PR #1293, opened 2023-10-27
feat: rewrite search highlighting so it can use css

Laurens256:
This code is more of a proof of concept of how a box with labels would look instead of using Pango markup for highlighting characters. I personally don't see any way to implement this in a good way, since this method requires hardcoding the `n` number of characters after which results will get ellipsized.

Maybe there's some hacky workaround way to make this work but I don't personally think it's worth the trouble. Feel free to do whatever you want with this pr :) kinda just made this for documentation's sake


friday:
I pushed some changes. To me it works perfect now, but maybe you can test and review what you think?

I tested by copying a desktop entry from /usr/share/applications/ to ~/.local/share/applications/ and changing the name to be very long with lots of nonsense in it.

This is exactly how I want it to behave (ellipsizing in the middle of the non-highlighted chunks):

image

The code that isn't optimal for this is actually python-levenstein. The native python fallback does it better (but it's much slower):

image

Laurens256:
Really seriously impressed with how you wrote this out, very cool :). 

Only found one minor nitpicky problem: when the title is very long and you type the entire title in the search box, the window becomes very wide because highlighted text is set to never ellipsize. I would show you but I still haven't figured out how to take screenshots without the ulauncher dialog dissapearing haha.

Other than that, it works flawlessly as far as I could tell, really nice job :)

friday:
> Really seriously impressed with how you wrote this out, very cool :).

Thank you :+1:

> Only found one minor nitpicky problem: when the title is very long and you type the entire title in the search box, the window becomes very wide because highlighted text is set to never ellipsize. I would show you but I still haven't figured out how to take screenshots without the ulauncher dialog dissapearing haha.

Yeah, I tested that too, but at least for me I had to put in something ridiculously long in, and it didn't end up looking broken.

In Gtk4 you can apparently set overflow hidden, but not in Gtk3. I tried listening to the `"size-allocate"` event and both preventing it and overriding with a new event, but I couldn't get it to work. Maybe with a scrollable/viewport element with set_hadjustment but no scrollbars it will work (like the scroll_to_focus method, but horizontally), but I don't have the time to work on this and don't think it's necessary to fix it. I can't see it happen unless you're intentionally trying to make it happen.

Issue #1288, opened 2023-10-24
Special character parsing error (&)


Laurens256:
### Environment

Mint 21.2 Victoria

### Bug description

Version 6.0.0 beta 1

I encountered a very strange glitch when searching for an application: when typing the letter `f`, my date & time settings item shows up. The & symbol is not parsed correctly and shows up as `&amp;`. ~~I haven't tested this but I suspect it may be similar for other special characters.~~
![20231024_172804](https://github.com/Ulauncher/Ulauncher/assets/85798751/08297def-f087-42e9-aea1-2e1bf91529ff)

The weird thing is, when searching for date and time regularly, the & symbol shows up as expected
![20231024_172816](https://github.com/Ulauncher/Ulauncher/assets/85798751/09b06720-6927-4299-8df2-f3549793f32d)

I'm not even sure date & time should show up at all, since there's no letter f in either the title or description. It doesn't show up for other letters either

(Sorry for pics of monitor btw)

### Log output
```
2023-10-24 17:37:07,201 ℹ️  INFO Ulauncher version 6.0.0-beta1 main.main:63
2023-10-24 17:37:07,201 ℹ️  INFO Extension API version 3.0 main.main:64
2023-10-24 17:37:07,201 ℹ️  INFO GTK+ 3.24.33 main.main:65
2023-10-24 17:37:07,201 ℹ️  INFO PyGObject+ 3.42.1 main.main:66
2023-10-24 17:37:07,201 ℹ️  INFO Desktop: X-CINNAMON (X11) on Linux Mint 21.2 main.main:68
2023-10-24 17:37:07,201 ⚠️  WARNING 

╔═════════════════════════════════════════════════════════════════════════════╗
║                  YOU ARE RUNNING A PRE-RELEASE of ULAUNCHER.                ║
║ Please do not report extension API support warnings to extension developers ║
║ We are still in the process of developing and documenting these features    ║
╚═════════════════════════════════════════════════════════════════════════════╝

 main.main:70
2023-10-24 17:37:07,201 ℹ️  INFO Migrating /home/laurens256/.config/ulauncher/ext_preferences/com.github.laurens256.ulauncher-random-number.json to /home/laurens256/.config/ulauncher/ext_preferences/com.github.laurens256.ulauncher-random-number.json migrate._migrate_file:44
[...]

PR #1294, opened 2023-10-29
Fix html entities

Laurens256:
Very simple fix for unescaping html entities in the search results. Something like this might also be useful for fuzzy searching (and possibly extended to make `and` match `&`). For now though, this simply fixes https://github.com/Ulauncher/Ulauncher/issues/1288


Laurens256:
whoops

Laurens256:
Somehow I added all the commits from the [search highlighting change](https://github.com/Ulauncher/Ulauncher/pull/1293) in this as well lmao. It's a very minor change though, you can just add these lines in some other push if you think the changes are okay

friday:
I didn't have this issue myself. I had the opposite issue where set_markup wouldn't work because they were already unescaped. But it seems safe and good if you got the &amp; showing in the result?

I think it's better that we merge #1293 first though since we're making changes in some of the same places, and we can cherry pick this after.

friday:
cherry-picked as e62850e0a3142cb3b51d7761c1d0371fc97da6bc

PR #1306, opened 2023-11-16
use `sys.path` instead of `os.getenv("PYTHONPATH")`

nazarewk:
When trying to package `v6` for NixOS I stumbled upon [this patch](https://github.com/NixOS/nixpkgs/blob/e44462d6021bfe23dfb24b775cc7c390844f773d/pkgs/applications/misc/ulauncher/fix-extensions.patch) and thought it could be merged upstream.

I don't know the whole story behind using `$PYTHONPATH`, but `sys.path` seems like a cleaner replacement.

### DIFF
--- a/ulauncher/modes/extensions/ExtensionRunner.py
+++ b/ulauncher/modes/extensions/ExtensionRunner.py
@@ -77,7 +77,7 @@ class ExtensionRunner:
             cmd = [sys.executable, f"{extension_path}/main.py"]
             env = {
                 "VERBOSE": str(int(self.verbose)),
-                "PYTHONPATH": ":".join(filter(bool, [PATHS.APPLICATION, os.getenv("PYTHONPATH")])),
+                "PYTHONPATH": ":".join([PATHS.APPLICATION] + sys.path),
                 "EXTENSION_ICON": manifest.icon,
                 "EXTENSION_PATH": extension_path,
                 "EXTENSION_PREFERENCES": json.dumps(backwards_compatible_preferences, separators=(",", ":")),


friday:
Thanks for the work packaging on the NixOS side. Hopefully the changes we've done for v6 doesn't add too much work (part of it should be at least be better since we handle config files better now).

I'm definitely open to addressing this in some way here so you don't need to maintain patches, but I would need to understand why you need this?

I didn't create the original code for this and don't even know why we need PYTHONPATH here (I don't think we do), but when I understand even less is why sending the whole sys.path would be better?

friday:
You can run these two commands in your terminal to demonstrate the difference, and why imo sys.path is not a cleaner replacement:

`PYTHONPATH="/custom/dir" python -c "import os; print(os.getenv('PYTHONPATH'))"`
`PYTHONPATH="/custom/dir" python -c "import sys; print(sys.path)"`

nazarewk:
I think I get what you mean with the last comment. I am not 100% sure what is the reasoning behind this Nix patch so I might dig into it further after I get it up and running.

I think at some point I stumbled upon build wrappers operating on the `sys.path` instead of the PYTHONPATH so it might be Nix-specific behavior.

nazarewk:
Would anyone be opposed to developing a base Nix package/development environment in this repo instead? It would involve a few Nix files and nix-specific patches.

friday:
> Would anyone be opposed to developing a base Nix package/development environment in this repo instead? It would involve a few Nix files and nix-specific patches.

Since the current maintainers don't know how to maintain that code I think it's better if it's not in the repo, but maybe for the future we can set things up to make releases that update the Nix package similar to how we do with the Arch AUR: https://github.com/Ulauncher/Ulauncher/blob/v6/.github/workflows/publish-release.yml#L36. If we need to host files I think in that it would be better with a separate repo though. I want to move all the debian packaging stuff to a separate repo as well.

friday:
I just merged #1314 which should work for you as well.

nazarewk:
> I just merged #1314 which should work for you as well.

yeah, this makes more sense :)

nazarewk:
Actually #1314 does not work in Nix, because specific modules/packages are disjointed from the interpreter/ulauncher package.

probably similar to #1320

friday:
> Actually #1314 does not work in Nix, because specific modules/packages are disjointed from the interpreter/ulauncher package.
> 
> probably similar to #1320

I see :( PATHS.APPLICATION is not the Ulauncher module path?

nazarewk:
> > Actually #1314 does not work in Nix, because specific modules/packages are disjointed from the interpreter/ulauncher package.
> > probably similar to #1320
> 
> I see :( PATHS.APPLICATION is not the Ulauncher module path for you?

No, it's not.

It would probably be true in a dev setup (`python.withPackages`) where it builds some kind of actual environment, but puts too much in `/bin` (like python, pip etc.) that should not be installed system-wide at all.

nazarewk:
PATHS.APPLICATION is for me:
- in a built package:
	```
	/nix/store/3748jl277pxfm4j0qpi18kqczwnbgmy1-python3.11-ulauncher-v6/lib/python3.11/site-packages
	```
	which holds just 2 dirs: `ulauncher` and `ulauncher-6.0.0b4.dist-info`
- in a dev package:
	```
	/nix/store/pic96dp7hw1rhcx2jpglj83r3hmm8ys0-python3-3.11.5-env/lib/python3.11/site-packages
	```
	which holds a complete `site-packages` with all the modules

Issue #1311, opened 2023-11-19
upgrade build system


friday:
### Description

This was among the first thing I started working on when making the v6 branch originally, so I rewrote setup.py to work on newer python versions and using setup.cfg, but since then the community decided to go with PEP-517 and pyproject.toml instead.

So for the most straight forward migration we have to stop calling setup.py directly and also move from setup.cfg to pyproject.toml, but there are alternatives such as Meson and meson-python. For example https://github.com/GNOME/meld (all of Gnome's Python app seems to have the same build system).

https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
https://drivendata.co/blog/python-packaging-2023

This also would be much easier if we did #1257 first as this has a custom build step we could then remove `build_prefs`.

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

### Ulauncher v6

- [X] I have checked and couldn't find my suggestion in v6.

PR #1312, opened 2023-11-19
build: wip build system refactoring draft

friday:
See #1311

Started to work on this, but I hit a point where I felt it was better to switch focus to the preferences rewrite and make a draft PR to look at for the future.

Also I think now that the better way would be to try to avoid customizations in __init__.py that we have to work around in setup.py.

* version: better to load this from a file so we can use the file directly when we use `./ul version` now (and previously the deprecated `./setup.py --version`). setuptools have support for this https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata and https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
* description: doesn't even have to be here (just hard code it in the docs config)
* gi_versions: can be hard coded in __init__.py (it was at one point, but I thought moving it to the project config was nicer
* data_dir: maybe we can avoid relying on this and instead handle it from the ulauncher binary where we already have similar logic to get the PROJECT_ROOT


friday:
I think we can consider this replaced by #1664. It solves the same problems without the toml.

Issue #1184, opened 2023-01-10
docs: clarify compatibility


friday:
### Description

From https://github.com/Ulauncher/Ulauncher/pull/1179#issuecomment-1374910733

This table represent the current state in the v6 branch. All of my tests were with Gnome X11, XWayland and pure Wayland. Tests for KDE and other environments (with all these modes) are needed.

✔️ Supported
❌ Unsupported
🛠️ Unsupported, but can be worked around

|                                     | X11 | XWayland (GDK_BACKEND=x11) | Wayland (GDK_BACKEND=wayland) | Layer Shell (GDK_BACKEND=wayland) |
|-------------------------------------|-----|----------|----------------------------------|-----------------------------------|
| Movable window (1)                      | ✔️   | ✔️        | ❌                                | ❌                                 |
| Positioning (2)                                | ✔️   | ✔️        | ❌                                | ✔️                                 |
| "Always on top" (2)                        | ✔️   | ✔️        | ❌                                | ✔️                                 |
| Display scaling ([1](https://gitlab.freedesktop.org/xorg/xserver/-/issues/1318), [2](https://www.reddit.com/r/swaywm/comments/oih4az/help_with_xwayland_scaling_for_hidpi_monitors/)) (3)                    | ✔️   | 🛠️        | ✔️                                | ✔️                                 |
| Multi-display support (4)              | ✔️   | ❌        |  ✔️                                | ✔️                                 |
| Raise running apps (5)                 | ✔️   | ❌        | ❌                                | ❌                                 |

1. This cannot be supported in Wayland, but fully works in Xwayland
2. This cannot be supported in Wayland, but fully works in Xwayland, in addition to layer shell. Window managers can open new windows centered by default though, which is certainly better than nothing (ex for Gnome/Mutter: `gsettings set org.gnome.mutter center-new-windows true`) 
3. XWayland doesn't support display scaling. You should use font-scaling instead.
4. Ulauncher currently has an option if you want to launch on the "default display" or the active display (the one with the focus/cursor). We should remove this option imo and let the window manager handle it in Wayland (the only way possible), which will use the active display. Then we implement it to work that way for X11 also (see #1195). But for XWayland it is not even possible
5. This is for the "raise-or run" behavior (#979). The API we can use for this is only for X11. On Wayland it technically works, but only for apps running in XWayland (so techincally it works if you force all apps to run in XWayland, but most people really wouldn't want that).

[...]

PR #1318, opened 2023-11-20
do not clear GDK_BACKEND=wayland

nazarewk:
I am not sure what exactly `$GDK_BACKEND` does, but I am setting this to `wayland` in my system config and the code comment indicates it should not clear value of `wayland` itself?

picked from #1307

### DIFF
--- a/ulauncher/utils/launch_detached.py
+++ b/ulauncher/utils/launch_detached.py
@@ -24,7 +24,8 @@ def launch_detached(cmd):
     env = dict(os.environ.items())
     # Make sure GDK apps aren't forced to use x11 on wayland due to ulauncher's need to run
     # under X11 for proper centering.
-    env.pop("GDK_BACKEND", None)
+    if env.get("GDK_BACKEND") != "wayland":
+        env.pop("GDK_BACKEND", None)
 
     try:
         envp = [f"{k}={v}" for k, v in env.items()]


friday:
Thank you :+1: 

I think you are right that it's wrong now, due to a partial revert of #800. I'll look into this later.

friday:
@nazarewk GDK_BACKEND is an environment variable for specifying if you want to use pure Wayland or XWayland. The default if you leave it empty *should be* pure Wayland, so I think this PR doesn't effectively do anything, and since you didn't observe any differences I think this is the case, but maybe you still want/need this change?

@troycurtisjr I tagged you as a reviewer because you know this better than me. @nazarewk is here as a Nix packager, so this is why he's asking to change code like this.

For context, #800 was partially reverted because defaulting to Xwayland became more controversial than not doing it https://github.com/Ulauncher/Ulauncher/issues/1184

nazarewk:
Yeah, it runs fine on Wayland even without this variable (confirmed by xeyes not tracking the ulauncher window). Stumbled upon it randomly and changed only because of the comment above the line of code.

troycurtisjr:
I wonder if the more appropriate thing here is to update the comment, unless there is specific behavior fixed with this change? Your comment seems to indicate it didn't have a functional change. It seems to me that having the launcher effectively set this particular environment variable either way is likely not desirable in general.

That said, this approach *seems* safe enough since:

1. It can only ever be reasonably set to `wayland` when running under `wayland`
2. When running under `wayland`, `wayland` is generally the assumed default anyway, so setting it explicitly presumably has no impact
3. If set to `x11` under `wayland` to make Ulauncher behave in a particular way, it definitely doesn't seem reasonable to force apps launched by Ulauncher to use `x11`. 

This PR seems like it doesn't negatively impact anything, and certainly makes the intent more clear, so it seems OK to me.

troycurtisjr (approved):

friday (approved):
I'll merge this until we figure out how we want it to behave.

Thanks again :+1:

PR #1326, opened 2023-11-28
nix: fix `ruff` check on PATHS

nazarewk:
`ruff check .` is not happy about the missing usage of `PATHS` after the patch

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -126,6 +126,7 @@ let
         --replace gapplication ${glib}/bin/gapplication
       substituteInPlace \
           ulauncher/modes/extensions/ExtensionRunner.py \
+        --replace 'from ulauncher.config import PATHS,' 'from ulauncher.config import' \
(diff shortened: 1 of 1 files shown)


friday (approved):
Oh, I guess that default variable I removed wasn't 100% redundant after all. Sorry! :D

I guess the better way would be to get rid of both patches. I don't recall why the current unpatched behavior didn't work for nix so if you have a suggestion to change that again feel free to PR or comment here.

nazarewk:
> I don't recall why the current unpatched behavior didn't work for nix so if you have a suggestion to change that again feel free to PR or comment here.

I'm pretty sure it's because the package itself doesn't run in an environment, but I can revise this behavior when I have some free time.

nazarewk:
probably this is the culprit: https://github.com/NixOS/nixpkgs/blob/5a09cb4b393d58f9ed0d9ca1555016a8543c2ac8/pkgs/development/interpreters/python/sitecustomize.py#L20-L23
processing dependencies from `$NIX_PYTHONPATH` straight into `site.addsitedir` (and probably `sys.path` finally)

Issue #1018, opened 2022-04-25
Extensions CMD Installation


enmanuelmoreira:
### Description

There is a way to installing extensions via CMD?

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

friday:
I would like that too. If I wrote the extension integration from scratch it's where I would start, before even considering making a GUI for it.

It also ties into being able to debug your extension #818 as well as #925. I would love it if you could install or update extensions just by typing `ulauncher extension add [url/path]` for example.

However, it's not a high priority issue as it's kind of complicated to mess with, and you can already do it with "git clone" to the right path.

For example:

```bash
git clone --depth=1 git@github.com:Ulauncher/ulauncher-emoji.git ~/.local/share/ulauncher/extensions/com.github.Ulauncher.ulauncher-emoji
```

(After that you would have to restart Ulauncher).

The difference is this would create a .git folder, with the remote branch and similar. But if you want to use the CLI instead of the UI, this is probably good anyway. If you remove `--depth=1` you get a full remote and can update extensions by using `git pull`.

PR #1330, opened 2024-01-02
refactor: major extension classes refactoring

### LINKED ISSUE #1044: Clarify extension mode classes architecture
### Description

We have many classes for this and their responsibilities bleed into each other and make the workflow complicated to change or write tests for.
* Server, Controller and Runner (I think Runner should be handled by controller, but in fact the Controller is created from a callback in Runner)
* Downloader and Remote (Downloader is just a thin layer over Remote at this point, so definitely candidate for merging Downloader into remote and renaming it appropriately)
* ~~Manifest and Preferences~~ (these are now merged)

Manifest and Preferences are basically data classes, but they store external properties to handle things like getting the full path to the icon and loading the preferences and falling back to the defaults in manifest. I think these *should* be modelling like pydantic data classes (optimal for cases like this) or at least something similar, and the methods that needs external properties should be moved to a wrapper/container class (not sure which one though).

Remote and Downloader are kind of the same class but different abstractions. This doesn't bother me much. ~~What's worse is that Downloader imports Runner and handles starting extensions after installing, as well as stopping before uninstalling, and somehow Uninstalling is done by the "Downloader" class, which makes it more of an extension "Manager"~~. Update: I pushed some improvements, so this is no longer the case.

"Controller" is also a manager that imports the manifest and Preferences, but it is initialized only from a callback from the extension event triggered by the runner. So these instances don't exist and can't be used unless the extension is running. Furthermore, we support starting extensions from outside of Ulauncher's runtime (for debugging), and that means they are never started through the `Runner`. This is almost certainly the reason for the current architecture which separates these classes.

I'm more into [this philosophy](https://www.youtube.com/watch?v=wfMtDGfHWpA) of coding, where objects is basically data and maybe some get dynamic props, but never things like writing business logic to handle a parent or sibling class in a data class. Python isn't a functional language, so we can't apply this straight to Ulauncher, but at least we should try to make improvements in this direction.

### Communication guidelines

friday:
Fixes #1044
This PR should also help for #1018 (need a common API for the UI and cmd installation) and #1063.

Creates a new `ExtensionController` class (not to be confused with the old `ExtensionController` class which I just renamed and pushed before this PR (e3af72a1)

`ExtensionController` is intended to be the single interface to be called by other classes, and to internally handle ExtensionManifest, ExtensionRemote, ExtensionRuntime and ExtensionDb/Record (now renamed to `ExtensionState`) so other classes won't touch those classes won't handle them, and they won't handle each other (with a couple of necessary exceptions for ExtensionManifest).

* `ExtensionDownloader` no longer exists (it is now part of ExtensionController)
* ExtensionRunner has been rewritten to "ExtensionRuntime". ExtensionRuntime handle one single instance and is really just a thin abstraction over Gio.Subprocess with Ulauncher extension specific stderr error handling. `ExtensionController.start()` handles the manifest validation, updating the state and preparing the ENV variables.
* `ExtensionState` (the new ExtensionDb/Record) is integrated into ExtensionController and only accessible from there. It now creates individual json files for each extension (like extension settings) instead of storing all to `extensions.json`

I also opted to rewrite the ExtensionRuntime to use asyncio.subprocess instead of Gio.Subprocess. I think that would be better for the architecture and maintenance, but I opted not to do it because we had some issues before Troy rewrote to use Gio.Subprocess.


friday:
@nazarewk 
Happy new year!

I made some much needed refactoring work on the extension classes, and I'm not sure how to handle PRs like this since I don't have the means to verify that I didn't break the nix package. Do you think it would be possible to add something to our [GitHub linting CI workflow](https://github.com/Ulauncher/Ulauncher/blob/v6/.github/workflows/tests.yml) that could continuously verify that the package can still be installed?

nazarewk:
Hello, I had a lot going on personally and didn't have a chance to move the topic forward.

While updating my system I just noticed the `mypy` failing and pinned the version to latest working commit as fixing it required some research on generating proper stubs.

friday:
> Hello, I had a lot going on personally and didn't have a chance to move the topic forward.
> 
> While updating my system I just noticed the `mypy` failing and pinned the version to latest working commit as fixing it required some research on generating proper stubs.

Ok. I can definitely relate to having to fight the linter tools. There's no hurry here. I will move on with PR in the meanwhile and you can make a new PR when you have time.

And I will rename `.extension-record.json` to `.default-state.json`.

nazarewk:
yeah no problem, I'm not using it yet anywhere anyway

nazarewk:
I wonder what is the Dockerfile for `ulauncher/build-image:6.0`?

friday:
> I wonder what is the Dockerfile for `ulauncher/build-image:6.0`?

This one: https://github.com/Ulauncher/Ulauncher/blob/v6/Dockerfile
I remember I had to revert/rebuild to fix some release (it's also used in the CI), so the published version might not have been the same as the current v6 branch.
If it doesn't work locally, maybe try building with `docker build -t ulauncher/build-image:6.0` and that will be used by the `./ul dev-container` script.

nazarewk:
I figured out the error, can you include this patch?

```patch
Subject: [PATCH] nix: fix build
---
Index: nix/default.nix
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nix/default.nix b/nix/default.nix
--- a/nix/default.nix	(revision 07de958e41a732dc5f11e9a79fe3e20f00bc02d9)
+++ b/nix/default.nix	(revision 4265f04fbc6f933a0d8bbdc308c98242a540a167)
@@ -62,6 +62,7 @@
   packages.tests.python = pp: (with pp; [
     black
     mock
+    (pygobject-stubs.overridePythonAttrs (old: { PYGOBJECT_STUB_CONFIG = "Gtk3,Gdk3,Soup2"; }))
     pytest
     pytest-mock
   ]);
@@ -126,7 +127,6 @@
         --replace gapplication ${glib}/bin/gapplication
       substituteInPlace \
           ulauncher/modes/extensions/ExtensionController.py \
-        --replace 'from ulauncher.config import PATHS,' 'from ulauncher.config import' \
         --replace '"PYTHONPATH": PATHS.APPLICATION,' '"PYTHONPATH": ":".join(sys.path),'
 
       substituteInPlace \
```

nix/default.nix
```diff
@@ -125,7 +125,7 @@ let
           io.ulauncher.Ulauncher.desktop \
         --replace gapplication ${glib}/bin/gapplication
       substituteInPlace \
-          ulauncher/modes/extensions/ExtensionRunner.py \
+          ulauncher/modes/extensions/ExtensionController.py \
         --replace 'from ulauncher.config import PATHS,' 'from ulauncher.config import' \
```
nazarewk, line 129:
```suggestion
```

nazarewk: also this:

```patch
Subject: [PATCH] nix: fix build
---
Index: nix/default.nix
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nix/default.nix b/nix/default.nix
--- a/nix/default.nix	(revision 07de958e41a732dc5f11e9a79fe3e20f00bc02d9)
+++ b/nix/default.nix	(revision 4265f04fbc6f933a0d8bbdc308c98242a540a167)
@@ -62,6 +62,7 @@
   packages.tests.python = pp: (with pp; [
     black
     mock
+    (pygobject-stubs.overridePythonAttrs (old: { PYGOBJECT_STUB_CONFIG = "Gtk3,Gdk3,Soup2"; }))
     pytest
     pytest-mock
   ]);
```

nazarewk (changes requested):
Changes joined together, but I'm not sure if that Github action would work:
```patch
Subject: [PATCH] nix: fix build
---
Index: nix/default.nix
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nix/default.nix b/nix/default.nix
--- a/nix/default.nix	(revision 07de958e41a732dc5f11e9a79fe3e20f00bc02d9)
+++ b/nix/default.nix	(date 1704284789757)
@@ -62,6 +62,7 @@
   packages.tests.python = pp: (with pp; [
     black
     mock
+    (pygobject-stubs.overridePythonAttrs (old: { PYGOBJECT_STUB_CONFIG = "Gtk3,Gdk3,Soup2"; }))
     pytest
     pytest-mock
   ]);
@@ -126,7 +127,6 @@
         --replace gapplication ${glib}/bin/gapplication
       substituteInPlace \
           ulauncher/modes/extensions/ExtensionController.py \
-        --replace 'from ulauncher.config import PATHS,' 'from ulauncher.config import' \
         --replace '"PYTHONPATH": PATHS.APPLICATION,' '"PYTHONPATH": ":".join(sys.path),'
 
       substituteInPlace \
Index: .github/workflows/tests.yml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.github/workflows/tests.yml b/.github/workflows/tests.yml
--- a/.github/workflows/tests.yml	(revision 07de958e41a732dc5f11e9a79fe3e20f00bc02d9)
+++ b/.github/workflows/tests.yml	(date 1704285533613)
@@ -20,6 +20,9 @@
         run: DISPLAY=:1 xvfb-run pytest tests
       - name: build preferences
         run: ./setup.py build_prefs --force
+      - uses: cachix/install-nix-action@v22
+      - name: nix build
+        run: nix build --out-link nix/result --show-trace --print-build-logs .#default
       - name: build docs
         run: |
           cd docs
```

friday:
Thank you! I split it up in two commits so I could test the build would fail, then work.

But it seems it's failing now because the docker image doesn't have curl. I'm building a new one now trying to fix that.

nazarewk:
Yeah, I'm not familiar with GitHub Actions at all, I guess those 3 lines would work in whatever default configuration is provisioned by GHA.

friday:
I moved the `(pygobject-stubs.overridePythonAttrs (old: { PYGOBJECT_STUB_CONFIG = "Gtk3,Gdk3,Soup2"; }))` part to a sepatate commit to v6, amended the new install-nix-action workflow (for now) and squashed the other fix to the commit that caused it.

So it should work now, minus the new workflow. Will try that in a separate branch.

PR #1332, opened 2024-01-04
ci: add test to validate nix package can be built

friday:
Cherry-picked from https://github.com/Ulauncher/Ulauncher/pull/1330#pullrequestreview-1801933060 (credit to nazarewk). It wasn't working in that PR, but that was due to issues with the docker image which have been fixed since.

I'm keeping this as a separate PR for now since this runner is a bit slow (understandably) compared to the others and I'm not sure I want this to run for every push (because we have limited free ci minutes).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,9 @@ jobs:
         run: DISPLAY=:1 xvfb-run pytest tests
       - name: build preferences
         run: ./setup.py build_prefs --force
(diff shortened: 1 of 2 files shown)


friday:
Actually I'll just merge it and we can revert if we get too close to using our git action limits.

@nazarewk just to let you know it works now (no changes to your code, only needed to fix some Docker image issues).

Thank you again! :pray:

PR #1358, opened 2024-02-24
fix: restore setsid if not running ulauncher through systemd

friday:
The switch to GLib.spawn_async reintroduced #754 because `child_setup=os.setsid` doesn't work. 

Reproduction: Stop the ulauncher systemd service and run `make run` or `ulauncher -v` in a terminal. Launch an app which isn't already running or part of the OS (for example Chrome, Firefox but not for example Nautilus). Switch back to the terminal and press ctrl+c to kill Ulauncher.

With this PR it should not kill any apps when Ulauncher is killed, but before it did.

I have tested and we specifically need both setsid and nohup  if Ulauncher is not running as a systemd service. Running systemd instead of this does not solve the problem because that process will still be killed.

`standard_output=True, standard_error=True` is needed to prevent nohup from creating `nohup.out` files

### DIFF
--- a/ulauncher/utils/launch_detached.py
+++ b/ulauncher/utils/launch_detached.py
@@ -1,22 +1,11 @@
 import logging
 import os
-from shutil import which
 
 from gi.repository import GLib
 
 from ulauncher.utils.systemd_controller import SystemdController
 
 logger = logging.getLogger()
-use_systemd_run = which("systemd-run") and os.system("systemd-run --user --scope true  2> /dev/null") == 0
-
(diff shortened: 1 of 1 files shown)


friday:
@troycurtisjr That looks awesome. I have tested and approve. Will you open up a PR with that branch (you can merge it directly as well since I have already tested).

It's good to know this only happened when running from a terminal.

Also, I realized that first part about the systemd condition as I was testing, and I realized later that we did do this for v5 #853, so this part was not affecting our stable releases.

PR #1360, opened 2024-02-25
fix: Don't setup unmonitored pipes to child processes.

troycurtisjr:
This replaces the call out to the nohup and setsid executables with an in code implementation of re-opening the standard file descriptors to /dev/null. This avoids creating the pipes that has no reader and thus could block after enough output from the child process.


friday (approved):
:heart:

PR #1363, opened 2024-03-03
refactor: replace wnck with ewmh

### LINKED ISSUE #1315: enhancement: try to replace Wnck
### Description

Wnck is only used for raising windows now (see #979), but Wnck is kind of big just for this, and it should be possible to do this without Wnck (which depends on Gtk3, so it makes upgrading to Gtk4 harder).

[python-xlib](https://pypi.org/project/python-xlib/) has a wrapper around [libX11](https://www.mankier.com/package/libx11-devel).[XRaiseWindow()](https://www.mankier.com/3/XRaiseWindow), which works for a lot of cases, but not reliably (resulting in focus issues for some apps). We want to use [_NET_ACTIVE_WINDOW](https://stackoverflow.com/a/2767663/633921) instead.

I'm unsure how wmctrl is implemented, but it's just a binary and not a library, and it felt clunky and unreliable when when my initial draft of #979 used wmctrl.

This small library implemented on type of python-xlib seems to do it really well: https://github.com/parkouss/pyewmh
It looks great and possibly usable for other window hints like `_NET_WM_ACTION_ABOVE` instead of  "set_keep_above".

It uses python-xlib as the way to communicate with the x11 server. I'm not sure if we can get around this and use libX11 directly instead (via ctypes) like we do in xinit.py (probably not).

If the solution we go with is good we could possibly also use it to handle GdkX11.X11Display and other parts of the GTK SDK that was removed in Gtk4.

More details about the "set_keep_above" and moving/centering details: https://discourse.gnome.org/t/gtk4-set-window-always-on-top-and-on-center-of-current-monitor/9068

### Communication guidelines

### Ulauncher v6

friday:
Closes #1315

I brought in wnck mainly for this feature, but I had hoped we could also use it to handle other X11 stuff that was removed from Gtk4. This turned out not to be the case. Wnck seems to be rather limited in functionality, while also strictly depending on Gtk3, making it worse.

Instead this replaces it with python-ewmh, a small wrapper library around python-xlib to support "Extended Window Manager Hints" (hence the name).

I also tried to rewrite the `get_monitor` method so that it used python-xlib, and I managed to get the mouse coordinates:
```py
from Xlib.display import Display
data = Display().screen().root.query_pointer()
# data.root_x and data.root_y 
```

However:
1. I haven't found a replacement for [`display.get_monitor_at_point(x, y)`](https://lazka.github.io/pgi-docs/Gdk-3.0/classes/Display.html#Gdk.Display.get_monitor_at_point). It's possible that we can use `data.child` (window) to get the actual "Window" (or virtual window if there is no window under your pointer?) but that doesn't help. We need to calculate this: https://www.reddit.com/r/commandline/comments/nko6rw/how_can_i_get_the_current_monitor_on_which_a/ (this is also what get_monitor_at_point does in the source code).
2. The coordinates we get differ from the coordinates we get from Gtk. The Gtk ones includes the scaling factor (so if your scaling factor is 2x, then the Gtk ones will be half of the Xlib ones. This might not be a problem though (since the replacement for we write for `get_monitor_at_point` likely won't either).

After replacing this also, we will no longer directly use GdkX11, which is a good initial step towards Gtk4. Then we should also handle positioning/moving the window with Xlib, and `set_keep_above` with `_NET_WM_ACTION_ABOVE`. But maybe not in this PR.

Arch, Debian and Nix package this library. But I don't think Fedora does, and smaller distros surely won't. Maybe we should vendor the library instead. It's just [500 lines of code](https://github.com/parkouss/pyewmh/blob/master/ewmh/ewmh.py). I'm not sure how much else of this library we need, but since it's small and we have at least two use cases (_NET_WM_ACTION_ABOVE) I think fully including it in that case makes sense (and of course making the necessary attributions to the author and link to the repo in the header).


troycurtisjr (approved):
The reasoning looks good to me, and the approach seems sound. I tested it on my machine and things seems to work as expected from what I can tell (though I wasn't overly familiar with the particulars of the wnck functionality.

friday:
@nazarewk @oklopfer tagging you too for awareness you are packaging Ulauncher v6. Nix should be handled by this PR though.

PR #1408, opened 2024-08-16
Update build-rpm.sh to include ulauncher.service

philnewm:
# Bugfix for rpm-build.sh

## Issue

In Ulauncher v5 the rpm build process currently returns an error.
`error: can't copy 'ulauncher.service': doesn't exist or not a regular file`
This was already mentioned in issue #1156 for Fedora 36.
Since I encountered the same issue when running `./ul build-rpm 5.15.7 alma9.4`, I investigated.

## Cause

This is due to the build script at `scripts/build-rpm.sh` missing the line for ulauncher.service in its rsync command.

## Testing
I only tested this for Almalinux 9.4 but it should make no difference when it comes to Fedora.

### DIFF
--- a/scripts/build-rpm.sh
+++ b/scripts/build-rpm.sh
@@ -49,6 +49,7 @@ build-rpm () {
         setup.py \
         ulauncher \
         ulauncher.desktop.in \
+        ulauncher.service \
         $tmpdir \
         --exclude-from=.gitignore
 


friday (approved):

friday:
Thank you.

v5 is about to become unsupported. I have had no time to work on Ulauncher lately, but the focus / plan the last 3 years has been on Ulauncher 6, and now the prio is to finish up the last parts and release it so we don't have to support v5.

We don't use that script for making releases any more and it was hard to maintain. When no one uses / tests scripts and you have to make changes around the build process etc, then they start breaking. For this reason I removed the script in the new branch completely, but I'm certainly not against reintroducing it in some form, adapted to the new scripts (in the makefile).

philnewm:
I'm excited for Ulauncher v6 already and totally understand your take on that.
Since I integarted Ulauncher in my automated setup processes I most likely also switch to Ulauncher v6 at some point and try to figure out the rpm build process again :D

PR #1430, opened 2024-12-20
EXPERIMENTAL: module for automatic installation of extensions 

iaacornus:
… dependencies

### DIFF
--- a/ulauncher/modes/extensions/extension_environment.py
+++ b/ulauncher/modes/extensions/extension_environment.py
@@ -8,17 +8,51 @@ from pathlib import Path
 from os.path import exists
 
 
-def check_requirements(ext_dir: Path) -> bool:
-    """check_requirements
+class ExtensionEnvironment:
+    def __init__(self, ext_dir: Path, venv_map: dict[str, set]) -> None:
+        self.ext_dir: Path = ext_dir
+        self.venv_map: dict[str, set[str]] = {
+            "test": set(
+                    [
+                        "hello==1.0"
+                    ]
+                )
+        }
 
-    Args:
-        ext_dir (Path): Path of the extension dir downloaded/cloned.
+    def check_requirements(self) -> bool:
+        """check_requirements
 
-    Returns:
-        bool: whether a requirements.txt file exists for the extension.
-    """
+        Args:
+            ext_dir (Path): Path of the extension dir downloaded/cloned.
 
-    if exists(ext_dir / "requirements.txt"):
-        return True # it should be requirements.txt only, not requirement.txt
-    return False
+        Returns:
+            bool: whether a requirements.txt file exists for the extension.
+        """
+
+        if exists(self.ext_dir / "requirements.txt"):
+            return True # it should be requirements.txt only, not requirement.txt
+        return False
+
+    def prep_extension(self):
+        req_path: Path = self.ext_dir / "requirements.txt"
+
+        if not self.check_requirements():
+            self.create_sole_venv()
+
+        with open(req_path, "r", encoding="utf-8") as req_file:
+            dependencies: set[str] = set(req_file.readlines())
+
+        for venv in self.venv_map.keys():
+            if dependencies >= self.venv_map.get(venv): # type: ignore
+                self.create_sole_venv()
+            else:
+                continue
+
+            self.install_ext(venv)
+
+    def create_sole_venv(self):
+        pass
+
+    def install_ext(self):
+        pass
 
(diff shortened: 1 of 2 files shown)


NoPlagiarism:
You can convert PR to draft using button in right sidebar

[GitHub docs article](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request)

iaacornus:
@friday , should I take care of blocks like this?

```
        except Exception as e:
            msg = f"Invalid URL: {url}"
            raise InvalidExtensionRecoverableError(msg) from e
```

friday:
> @friday , should I take care of blocks like this?

Error handling doesn't have to be part of the first draft, but it would be great to cover that for the PR when it's ready to be merged. You can make `TODO` comments if you are not sure how to handle it.

ulauncher/modes/extensions/extension_environment.py
```diff
@@ -0,0 +1,58 @@
...
+
+        if exists(self.ext_dir / "requirements.txt"):
+            return True # it should be requirements.txt only, not requirement.txt
+        return False
+
+    def prep_extension(self):
+        req_path: Path = self.ext_dir / "requirements.txt"
+
+        if not self.check_requirements():
+            self.create_sole_venv()
```
troycurtisjr, line 40: If there is no requirements.txt then there probably shouldn't be a venv.

ulauncher/modes/extensions/extension_environment.py
```diff
@@ -0,0 +1,58 @@
...
+
+        with open(req_path, "r", encoding="utf-8") as req_file:
+            dependencies: set[str] = set(req_file.readlines())
```
troycurtisjr, line 43: We should definitely not attempt to parse dependencies ourselves, this should be offloaded entirely to `pip`

troycurtisjr:
Moving general discussions on this idea over from issue #273 .

I agree with @friday that each extension which has a requirements.txt should get their own venv, even if they happen to have the same requirements as another extension. It will get too messy to try to manage as they change independently, and it will likely be extremely rare that two extensions really do have the same requirements, which means there will be little benefit realized for the complexity.

My comment on using the system packages when possible allows for more reuse, without the entire requirements being the same, but most importantly it allows for the extension to access the actual system ulauncher modules, which really need to be in sync with the parent ulauncher process. Perhaps there are other methods to accomplish that, like modifying the system path to point to the proper ulauncher modules or similar, but whatever the implementation the extensions need to see the right ulauncher modules to properly implement the client side of the connection.

ulauncher/modes/extensions/extension_environment.py
iaacornus: well, i think a general/isolatd venv for an extension with no requirements.txt will be better than none at all, so even if it messes up, it can be contained

ulauncher/modes/extensions/extension_environment.py
iaacornus: the reasoning behind that is that it will be stored into a json.txt, and it will be later read when another extension is needed to be installed. if the requirements.txt of the new extension is a subset or is similar to the another one, then we can just install that similar extension to the same venv

iaacornus:
> Moving general discussions on this idea over from issue #273 .
> 
> I agree with @friday that each extension which has a requirements.txt should get their own venv, even if they happen to have the same requirements as another extension. It will get too messy to try to manage as they change independently, and it will likely be extremely rare that two extensions really do have the same requirements, which means there will be little benefit realized for the complexity.
> 
> My comment on using the system packages when possible allows for more reuse, without the entire requirements being the same, but most importantly it allows for the extension to access the actual system ulauncher modules, which really need to be in sync with the parent ulauncher process. Perhaps there are other methods to accomplish that, like modifying the system path to point to the proper ulauncher modules or similar, but whatever the implementation the extensions need to see the right ulauncher modules to properly implement the client side of the connection.

on that consideration, then i think that will be an easier approach, i will push some commits incorporating these ideas

ulauncher/modes/extensions/extension_environment.py
troycurtisjr: Note that a Python virtual environment has no meaningful isolation beyond the process isolation that is already present. Despite the name, it doesn't imply anything akin to a java virtual machine or similar. It really is just a directory of modules, and the PYTHONPATH is modified to point to those modules in preference to the system modules. Generally the Python executable in the `bin` path of the venv is just a wrapper script which sets up the path information and then exec the actual system Python executable.

PR #1433, opened 2024-12-30
Remove keyword from description

willuhmjs:
There is a bug that skips over the logic that removes keyword from the description when using default search. This PR adds that logic. I also took the liberty to the typos in CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,15 +4,15 @@ Please read our General communication guidelines [here](CODE_OF_CONDUCT.md#Gener
 
 ## Code contributions
 
-Thank you for you interest in contributing to Ulauncher! We very much appreciate it.
+Thank you for your interest in contributing to Ulauncher! We very much appreciate it.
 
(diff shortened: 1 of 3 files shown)


friday:
I haven't had the time to look at this yet, but thanks for the much needed typo fixes. I cherry-picked that commit directly to the v6 branch.

willuhmjs:
Thanks! This PR addresses an issue where the keyword is still included in the description. 

image

friday:
Thanks for finding the issue and where to fix it :pray: 

There was a less complex solution I went for instead. The core semantics are really confusing/misleading here, so I understand how it wasn't easy to fix.

`query` is basically the whole user input, and the `query.keyword` is always the first word (see the Query class), even if there wasn't any actually matched keyword.

So the logic where we check `query.keyword == self.keyword` it means "the first word of the user query matches the shortcut keyword".

So what I did was clean up the conditions a bit in 607b6e8e, and with that change I only had to move the condition down to make sure the `is_default_search` condition would only apply if the keyword didn't match.

Sorry for side-stepping your PR this way, but it was easier to code than to feedback and get you to do it. And I put you as the co-author of the commit since you were :slightly_smiling_face:

willuhmjs:
Yeah, I tried fixing the issue a few different ways and wasn't sure what the best way was. Glad you were able to find a more straightforward solution! I look forward to contributing to this project more in the future.

friday:
> I look forward to contributing to this project more in the future.

Much appreciated. This PR was very helpful :+1: 

> Glad you were able to find a more straightforward solution!

Considering I couldn't spell straightforward, that's not too bad :smile:

PR #1443, opened 2025-01-31
fix: fix nix flake

friday:
I don't know anything about this, but I got notified by that this was needed to fix the flake :D

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -6,7 +6,7 @@
 , gdk-pixbuf
 , git
 , glib
-, gnome
+, adwaita-icon-theme
 , gnumake
 , gnused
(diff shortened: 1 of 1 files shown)


friday:
I think the tests are only failing because it's a branch from another repo. Will ignore that for now and merge

omarcresp:
Hello just saw this. Thanks for merging it! I quick solved it for my machine but didn’t open bc life happens and then I forgot sorry, the nix build seems to fail in the repo probably bc it is locked to be using the older version of nix where gnome group exist. I can look into it

Vortriz:
yupp, the lockfile was updated 1 year back. updating it would fix that ig.

PR #1445, opened 2025-02-06
Update flake 05.02.2025

omarcresp:
Fix for the build that was missing the flake update


omarcresp:
it should be fixed now. The error is giving currently looks like its caused bc the pipeline cant run from a fork.
build

friday:
Looks good. Thank you!

PR #1461, opened 2025-05-01
feature: #273 automatically install extensions

gornostal:
#273 
Implements automatic dependency installation for extensions.

Uses `pip install ... --target /path` to download dependencies. Then, it passes the deps dir in `PYTHONPATH` when running extensions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 4
+indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/docs/extensions/tutorial.rst
+++ b/docs/extensions/tutorial.rst
@@ -218,7 +218,7 @@ ExtensionCustomAction
 
 ActionList
   List of actions
-  
+
 
 Custom Action on Item Enter
 ---------------------------
@@ -265,6 +265,21 @@ Custom Action on Item Enter
   Now this will be rendered when you click on any item
 
 
+Dependencies and requirements.txt file
+--------------------------------------
+
+If your extension requires any python dependencies, you can specify them in a ``requirements.txt`` file.
+Ulauncher will install them automatically when the extension is installed.
+
+Avoid specifying development dependencies such as mypy, pytest, etc. in the ``requirements.txt`` file — users don't need them.
+Instead, you can create a separate ``requirements-dev.txt`` file for your development dependencies.
+
+It's a good idea to specify the version of the dependencies you are using to avoid breaking changes in the future.
+For example, if your extension requires the `requests` library, you can create a `requirements.txt` file with the following content::
+
+  requests==2.32.3
+
+Refer to this `requirements.txt documentation <https://pip.pypa.io/en/stable/reference/requirements-file-format/>`_ for more information.
 
 .. NOTE::
   Please take `a short survey <https://goo.gl/forms/wcIRCTjQXnO0M8Lw2>`_ to help us build greater API and documentation
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -119,7 +119,7 @@ let
 
       substituteInPlace \
           ulauncher/modes/extensions/extension_controller.py \
-        --replace-fail '"PYTHONPATH": paths.APPLICATION,' '"PYTHONPATH": ":".join(sys.path),'
+        --replace-fail 'paths.APPLICATION,' '":".join(sys.path),'
 
       substituteInPlace \
           ulauncher.service \
--- a/ulauncher/modes/extensions/extension_dependencies.py
+++ b/ulauncher/modes/extensions/extension_dependencies.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+
+import logging
+import subprocess
+import sys
+from os.path import isdir, isfile
+
+logger = logging.getLogger()
+
+
+class ExtensionDependencies:
+    """
+    Manages python dependencies of the extension.
+    """
+
+    ext_id: str
+    path: str
+
+    deps_out_subdir: str = ".dependencies"
+
+    def __init__(self, ext_id: str, path: str):
+        self.ext_id = ext_id
+        self.path = path
+
+    def get_dependencies_path(self) -> str | None:
+        """
+        Get the path to the dependencies directory.
+        Returns None if there's no dependencies for the extension.
+        """
+        requirements_file = f"{self.path}/requirements.txt"
+        deps_path = f"{self.path}/{self.deps_out_subdir}"
+        if not isfile(requirements_file) or not isdir(deps_path):
+            return None
+
+        return deps_path
+
+    def install(self) -> None:
+        """
+        Install the dependencies for the extension.
+        Runs `pip install -r extension-X/requirements.txt --target extension-X/deps`
+        """
+        requirements = self._read_requirements()
+
+        if not requirements:
+            return
+
+        command = [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "-r",
+            f"{self.path}/requirements.txt",
+            "--target",
+            f"{self.path}/{self.deps_out_subdir}",
+        ]
+        logger.info("Installing dependencies for %s. Running: %s", self.ext_id, " ".join(command))
+
+        try:
+            result = subprocess.run(command, check=True, capture_output=True, text=True)
+            logger.info("Installation successful. Output: %s", result.stdout)
+        except subprocess.CalledProcessError as e:
+            logger.exception("Error during installation. Output: %s", e.stderr)
+
+    def _read_requirements(self) -> str | None:
+        """
+        Read the requirements file and return its content.
+        """
+        requirements_file = f"{self.path}/requirements.txt"
+        if not isfile(requirements_file):
+            return None
+
+        with open(requirements_file) as f:
+            return f.read().strip()
--- a/ulauncher/modes/extensions/extension_dependencies.py
+++ b/ulauncher/modes/extensions/extension_dependencies.py
@@ -8,6 +8,10 @@ from os.path import isdir, isfile
 logger = logging.getLogger()
 
 
+class ExtensionDependenciesRecoverableError(Exception):
+    pass
+
+
 class ExtensionDependencies:
     """
     Manages python dependencies of the extension.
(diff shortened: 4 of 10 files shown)


gornostal:
@friday The implementation ended up much simpler than I thought. It almost seems like I missed something.
I won't merge this yet. I'll run some more tests and update the extension docs.

Please let me know if you think anything is missing.

nix/default.nix
```diff
@@ -119,7 +119,7 @@ let
 
       substituteInPlace \
           ulauncher/modes/extensions/extension_controller.py \
-        --replace-fail '"PYTHONPATH": paths.APPLICATION,' '"PYTHONPATH": ":".join(sys.path),'
+        --replace-fail 'paths.APPLICATION,' '":".join(sys.path),'
```
gornostal, line 122: I haven't used nix before.
As I understand it helps setting up a local dev environment? Curious, what was the main challenge that nix solves for you?

friday: It's a package manager: https://www.youtube.com/watch?v=FJVFXsNzYZQ. I don't use it. It was contributed by @nazarewk.

ulauncher/modes/extensions/extension_remote.py
```diff
@@ -86,6 +87,7 @@ def _get_download_url(self, commit: str) -> str:
 
     def _get_refs(self) -> dict[str, str]:
         refs = {}
+        ref = None
```
friday, line 90: This looks unused?

friday:
> @friday The implementation ended up much simpler than I thought. It almost seems like I missed something. I won't merge this yet. I'll run some more tests and update the extension docs.
> 
> Please let me know if you think anything is missing.

I'm not at all surprised his was simple. I'll  try to look at the PR more and test it in the weekend.

ulauncher/modes/extensions/extension_remote.py
gornostal: This is to fix the "could be unassigned variable" linter error

nix/default.nix
friday: I definitely remember this PYTHONPATH patch was something we talked about when merging it originally. I definitely would want confirmation from some nix user about this.

friday (commented):
I have some small comments/concerns only. Looks great otherwise, and nice work on the documentation as well.

I didn't actually test it yet.

We're not using venvs here, only pip to a `.dependencies` path that we later add to the extension path. I think this is more disk-space friendly approach than #721 (issue created by the virtualenv author btw), and from what I can see it should work as well for us (right @troycurtisjr?).

I'm also considering how to handle upgrading extensions. I think the current logic would delete the dependencies subdir when you upgrade the extension and then reinstall them after, but if you're on the newest version of the extension, then updating won't update the dependencies? This is a reasonable for the PR, but I do want to add some extension update CLI command.

We should consider the `--upgrade` and `--ignore-installed` flag. I would prefer not using (to enable reusing system dependencies), but I'm not sure it's that can lead to similar issues as I had with my+gobject stubs. And if we don't pass `--upgrade`, then the version installed would not upgrade of course. But if we do it will likely also override system packages.

docs/extensions/tutorial.rst
```diff
@@ -265,6 +265,21 @@ Custom Action on Item Enter
...
+If your extension requires any python dependencies, you can specify them in a ``requirements.txt`` file.
+Ulauncher will install them automatically when the extension is installed.
+
+Avoid specifying development dependencies such as mypy, pytest, etc. in the ``requirements.txt`` file — users don't need them.
+Instead, you can create a separate ``requirements-dev.txt`` file for your development dependencies.
+
+It's a good idea to specify the version of the dependencies you are using to avoid breaking changes in the future.
+For example, if your extension requires the `requests` library, you can create a `requirements.txt` file with the following content::
+
+  requests==2.32.3
```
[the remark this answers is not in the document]
troycurtisjr: I also wonder if whether we should use a custom name in the repo instead of co-opting `requirements.txt`. Generally `requirements.txt` is really used for repeatable builds and not generally runtime requirements. Since there is an established purpose for this file which is just tangentially related to what ulauncher needs from it, it seems best not to reuse it.

ulauncher/modes/extensions/extension_dependencies.py
```diff
@@ -0,0 +1,81 @@
...
+
+        try:
+            result = subprocess.run(command, check=True, capture_output=True, text=True)
+            logger.info("Installation successful. Output: %s", result.stdout)
+        except subprocess.CalledProcessError as e:
+            logger.exception("Error during installation. Output: %s", e.stderr)
+
+            raise ExtensionDependenciesRecoverableError(f"$ {command_str}\n{e.stderr}") from e  # noqa: TRY003, EM102
+
+    def _read_requirements(self) -> str | None:
```
troycurtisjr, line 72: I suspect we'll want some protections here. The main one I can think of is an installation of a different ulauncher version than is currently being run.

troycurtisjr (commented):
Left a few initial comments, I'll also try running it locally as well tomorrow night.

nix/default.nix
nazarewk: I don't remember the exact details, but `paths.APPLICATIONS` assumes Ulauncher lives inside a single directory (probably `~/.local/share`).

In Nix it's not a single directory, but a combination of **many** `/nix/store/` PYTHONPATHs: one per dependency. As far as I remember, there was nothing there except for `ulauncher` module without this patch.

nazarewk: Again I don't remember the details apart from the non-Nix way having a good reason to be there, but this seems to be the whole discussion from when I wanted to "upstream" the change instead of using a patch: https://github.com/Ulauncher/Ulauncher/pull/1306

ulauncher/modes/extensions/extension_remote.py
gornostal: It was Pyright. It's probably better to keep this change. If the `try` block fails, the `except` block will try to reference an undeclared variable.

.editorconfig
```diff
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
```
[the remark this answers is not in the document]
gornostal: You're right. I missed the python section below. Reverted.

ulauncher/modes/extensions/extension_dependencies.py
gornostal: Could you elaborate, please? I'm not sure what edge case you have in mind.

friday (approved):
I tested with [the first extension I could find that already had requirements.txt](https://github.com/tuanpham-dev/ulauncher-better-calculator) and it worked. This extension is just a wrapper around py_expression_eval, so if it runs at all and can do computations it works.

The only concern I have now is nix.

Side note: This extension does absolutely nothing our built in calculator doesn't do, so the name is questionable...

docs/extensions/tutorial.rst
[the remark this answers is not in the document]
troycurtisjr: It is the fact that is a standard naming convention that makes me suggest we should be using a different namespace for ulauncher extension purposes. 

It is true that at its core a `requirements.txt` file is just a list of packages to install and thus can be used for many purposes, including for runtime requirements. However, the typical purposes would be for things as described in [the Pip docs](https://pip.pypa.io/en/latest/user_guide/#requirements-files). Things like repeatable builds with the output of `pip freeze`, forcing specific versions of transitive dependencies, etc. I think it is this variety of existing and established usages that means we should have a dedicated name for ulauncher's purposes. Already the docs mention needing to move dev only deps out into a separate file, implicitly acknowledging that any existing `requirements.txt` file is probably for local development setup. A more standardized form of specifying dependencies would be through package meta-data, but I know we aren't planning on having ulauncher extensions turned into packages distributed on pypi. 

Specifying a distinct name I think gives a few advantages:
1. It clearly communicates it's purpose: to install dependencies when installed by ulauncher
2. It doesn't collide with existing usages of any `requirements.txt`
3. Because of point 2, it provides a cleaner upgrade path since no existing `requirements.txt` will get misinterpreted as being for `ulauncher`

ulauncher/modes/extensions/extension_dependencies.py
troycurtisjr: I was actually thinking of ulauncher specifically (i.e. they had something like `ulauncher==5.0.0` or similar), forgetting that ulauncher isn't distributed via PyPi anyway :grin: 

But instead of ulauncher specifically, if the extension specified different versions of ulauncher dependencies, it could break the install. Though I suppose that would simply be breaking the extension itself, and presumably for some version of ulauncher that it actually worked in, they wouldn't have specified known busted packages. And only that particular extension would be busted, not impacting anyone else.

We already have to deal with extension problems being perceived as ulauncher issues, so it probably doesn't change that issue much. Still I'd be tempted to error on specification of packages that ulauncher itself needs, such that only additional packages should be specified by the extension.

I think the messaging interchange is well defined enough that we probably don't have to worry about being overly sensitive to packages being different versions between the ulauncher and extension processes.

docs/extensions/tutorial.rst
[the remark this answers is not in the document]
gornostal: I see your points Troy. I feel different about the purpose of `requirements.txt`, tbh.

I did a small experiment and I learned that the current implementation would "cure" over 80% of the existing extensions that include `requirements.txt`. 

That's my main argument in favor of keeping the naming, besides that it's a convention too.

I downloaded 230 of the most recent extensions, 35 of them had requirements.txt file, and only 5 of 35 failed on `pip install -r requirements.txt`. Sure, some dependencies were development-related (mypy, etc.), but most of them were runtime deps.

ulauncher/modes/extensions/extension_dependencies.py
troycurtisjr: Yeah probably just being overly cautious without a clear failure mode.

docs/extensions/tutorial.rst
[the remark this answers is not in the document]
troycurtisjr: Ultimately the worst case is some grumbling and momentary confusion with extensions authors, and then they can just use a differently named file if they need to use it for one of the other purposes. On the user side, there might just be a few extra deps installed that are only needed for dev and shouldn't cause any real issues.

troycurtisjr:
I also tested locally and things seemed to work well. I also waffle on whether we should use the `--upgrade` option or not. It seems like there are good reasons either way. I'd probably lean towards using `--upgrade` and then the requirements.txt could always be updated to constrain versions if some newer version causes breakage.

troycurtisjr (approved):

ulauncher/modes/extensions/extension_dependencies.py
gornostal: I see what you mean here. 

> if the extension specified different versions of ulauncher dependencies, it could break the install. Though I suppose that would simply be breaking the extension itself

That could definitely happen. It's good that Ulauncher doesn't use too many dependencies.

gornostal:
Thanks for taking a look at this! I really appreciate your perspective on potential "what if" scenarios

PR #1462, opened 2025-05-03
chore: improved python env setup, fixed mypy errors

gornostal:
- removed deprecated config properties from `.vscode/settings.json`
- added `make python-venv` that simplifies local env setup. Now IDE correctly reads Gkt, Gdk imports!
- simplified dev env instructions in CONTRIBUTING.md

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   draft_release:
     runs-on: ubuntu-latest
-    container: ulauncher/build-image:6.0
+    container: ulauncher/build-image:6.1
     steps:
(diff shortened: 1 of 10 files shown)


friday:
Awesome :+1: I never took the time to look into this.

As a :fish: user I appreciate adding the instructions for us too :slightly_smiling_face:

I initially ran into some issues with the gobject stubs not working, resulting in mypy errors. That was because I had the dependencies installed as system packages, and the mypy version was satisfied by my system, but not the gobject stubs. I found a way to fix that: https://github.com/Ulauncher/Ulauncher/pull/1463

gornostal:
Thanks!

PR #1463, opened 2025-05-04
chore: ignore system packages that satisfy the required dev dependencies

friday:
If you have these packages installed on your system pip tries to use those versions.
This has the potential to break our scripts since a global mypy for example won't
work with the pgobject stubs installed for the .venv

This does make it a few seconds slower to re-run "make python-venv" on my system (6-7 seconds in total instead of 3-4 seconds, so no big deal imo).


gornostal (approved):
Good improvement :+1:

PR #1467, opened 2025-05-15
Started updating extension documentation

gornostal:
Started updating extension documentation, learning about new changes in v6 along the way.

@friday I like that you maintained compatibility in the extension API! I think it's the right way for v6. 
This new API looks much simpler and more intuitive, too.

What I want to do next is to make it more type safe and ready for documentation autogeneration. Also by extracting logic from DTO entities.


docs/index.rst
```diff
@@ -60,6 +39,19 @@ Extension Development Guide
...
+===================
+
+.. toctree::
+   :caption: Custom Color Themes
+   :hidden:
+
+   themes/themes
+
+:doc:`themes/themes`
+    Create your own color themes
```
friday, line 52: Related to this, I really think we should recommend people to use https://github.com/Ulauncher/Ulauncher/discussions/categories/themes instead of your gist (I just created this discussion).

I'm not sure what validation we want for it. We can test it here first: https://github.com/Ulauncher/Test-Discussions

friday (approved):
These are all good changes. I think better types might be better as well, but I went the other way and use native python as much as possible, so extension devs don't have to import for example `Query`.

As mentioned earlier I think we should switch to rspress and markdown though, and if we want to keep flowcharts I think https://github.com/mermaid-js/mermaid would be even better for the ease of editing (the new one you uploaded looks nicers, but GitHub markdown supports mermaid natively, and rspress and all similar have plugins for it.

And about the code doc generation, I think we would be better off writing the docs ourselves. The current readthedocs site leaked the internal Ulauncher classes to the extension developer docs. All it really needs is event and actions?  We have simplified both a lot by now.

docs/index.rst
gornostal: I like the discussions approach for themes. I want to create an automation for migrating them from the Gist. Post v6 release though.

PR #1477, opened 2025-06-19
Update webkitgtk parameter to webkitgtk_4_1 for compatibility with latest nixpkgs

j3hn-prof:
This PR updates the `webkitgtk` parameter to `webkitgtk_4_1` to align with the changes made to nixpkgs. The previous attribute name `webkitgtk` has been removed from nixpkgs, and using it would result in a build error: "error: 'webkitgtk' attribute has been removed from nixpkgs, use attribute with ABI version set explicitly".

This change fixes the build issue and ensures that the Ulauncher flake can be successfully built with the current version of nixpkgs.

**Changes**:

- Updated `webkitgtk` parameter to `webkitgtk_4_1` in `nix/default.nix`

**Testing**: The build should now succeed with the latest version of nixpkgs. You can verify this by running nix build or nix flake build after applying this PR.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1738453229,
-        "narHash": "sha256-7H9XgNiGLKN1G1CgRh0vUL4AheZSYzPm+zmZ7vxbJdo=",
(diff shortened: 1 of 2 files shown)


j3hn-prof:
Changed parameter from `webkitgtk_6_0` to `webkitgtk_4_1` due to being the wrong version.

Issue #1484, opened 2025-06-30
feat: Add dynamic getter/setter for the extension to access/set the query argument


friday:
### Description

This should be easier for extension devs, and will allow us to simplify `ActionMetadata` a little to remove the `str` variant.

Setting the query arg should update the query in the UI, and trigger on_input for the same extension again. Optimally it should bypass the regular query parser logic since we know the mode, keyword and extension will stay the same. This is also more "sane" than the old logic, because extensions shouldn't know/handle their own keyword. It's a user preference. And they could intentionally or accidentally trigger other extensions now or other modes. 

There is also an argument for allowing to setting the query arg more as a way to reflect a current change. For example, with our file browser mode if you highlight a result and expand it with alt+enter, the query won't update. So potentially we'd want to support updating the query to reflect what is happening _currently_ But I think that makes everything less straight forward.

### Communication guidelines

- [x] I have read and agree with Ulauncher's communication guidelines and checked the [existing feature requests](https://github.com/Ulauncher/Ulauncher/labels/feature) before suggesting this.

### Feature isn't possible to support as an extension.

- [x] This feature request cannot satisfactory be supported as a third party extension.

### Ulauncher v6

- [x] I have checked and couldn't find my suggestion in v6.

friday:
We went with the `effects` semantics instead, which more makes sense. So this is no longer something we want.

PR #1485, opened 2025-06-30
refactor: move activate_result to mode

friday:
Related to and preparing for #931 and #1484,

I want to align the relationship between results and modes with results and extensions, so that the results have none or very simple logic, and the mode is more similar to the extension.

I realize this is a big change, so no need to review in detail. And I have tested it of course.


gornostal (approved):

gornostal:
I like the change. Definitely makes sense to keep the logic of handling user actions in modes instead of result items.

PR #1490, opened 2025-07-07
fix: kill Ulauncher v5 on the first run of v6 #1093

### LINKED ISSUE #1093: Handle starting v6 while Ulauncher v5 is still running
### Description

If you upgrade to Ulauncher v6 (once released), and try to start it while v5 is running, then you get this error or something similar:
```
Failed to register: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/dbus/service.py", line 662, in _message_cb
    (candidate_method, parent_method) = _method_lookup(self, method_name, interface_name)
  File "/usr/lib/python3.10/site-packages/dbus/service.py", line 252, in _method_lookup
    raise UnknownMethodException('%s is not a valid method of interface %s' % (method_name, dbus_interface))
dbus.exceptions.UnknownMethodException: org.freedesktop.DBus.Error.UnknownMethod: Unknown method: DescribeAll is not a valid method of interface org.gtk.Actions
```

Crashing here is good imo, but this message is very unhelpful.

This is from the old dependency v5 used. The dbus package doesn't exist on my system as of running the code, but it's loaded in memory in Ulauncher v5's runtime. This error also can't be caught (`app.run()` in main.py doesn't fail, the except_hook doesn't catch anything and the application "shutdown" or "name-lost" singals aren't triggered).

One way to "fix" it is doing #1092 However, this change alone would be worse than not crashing, because then two different instances of Ulauncher will run simultaneously, which can cause all kinds of issues.

We should probably try to detect if Ulauncher 5 is responding to the dbus calls and crash with a nicer error message (maybe send a desktop notification as well).

### Communication guidelines

gornostal:
Fixes #1093

### DIFF
--- a/ulauncher/main.py
+++ b/ulauncher/main.py
@@ -81,6 +81,13 @@ def main() -> None:
 
     sys.excepthook = except_hook
 
+    # Ensure that Ulauncher v5 is not running
+    # TODO: Remove this 4-6 months after v6 release
+    # Import here because of the dependency on the logger setup
+    from ulauncher.utils.v5_killer import kill_ulauncher_v5
+
+    kill_ulauncher_v5()
+
     # Migrate user data to v6 compatible
     v5_to_v6()
 
--- a/ulauncher/utils/v5_killer.py
+++ b/ulauncher/utils/v5_killer.py
@@ -1,0 +1,52 @@
+import logging
+import os
+import signal
+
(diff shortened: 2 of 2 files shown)


gornostal:
I agree.

PR #1499, opened 2025-07-09
fix: false warning for git ssh remote format

### LINKED ISSUE #1496: Unsupported URL protocol: "git" when installing an extension using ssh syntax
### Environment

Any

### Bug description

Ex: Installing using the link `git@gitlab.com:friday/ulauncher-locate.git`

It still works, but it warns, and makes some bad assumptions that I'd like to clean up.

### Log output

```shell

```

### Communication guidelines

- [x] I have read and agree with Ulauncher's communication guidelines and checked the troubleshooting page for self-help.
- [x] I have tried to understand the error message and help myself, before reporting it.
- [x] My issue description has clear steps for how to reproduce the issue.
- [x] This issue applies to the whole Ulauncher application, not for a specific extension.

### Ulauncher v6

- [x] I have made sure the bug isn't already fixed in v6.

friday:
Prevents false warning due to url_parts.scheme being "git"

Additionally this was also needlessly complex

fixes #1496

### DIFF
--- a/ulauncher/modes/extensions/extension_remote.py
+++ b/ulauncher/modes/extensions/extension_remote.py
@@ -198,16 +198,12 @@ def parse_extension_url(url: str) -> UrlParseResult:
 
     url_parts = urlparse(url)
     path = url_parts.path[1:]
+    host = url_parts.netloc
+    protocol = "https"
 
     if url_parts.scheme in ("", "file"):
         assert isdir(url_parts.path)
-        host = ""
         protocol = "file"
-    else:
-        if url_parts.scheme != "https":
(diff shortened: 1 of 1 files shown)


ulauncher/modes/extensions/extension_remote.py
```diff
@@ -198,16 +198,12 @@ def parse_extension_url(url: str) -> UrlParseResult:
 
     url_parts = urlparse(url)
     path = url_parts.path[1:]
+    host = url_parts.netloc
```
friday, line 201: `url_parts.netloc` will be an empty string if it's a local file, so we didn't have to handle this differently for the two cases.

gornostal (approved):
Yeah, this looks simpler. Thanks for the quick fix!

Issue #1400, opened 2024-06-28
Wayland positioning via margins


friday:
### Description

I considered that we might be able to solve the Wayland positioning issues by drawing the window over the whole screen, but with a transparent background, and drawing a non-transparent window in the window, but it seems I was over-thinking and @hedgieinsocks [found a workaround much easier than that](https://github.com/Ulauncher/Ulauncher/issues/205#issuecomment-2196228549):
> Since xwayland caused black box flashing for me (https://github.com/Ulauncher/Ulauncher/issues/1399), I've tackled the issue by editing CSS for the custom theme I use. This way I can now force the ulauncher window to appear wherever I please on wayland.
>
> 1) in ulauncher prefs set the window width to max of your display (in my case 1920)
> 2) edit `margin` values under `.app` (in my case `160px 640px 180px 640px;`) 
> 
> I'm using ulauncher-theme-libadwaita theme so that's the line to edit https://github.com/leodr/ulauncher-theme-libadwaita/blob/59ba686189726ba0cbb59c0a60287e48378f7eec/theme.css#L40
https://github.com/Ulauncher/Ulauncher/issues/205#issuecomment-2196228549

I don't have much time to work on or test Ulauncher on other platforms atm, so creating this issue to remember it or get help from volunteers (@hedgieinsocks maybe). I'm interested in the following:
1. Does it work in Gnome wayland? We already have a [working solution with Wayland  Layer shell](https://github.com/Ulauncher/Ulauncher/pull/1172), which all other Major DEs support (also see the general [Wayland compatibility](https://github.com/Ulauncher/Ulauncher/discussions/1353)).
2. Can we calculate the margin to use dynamically in the ulauncher window class
3. Can we support dragging the window by changing the margin?

And thank you hedgieinsocks! :pray: 

### Communication guidelines

- [X] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

### Ulauncher v6

- [X] I have checked and couldn't find my suggestion in v6.

hedgieinsocks:
env: fedora gnome wayland
display: 1920x1200
window width in ulauncher prefs: 1920
custom margins in theme's css:
```css
.app {
  margin: 10em 40em 15em 40em;
}
```
https://github.com/Ulauncher/Ulauncher/assets/26518927/b03c91b6-f2ab-4124-9933-947a30767312

[...]

PR #1502, opened 2025-07-17
fix: replace box shadow workaround with a more direct one

friday:
Window managers without composition get a black box around Ulauncher without this workaround. But we have another issue with the window placement in Wayland that we can't solve without adding margins (#1400).

So this is an attempt to separate the logic for the workaround so we can still set margins.

I'm not sure if it will actually work, because I don't even know how to test it in a DE without a compositor. If full transparency is a problem then this shouldn't work.

### DIFF
--- a/ulauncher/ui/windows/ulauncher_window.py
+++ b/ulauncher/ui/windows/ulauncher_window.py
@@ -126,19 +126,12 @@ class UlauncherWindow(Gtk.ApplicationWindow):
         is_composited = screen.is_composited()
         logger.debug("Screen RGBA visual: %s", visual)
         logger.debug("Screen is composited: %s", is_composited)
-        shadow_size = 20 if is_composited else 0
-        self.window_frame.set_properties(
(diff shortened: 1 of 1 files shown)


ulauncher/ui/windows/ulauncher_window.py
```diff
@@ -301,6 +297,13 @@ def position_window(self) -> None:
             else:
                 self.move(pos_x, pos_y)
 
+        self.window_frame.set_properties(
+            margin_top=margin,
+            margin_bottom=margin,
+            margin_start=margin,
+            margin_end=margin,
```
friday, line 304: it's actually possible to set margin=margin as a shorthand, but we'd want different values for x and y axis most likely.

friday:
Managed to test on i3 and finally confirmed that
1. The main branch works now (automatically detects if composited and disables the shadows.
2. This workaround did not work :(

This makes #1400 trickier

PR #1504, opened 2025-07-20
fix: refactor remote parser and fix url in preferences UI

friday:
The major part of this PR is refactoring. After reviewing #1488 I was confused about my own code, and unsure if it was doing the right thing, and I also wanted to simplify the tests so that they would only check the props that are really affected by the tests (otherwise all tests will break if you change any prop). But that lead to a bigger refactoring to make the parse method and result more self contained, ie: handle all parsing and the host, path and id in one place, so the parse result doesn't have to expose internals.

Adds a new "download_url_template" prop so we don't need to conditionally check the host outside of the parser.

Removes the "use_git" prop, which didn't belong in the url parser anyway (I think I put it there, so not blaming anyone else), and now we can check if we have git and not the "download_url_template" prop instead.

Also replaced TypeDict with BaseDataClass as the baseclass for UrlParseResult. Then used UrlParseResult as the baseclass for ExtensionRemote (avoids some repetitious code).

Rewrote the tests to account for the new changes, to simplify, and as mentioned above to be more isolated (only check the props that has to do with the test). Added some new tests too.

**As for the actual fix**, if you install the url `git@gitlab.com:friday/ulauncher-locate.git`, then the preferences would also use that as the link (which wouldn't work). Now it will use the converted web link if available and allow to copy the link otherwise. 

This PR will introduce a lot of conflicts with #1488. Sorry about that.

### DIFF
--- a/preferences-src/src/components/pages/ExtensionPreferences.vue
+++ b/preferences-src/src/components/pages/ExtensionPreferences.vue
@@ -10,10 +10,16 @@
         <div class="repo" v-if="extension.url">
           <div v-if="extension.commit_hash"><i class="fa fa-code-fork fa-fw"></i><span class="text-monospace">{{ extension.commit_hash.slice(0, 7) }}</span></div>
           <div v-if="extension.commit_hash"><i class="fa fa-calendar fa-fw"></i>{{ extension.updated_at.slice(0, 10) }}</div>
-          <div>
-            <a class="text-muted" href @click.prevent="openUrl(extension.url)" :title="extension.url">
+          <div v-if="extension.browser_url">
+            <a class="text-muted" href @click.prevent="openUrl(extension.browser_url)" :title="extension.browser_url">
               <i class="fa fa-external-link"></i> Open repository
             </a>
+          </div>
+          <div v-if="!extension.browser_url">
+            <i class="fa fa-copy"></i>
+            <a class="text-muted" href @click.prevent
+             v-clipboard:copy="extension.url"
+             :title="'Click to copy: ' + extension.url">Copy url</a>
           </div>
         </div>
         <div class="notes">
@@ -59,7 +65,7 @@
 
     <div class="error-wrapper" v-if="extension.error_type && extension.is_enabled">
       <ext-runtime-error
-        :extUrl="extension.url"
+        :extUrl="extension.browser_url"
         :errorMessage="extension.error_message"
         :errorType="extension.error_type"
       />
@@ -137,7 +143,7 @@
 
       <ext-install-error
         v-if="updateError"
-        :extUrl="extension.url"
+        :extUrl="extension.browser_url"
         :errorMessage="updateError.message"
         :errorType="updateError.type"
       />
--- a/tests/modes/apps/extensions/test_extension_remote.py
+++ b/tests/modes/apps/extensions/test_extension_remote.py
@@ -40,6 +40,9 @@ class TestParseExtensionUrl:
         assert parse_extension_url("https://gitlab.com/u/repo/issues").remote_url == "https://gitlab.com/u/repo.git"
         assert parse_extension_url("https://codeberg.org/u/repo/wiki").remote_url == "https://codeberg.org/u/repo.git"
 
+    def test_browser_url(self) -> None:
+        assert parse_extension_url("git@gitlab.com:user/repo.git").browser_url == "https://gitlab.com/user/repo"
+
     def test_ext_id(self) -> None:
         assert parse_extension_url("https://github.com/user/repo").ext_id == "com.github.user.repo"
         assert parse_extension_url("https://example.co.uk/user/repo").ext_id == "uk.co.example.user.repo"
@@ -73,6 +76,7 @@ class TestParseExtensionUrl:
         result = parse_extension_url("/local/path/to/extension")
         assert result == parse_extension_url("file:///local/path/to/extension")
         assert result.remote_url == "file:///local/path/to/extension"
+        assert result.browser_url == "file:///local/path/to/extension"
 
     def test_empty_path_raises_assertion_error(self) -> None:
         with pytest.raises(AssertionError):
--- a/tests/modes/apps/extensions/test_extension_remote.py
+++ b/tests/modes/apps/extensions/test_extension_remote.py
@@ -5,7 +5,6 @@ import pytest
 from ulauncher.modes.extensions.extension_remote import (
     ExtensionRemote,
     InvalidExtensionRecoverableError,
-    UrlParseResult,
     parse_extension_url,
 )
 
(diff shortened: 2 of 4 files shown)


ulauncher/modes/extensions/extension_remote.py
```diff
@@ -34,6 +34,7 @@ class ExtensionNetworkError(Exception):
 class UrlParseResult(BaseDataClass):
     ext_id: str
     remote_url: str
+    browser_url: str | None = None
```
friday, line 37: It's optional and won't be set if we can't derive the browser url from the input url (unless it's on GitHub, GitLab or Codeberg, the input url must be a remote that can be cloned with `git`, and then it won't be a web url also)

friday: Also, note that this URL can still be a file url (`FILE:///...`), which is why I named it browser_url, not web_url. Both work in a browser/can be browsed, but file urls are not web urls.

preferences-src/src/components/pages/ExtensionPreferences.vue
```diff
@@ -10,18 +10,24 @@
...
+            <a class="text-muted" href @click.prevent="openUrl(extension.browser_url)" :title="extension.browser_url">
               <i class="fa fa-external-link"></i> Open repository
             </a>
           </div>
+          <div v-if="!extension.browser_url">
+            <i class="fa fa-copy"></i>
+            <a class="text-muted" href @click.prevent
+             v-clipboard:copy="extension.url"
+             :title="'Click to copy: ' + extension.url">Copy url</a>
+          </div>
```
friday, line 23: If we have no `browser_url` show a different UI to copy the input URL instead. This is pretty hard to test. It requires a git remote that is not GitHub, GitLab or Codeberg.

(so I tested it by temporarily flipping the `extension.browser_url` and `!extension.browser_url` conditions instead)

ulauncher/modes/extensions/extension_remote.py
```diff
@@ -6,13 +6,13 @@
...
 from shutil import move, rmtree, which
 from tempfile import NamedTemporaryFile, TemporaryDirectory
-from typing import TypedDict
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
 from urllib.request import urlopen, urlretrieve
 
 from ulauncher.config import API_VERSION, paths
 from ulauncher.modes.extensions.extension_manifest import ExtensionIncompatibleRecoverableError, ExtensionManifest
+from ulauncher.utils.basedataclass import BaseDataClass
```
friday, line 15: Our own custom class. It is similar to TypedDict, but props can be accessed with the dot notation too (which I  think is much easier in code).

ulauncher/modes/extensions/extension_remote.py
```diff
@@ -31,45 +31,32 @@ class ExtensionNetworkError(Exception):
...
 
-class ExtensionRemote:
+class UrlParseResult(BaseDataClass):
+    ext_id: str
+    remote_url: str
+    download_url_template: str | None = None
+
+
+class ExtensionRemote(UrlParseResult):
+    url: str
```
friday, line 42: It also gets the other type defs from UrlParseResult.

ulauncher/modes/extensions/extension_remote.py
```diff
@@ -31,45 +31,32 @@ class ExtensionNetworkError(Exception):
...
-        )
         self._dir = f"{paths.USER_EXTENSIONS}/{self.ext_id}"
         self._git_dir = f"{paths.USER_EXTENSIONS}/.git/{self.ext_id}.git"
 
-    def _get_download_url(self, commit: str) -> str:
-        if self.host == "gitlab.com":
-            repo = self.path.split("/")[1]
-            return f"{self.url}/-/archive/{commit}/{repo}-{commit}.tar.gz"
-        return f"{self.url}/archive/{commit}.tar.gz"
-
```
friday, line 66: Replaced by creating a "template" (string with all parts replaced, except the  commit hash, which we replace later).

ulauncher/modes/extensions/extension_remote.py
```diff
@@ -173,55 +147,79 @@ def download(self, commit_hash: str | None = None, warn_if_overwrite: bool = Fal
                         rmtree(self._dir)
                     move(tmp_dir, self._dir)
             commit_timestamp = getmtime(self._dir)
+            return commit_hash, commit_timestamp
```
friday, line 151: Changed to early returns instead of big if/elif statements, because the latter can get confusing and/or break more easily.

tests/modes/apps/extensions/test_extension_remote.py
```diff
@@ -17,172 +16,73 @@ class TestExtensionRemote:
...
-        assert ExtensionRemote("https://host.tld/user/repo").ext_id == "tld.host.user.repo"
-        assert ExtensionRemote("http://host/user/repo").ext_id == "host.user.repo"
-        assert ExtensionRemote("https://host.org/user/repo.git").ext_id == "org.host.user.repo.git"
-        assert ExtensionRemote("http://host/user/repo.git").ext_id == "host.user.repo.git"
-        assert ExtensionRemote("git@host.com:user/repo").ext_id == "com.host.user.repo"
-        # verify sanitizing github/gitlab/codeberg urls, but leave all others
-        assert ExtensionRemote("https://github.com/user/repo/tree/HEAD").ext_id == "com.github.user.repo"
-        assert ExtensionRemote("https://gitlab.com/user/repo.git").ext_id == "com.gitlab.user.repo"
-        assert ExtensionRemote("https://other.host/a/b/c/d").ext_id == "host.other.a.b.c.d"
-
```
friday, line 30: `test_valid_urls_ext_id` moved to the parser since that returns the template now.

tests/modes/apps/extensions/test_extension_remote.py
```diff
@@ -17,172 +16,73 @@ class TestExtensionRemote:
...
-    def test_get_download_url(self) -> None:
-        assert (
-            ExtensionRemote("https://github.com/user/repo")._get_download_url("master")
-            == "https://github.com/user/repo/archive/master.tar.gz"
-        )
-        assert (
-            ExtensionRemote("https://gitlab.com/user/repo")._get_download_url("master")
-            == "https://gitlab.com/user/repo/-/archive/master/repo-master.tar.gz"
-        )
-
```
friday, line 44: Same as above. These tests are moved to the parser since that returns the template now (see `test_download_url_template` method).

tests/modes/apps/extensions/test_extension_remote.py
```diff
@@ -17,172 +16,73 @@ class TestExtensionRemote:
...
 
 class TestParseExtensionUrl:
-    @patch("ulauncher.modes.extensions.extension_remote.which")
-    def test_https_url(self, mock_which: MagicMock) -> None:
-        mock_which.return_value = "/usr/bin/git"
+    def test_https_url(self) -> None:
         result = parse_extension_url("https://example.com/user/repo")
+        assert result.remote_url == "https://example.com/user/repo"
```
friday, line 27: Intentionally changed to test only the affected props, not the full output. Otherwise any change will break all tests.

tests/modes/apps/extensions/test_extension_remote.py
```diff
@@ -17,172 +16,84 @@ class TestExtensionRemote:
...
-            path="user/repo",
-            protocol="https",
-            host="codeberg.org",
+    def test_url_sanitization(self) -> None:
+        assert parse_extension_url("  https://example.com/path  ").remote_url == "https://example.com/path"
+        assert parse_extension_url("https://github.com/user/repo").remote_url == "https://github.com/user/repo.git"
+        assert parse_extension_url("git@gitlab.com:user/repo.git").remote_url == "https://gitlab.com/user/repo.git"
+        assert (
+            parse_extension_url("https://github.com/user/repo/blob/master").remote_url
+            == "https://github.com/user/repo.git"
```
friday, line 38: This test suite cover several different test methods from before (`test_codeberg_url_sanitization`, `test_github_url_sanitization`, `test_gitlab_url_sanitization`, `test_url_with_whitespace`, `test_ssh_url_reformatting`).

tests/modes/apps/extensions/test_extension_remote.py
```diff
@@ -17,172 +16,84 @@ class TestExtensionRemote:
...
-    @patch("ulauncher.modes.extensions.extension_remote.which")
-    def test_no_git_available(self, mock_which: MagicMock) -> None:
-        mock_which.return_value = None
-        result = parse_extension_url("https://example.com/user/repo")
+    def test_ext_id(self) -> None:
+        assert parse_extension_url("https://github.com/user/repo").ext_id == "com.github.user.repo"
+        assert parse_extension_url("https://example.co.uk/user/repo").ext_id == "uk.co.example.user.repo"
+        assert parse_extension_url("https://gitlab.com/user/repo/issues").ext_id == "com.gitlab.user.repo"
+        assert parse_extension_url("https://local/path/to/extension").ext_id == "local.path.to.extension"
+        assert parse_extension_url("https://localhost/extension").ext_id == "localhost.extension"
```
friday, line 51: New tests (copied from https://github.com/Ulauncher/Ulauncher/pull/1488/files#diff-7448673cfee888a244df38ce89fe74837aa4e231457092bd0ecdc4ace3b9a938R196-R205 excdept I changed the GitLab one).

gornostal (approved):
Good stuff :+1: 

I'm going to close #1488. If we need it later we can always pull the code from the PR.

Issue #1150, opened 2022-11-06
feat: manage extensions via CLI commands


FinnHering:
### Description

Edit @friday 2025:
Most of this thread has been about supporting the ground work for CLI commands to manage extensions in general, so to make it easier to pick up the work from that I'm rephrasing this ticket to be about that. The original ticket description and title is now recreated as #1506

Also, this implementation got stuck on and discussed a former issue with the `ExtensionRemote` class that has now been solved. So I marked those parts as outdated.

I think we should aim to support these commands:
* `ulauncher install {url}`
* `ulauncher uninstall {url or app_id}`
* `ulauncher upgrade` (can optionally take an argument, which would make it behave the same as install, but for the initial version we just need it to upgrade all.

It should be possible to this with argparse [sub-commands](https://docs.python.org/3/library/argparse.html#sub-commands), which I didn't know about when we initially discussed the implementation.

Resources:
* https://docs.python.org/3/library/argparse.html#sub-commands
* https://realpython.com/command-line-interfaces-python-argparse/
* The file to change: https://github.com/Ulauncher/Ulauncher/blob/main/ulauncher/cli.py
* Example: https://raw.githubusercontent.com/essembeh/gnome-extensions-cli/refs/heads/main/gnome_extensions_cli/cli.py
* Other examples: pip or pipx source code for exampl (but they are much more complex than the above)
* Draft PR with FinnHerings work in: #1507 (I think we should try to support roughly the same commands in the end, but without click, and without the separate binary. Also, we only have to support `install`, `upgrade` and `uninstall` initially, no aliases, no completion).

friday:
I'm not against adding this back as a cli command, but I want to support all the obvious extension management commands (add/install, remove/uninstall, update, show, etc), and also commands for debugging extensions without copying them to the extension directory. And for migrating extension manifests between API versions APIv2 to v3).

Adding all of that to the current `ulauncher` command and the existing default python CLI parser (which is very limited) would probably create a mess. Imo it's better to make a new `ulauncher-extension` (or preferably something shorter) command and add a more powerful parser like https://click.palletsprojects.com/ for that. If you're up for helping with that it'd be very welcome. I think supporting this particular issue with `ulauncher-extension upgrade` with an optional flag `--restore` might make sense for example.

Not expecting you to do all that of course, just clarifying the context I think is right for this.

FinnHering:
Sounds awesome! 
So initially – I propose – we'd have the following options available:

[...]

PR #1508, opened 2025-07-25
feat: basic CLI for extension management

gornostal:
#1150

I will extract logger config in a separate PR. I also want to add a bit of color to the logs and remove unnecessary logs (such as `Using selector...`)

### DIFF
--- a/ulauncher/cli.py
+++ b/ulauncher/cli.py
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
 import argparse
-from functools import lru_cache
+from functools import lru_cache, partial
 from gettext import gettext as _
 
 from ulauncher import version
+from ulauncher.modes.extensions.extension_cli_handlers import (
+    install_extension,
+    list_active_extensions,
+    uninstall_extension,
+    upgrade_extensions,
+)
 
 
 class CLIArguments(argparse.Namespace):
@@ -24,7 +30,7 @@ def get_cli_args() -> CLIArguments:
     # but GTK adds in their own options we don't want like --help-gtk --help-gapplication --help-all
     parser = argparse.ArgumentParser(
         None,
-        None,
+        "%(prog)s OPTIONS or ARGUMENT",
         "Ulauncher is a GTK application launcher with support for extensions, shortcuts (scripts), calculator, file browser and custom themes.",  # noqa: E501
         add_help=False,
     )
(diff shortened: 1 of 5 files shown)


friday (commented):
Looks and works great :muscle: But I have some opinions/feedback anyway:

1. `Using selector: EpollSelector` is in the output log because we're using async. I think we should prevent that: https://stackoverflow.com/questions/60503705/how-can-i-avoid-using-selector-epollselector-log-message-in-django (this can be done in a separate PR or directly to the main branch if you prefer).
2. I don't like the alias `un` for `uninstall` - Can we replace it with `rm`? I think that is much better and more memorable (people already know it). If you think that makes it better to switch to `remove` also (I can go with either). I have been conflicted about this before, and went with `uninstall` because it's symmetric with "install", but there is no good abbreviation for it.
3.  I tried "ulauncher e i ." and other relative urls (ex `../ulauncher-hash`) and it didn't expand them. I thought that if we specified the parser format for the install command to be a file or URL, that it world do that for us (the GTK arg parser does I think, but it's worse in other aspects). We can maybe handle this ourselves with https://docs.python.org/3/library/pathlib.html (something like `Path(os.getcwd(), pathArgument).resolve()`), although I'm not sure if `os.getcwd()` gets the "right" working dir.

Also, I personally prefer no `extension` in the argument, ie `ulauncher install` or `ulauncher i` etc directly (easier to use/remember imo). I know this would be a large change, and that I have been contradictory about it. A lot of my comments were written years ago and I wanted to reserve the possibility to use other positional commands that wasn't about extensions or if some command could apply to both extensions and ulauncher (like start/enable/configure), but I think now that was over-thinking it. We can solve that problem when/if we get to it.

gornostal:
I will update as you suggested.

> I personally prefer no `extension` in the argument
I was going to do it like that, but then realized `ulauncher upgrade` may be confusing. People may think it will upgrade Ulauncher.
Have you considered that?

gornostal:
Pushed a new commit.

I'll have to think about installing extensions while the app is running.

gornostal:
Pushed one more commit, which solves the problem with extensions installed via CLI not being started in the main process.

Better to review by commit.

No hurry with this. Whatever time it takes, it's fine.

friday:
> `ulauncher upgrade` may be confusing. People may think it will upgrade Ulauncher.
> Have you considered that?
> 
> Changing to a flat command structure is no big deal. So I can do that if you still prefer.

I agree to an extent. At least if you read that command first and try to guess what it means. I think Linux users are used to package managers or graphical front-ends for them and would look at where they usually upgrade apps to upgrade Ulauncher though, rather than trying to use the ulauncher cli to upgrade itself. The `--help` command also already specify that this command and all the other ones added are for extensions, and if they would still run the `ulauncher upgrade` expecting the app to upgrade, it won't harm them. The output will make it clear that it is to upgrade the extensions.

So I do think flat command structure is better still, but neither solution are without problems. If you want a third opinion tag Troy as a reviewer :)

Also related to this, I was reminded of an old idea before when testing this PR, so I made a ticket for it: #1510 If we do this ticket we would have some positional arguments for Ulauncher that are for extensions and some that are not.

> Pushed one more commit, which solves the problem with extensions installed via CLI not being started in the main process.

You're working very fast! :muscle: I do think using the socket server and json framer etc for this adds a lot of overhead for a  one-off message though. I have a pretty good idea how to do this with Dbus instead. I think that would need something like ~10 lines of code for passing the message. If you don't mind I can push this as a commit to this branch/PR? It's easier for me to do this than to explain it.

gornostal:
>  I do think using the socket server and json framer etc for this adds a lot of overhead for a one-off message though.

My intuition was telling me it was a bit of overkill. Thanks for calling that out!
I don't mind if you push Dbus solution instead. Feel free to remove my commit.

gornostal:
Thanks!
I’ll it flat.

gornostal:
> This branch cannot be rebased due to conflicts

@friday I tried to rebase all commits, but it wasn't easy. I prefer squashing all in one and then rebasing. Any objections?

friday:
> > This branch cannot be rebased due to conflicts
> 
> @friday I tried to rebase all commits, but it wasn't easy. I prefer squashing all in one and then rebasing. Any objections?

I rebased the whole branch and pushed and compared the output. It wasn't hard when using `git rebase main` from the cli (only one easy conflict, which I already knew about). Imo GitHub's web UI makes conflicts more complicated than the CLI.

I don't mind of we squash merge the branch though. Most of the commits are just fixes for the CLI. We can do that and still be able to see the different commits in this PR later if we need to retract for some reason/bug.

troycurtisjr (approved):
I think it looks really good, and I think the flat args is probably the way to go. 

I also don't think nesting under `extensions` would be bad, since it reserves the namespace and makes the functionality explicit. I don't think this CLI is likely something that is interacted with a ton, like some kind of shell interface for some utility that is used many times per day. Since it would be an occasional thing I don't know that having a short CLI argument set is that needed, and sometimes having something explicit is more clear, especially when you don't use something frequently.

That said, I also don't really see a strong need to reserve the top-level namespace of subcommands, since I don't imagine this will become such a rich set of capability that it would be necessary. I expect this will be mostly used for scripting a handful of actions. And if there really do become potential conflicts in the future for something like `install`, then we can also just do something like `install-snippet` to make the new command explicit.

Since the flat approach is simpler, both in implementation and in usage, and the complexity is unlikely to be needed. I agree this is the way to go!

friday:
I merged the Gnome Wayland positioning fix. Going to try to set a more somewhat reasonable margin-top, then make another beta release (likely the last one). This PR we can merge after that, and that means we will have more time to use it and revise it before users do (except Arch users who uses https://aur.archlinux.org/packages/ulauncher-git).

friday:
Just made the release. Will merge now. Rebased to fix small conflict and lifted out f42e4df7 and e3a1a4dd as separate commits to main.

PR #1511, opened 2025-07-27
chore: use python3-pip as a build frontend instead of setuptools

### LINKED ISSUE #1486: chore: update build to avoid running `setup.py` directly (will stop working by 2025-10-31)
### Environment

Arch

### Bug description

I tried to fix this already (see #1311), but couldn't get it to work with the easy workaround. Might need to move to meson/ninja.

This strictly speaking doesn't have to be fixed in v6, but I'm adding it to the milestone because I don't want to forget it.

### Log output

```shell
running install
/usr/lib/python3.13/site-packages/setuptools/_distutils/cmd.py:90: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        By 2025-Oct-31, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
  self.initialize_options()
```

### Communication guidelines

- [x] I have read and agree with Ulauncher's communication guidelines and checked the troubleshooting page for self-help.
- [x] I have tried to understand the error message and help myself, before reporting it.
- [x] My issue description has clear steps for how to reproduce the issue.
- [x] This issue applies to the whole Ulauncher application, not for a specific extension.

### Ulauncher v6

- [x] I have made sure the bug isn't already fixed in v6.

gornostal:
Additionally:
- Fixed chagelog file warnings. The problem was that dch --empty creates a changelog entry without any change description, which causes dpkg-buildpackage to expect change data but find only the trailer line (the author/date signature).

Fixes #1486


makefile
```diff
@@ -163,7 +163,7 @@ prefs: # Build the preferences web app
 
 sdist: manpage prefs # Build a source tarball
 	@set -euo pipefail
-	# See https://github.com/Ulauncher/Ulauncher/pull/1337 for why we're not using setuptools
+	# See https://github.com/Ulauncher/Ulauncher/pull/1337 for why we're not using setuptools directly
```
friday, line 166: This change makes no sense. Did you actually read the job name, the code and the link I put in before you changed this? :thinking:

friday (commented):
I don't know how to test this. `make deb` seems to work, but the real problem is on Arch.

Except the sdist comment change it looks good as a solution for debian, but we should not mark this as a fix for #1486 until it works for Arch (maybe it's ok that it's slow though).

debian/rules
```diff
@@ -4,11 +4,18 @@ export DEB_BUILD_OPTIONS=nocheck
...
 
 %:
-ifneq ($(shell dh -l | grep -xF translations),)
-	dh $@ --with python3,translations --buildsystem=pybuild
-else
 	dh $@ --with python3 --buildsystem=pybuild
-endif
+
+override_dh_auto_build:
+	python3 -m pip wheel --no-build-isolation --no-deps .
```
friday, line 10: When I tested replacing `python setup.py build` with this on my Arch system it gets significantly slower: 
image

makefile
gornostal: AI added that. It made sense to me, so I kept it.
I will revert it. No big deal.

gornostal:
I documented how I tested here: https://github.com/Ulauncher/Ulauncher/wiki/Development-and-testing
And I did successfully test `deb`.
I'm going to re-learn how to make `rpm` builds now and verify them too.

gornostal:
oh, well
I don't know where I was looking.
Then there's more work here.

gornostal:
OK. I pushed new changes: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ulauncher-git
It takes the same time to build: ~15s. It also builds preferences. This is the most time-consuming part.

I still need to test RPMs...

friday:
> typing_extensions

That was my bad. Just fixed that and pushed to main

gornostal:
Could you share the output of "makepkg -Cifs" please?

gornostal:
Awesome!
Thanks for helping with this!

I should probably push `--ignore-installed` to `debian/rules` too, right?

gornostal:
Done

friday:
Unfortunately I had to revert this PR because it didn't build on Ubuntu 22.04 and up, see https://github.com/Ulauncher/Ulauncher/issues/1486#issuecomment-3177052234

But the Arch AUR change was the urgent part, and that's not reverted.

friday:
> Thank you.
> 
> I should have tested in launchpad too.

I don't even know how to do that. I just tested in docker. And apparently that wasn't enough. After trying again it didn't work so I had to revert again.

PR #1512, opened 2025-07-29
refactor(ext-socket-server): remove logic to start self if __main__

friday:
I never touched this, but I can't imagine it's anything other than a leftover from some testing.

This module should never be `__main__`


gornostal (approved):
I guess i was debugging it and forgot to remove this thing.

PR #1513, opened 2025-07-30
perf: defer expensive window operations

friday:
1. Sets the window to be invisible (opacity=0) initially
2. Moves all the expensive/slow styling and positioning logic to a new method `apply_styling()`, which also removes the opacity=0
3. If not is_composited, runs `apply_styling()` before showing the window (because opacity doesn't work), otherwise wait until after the initial render.

In my profiling it now takes about ~65ms for the window to render and be able to get the keyboard input (~35% speed improvement). This seems to be the last improvement we can do. The rest is almost all GTK internal logic we can't avoid (even if I removed everything from the elements in the window it would only save a couple of ms).

### Profiling results
See 7dcb5d9d for reproduction, this also removed migration and kill_ulauncher_v5 scripts since they will be gone some time after v6 anyway)
image

### DIFF
--- a/ulauncher/ui/windows/ulauncher_window.py
+++ b/ulauncher/ui/windows/ulauncher_window.py
@@ -29,13 +29,14 @@ class UlauncherWindow(Gtk.ApplicationWindow):
     settings = Settings.load()
     query_handler = QueryHandler()
 
-    def __init__(self, **kwargs: Any) -> None:  # noqa: PLR0915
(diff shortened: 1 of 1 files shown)


friday:
Don't really need reviews for this imo. Just testing, but I already tested on Gnome X11, Gnome Wayland and i3, so I'm merging. We can always fix stuff later.

gornostal:
great results!

Issue #1527, opened 2025-08-26
Improvement in the error message for missing prerequisites during development.


Mr-Sunglasses:
### Description

While building and running the project from the source using  using `make run` , if yarn is not installed it gives the error

```
find: ‘data/preferences’: No such file or directory
bash: line 12: build: command not found
make: *** [makefile:150: prefs] Error 127
```
but it does not says that yarn is not installed, which makes it difficult for any dev to understand why it breaks apart from reading the makefile. It is good to check if yarn is installed during the `prefs:` step in makefile and log in the error message that yarn is not installed.

### Communication guidelines

- [x] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

### Ulauncher v6

- [x] I have checked and couldn't find my suggestion in v6.

PR #1528, opened 2025-08-26
fix: check for yarn when running prefs in makefile

Mr-Sunglasses:
fix: #1527 

Before:

```
find: ‘data/preferences’: No such file or directory
bash: line 12: build: command not found
make: *** [makefile:150: prefs] Error 127
```

After:

```
find: ‘data/preferences’: No such file or directory
yarn is not installed or not executable.
yarn is needed to set up the local build environment.
Please visit https://classic.yarnpkg.com/en/docs/install to install yarn.
make: *** [makefile:150: prefs] Error 1
```


friday:
I renamed both of your PRs btw. Even if both were fixing things they are not fixes according to the conventional commits:

https://cheatography.com/albelop/cheat-sheets/conventional-commits/

PR #1532, opened 2025-08-28
refactor: preserve function metadata in debounce with functools.wraps

Mr-Sunglasses:
refactor: preserve function metadata in debounce with `functools.wraps`


gornostal:
Thanks @Mr-Sunglasses!

Issue #1505, opened 2025-07-20
feat: install or run/debug extension from local path with uncommitted changes


friday:
### Description

The current way to install from a local path is using git, will not copy any changes that haven't been committed. You also can't conveniently do this from the CLI, although that would have been part of #1018.

There are multiple ways this can be improved:
1. Use regular Copy instead of git for this case (will not get a proper git hash or date)
2. Make a new temporary commit before, then reset it (feels ugly and will get a fake hash or date)
3. Add the ability to run or debug an extension from Ulauncher without installing it (will need a lot more work than the other solutions).

This possibly substitutes #1324 and #1488

### Communication guidelines

- [x] I have read and agree with Ulauncher's communication guidelines and checked the [existing enhancement requests](https://github.com/Ulauncher/Ulauncher/labels/enhancement) before suggesting this.

### Ulauncher v6

- [x] I have checked and couldn't find my suggestion in v6.

gornostal:
I can see the following solution:

1. Replace the existing "install from the local path" behavior (one that uses the last git commit) with a symlink saved to the extension discovery directory. Ulauncher should be able to start a symlinked extension.
2. Add a checkbox `Run with debugger` to the extension preferences UI. When checked, it will restart the extension subprocess with `--listen 0.0.0.0:5678 --wait-for-client`
3. Provide instructions on the prefs UI on how to connect with a debugger (VSCode's launch.json example, etc.)

@friday how does that sound?

friday:
So I'm thinking:
1. Dev runs `ulauncher [ext-debug-command] .`
2. Instructions are printed in the terminal. Press y to continue or maybe multiple choices for debug or testing without a debugger.
3. If continuing, Ulauncher is started in extension debug mode for this specific extension (set some global state ex `debug_extension=/path/to/ext`
   * Checks if Ulauncher is already running normally and kills that pid before starting itself (if using systemd we also have to stop that first)
   * Extension finder checks for `debug_extension` and if it is set, it should only "find" this path.
   * Everything else must use the extension finder logic to fins extension paths (extension manifest, settings etc) - I think this may be the case now though (it used to be more hard coded than it is now).
 4. Press ctrl+c to exit when done. This should probably re-enable the systed service and restart Ulauncher it it was running before.

[...]

PR #1534, opened 2025-08-30
refactor: use stdin and stderr for extension communication

friday:
### TLDR
I'm in the process of first fixing the current implementation (I left out some things because something else was blocking). After that I will try to rewrite this solution to use socket pairs instead.

### The problem

The current extension communication (although a big upgrade from the web sockets), is still very complicated, and separates the extension communication from the normal extension classes in a way that makes it very difficult to control. And that lead to hacky workarounds like the async `ExtensionController.start()` [looping while waiting](https://github.com/Ulauncher/Ulauncher/blob/18bbddde1f0021c8916bec59862fc3cc2363908c/ulauncher/modes/extensions/extension_controller.py#L327-L335) for [the extension socket server to set the `is_running` property to `True`](https://github.com/Ulauncher/Ulauncher/blob/18bbddde1f0021c8916bec59862fc3cc2363908c/ulauncher/modes/extensions/extension_socket_server.py#L82-L83):

Another example is that the `ExtensionSocketController` init method takes a list of `ExtensionSocketController`s (that list is co-owned by the ext socket server and all the socket controllers, so that it can attach and detach itself from it (I think I may have created that mess trying to clean up some other mess, and it was never meant to be a long time solution).

This also blocked me from implementing #1063. I made a prototype to test implementing this (#1020), but the extension mode really had no good way to handle starting and stopping extensions on demand and interacting with them immediately (I made a hacky workaround that didn't perform well), but with this PR that shouldn't be a problem.

This is something I wanted to rewrite for a very long time, and I started experimenting on this approach years ago, but I could never figure out some of the details, and I created no ticket for it, since I wasn't fully convinced it was a good idea.

### TO DO
- [x] Ensure outdated responses are discarded/ignored if the input has changed (done outside of this branch in 1d335a55)
- [x] Ensure outdated responses are discarded/ignored if the active extension has changed
- [x] Fix "Loading..." message (8875efe7 or if PR: #1540 will get merged).
- [x] Support "query debounce" (0fb63645).
- [ ] Rewrite, using socket pairs: https://github.com/Ulauncher/Ulauncher/pull/1534#issuecomment-3243560666
- [ ] Make sure the code is async safe
- [ ] Add tests maybe?

### Status / Concerns

1. Will debugging work if the process can't be started separately?
2. **With the stdio solution:** We don't have exclusivity to stderr. Extensions can/will use that too. As it is now extensions can print JSON to stderr, and then the app will think it's a message for it to parse.
   * I'm not sure if this could be "exploited" in any way worse than what extensions can already do?
   * The current logic prints stderr output that is not valid JSON (and also store them for error handling messages), but  Extension devs might want to print JSON to debug their extensions though, and then they might end up not seeing that if they print to stderr.
   * Possible workaround: Change the delimiter to "\n"? We could actually generate a unique delimiter per runtime and use in case we think 3 is not safe and extensions could try to hijack the communication by printing json to stderr.

### Performance and resource usage

This change was not really motivated by performance, but I also did some brief non-scientific tests and it seems each extension use 0.6-0.7mb less ram now and Ulauncher itself uses less ram also (1.3mb or something for me, but it probably depends on how many extensions).

BUT again this is a WIP draft, and it's missing some features

### DIFF
--- a/ulauncher/modes/extensions/extension_controller.py
+++ b/ulauncher/modes/extensions/extension_controller.py
@@ -8,7 +8,7 @@ import tempfile
 from datetime import datetime
 from pathlib import Path
 from shutil import copytree, rmtree
-from typing import Any, Iterator
+from typing import Any, Callable, Iterator
 from weakref import WeakValueDictionary
 
 from ulauncher import paths
@@ -27,6 +27,7 @@ from ulauncher.modes.extensions.extension_manifest import (
 )
 from ulauncher.modes.extensions.extension_remote import ExtensionRemote
 from ulauncher.modes.extensions.extension_runtime import ExtensionRuntime
+from ulauncher.utils.decorator.debounce import debounce
 from ulauncher.utils.get_icon_path import get_icon_path
 from ulauncher.utils.json_conf import JsonConf
 from ulauncher.utils.json_utils import json_load
@@ -74,6 +75,7 @@ class ExtensionController:
     state: ExtensionState
     manifest: ExtensionManifest
     is_manageable: bool
+    debounced_send_message: Callable[[dict[str, Any]], None]
     _state_path: Path
 
     def __init__(self, ext_id: str, path: str) -> None:
@@ -81,6 +83,7 @@ class ExtensionController:
         self.path = path
         self.manifest = ExtensionManifest.load(path)
         self.is_manageable = extension_finder.is_manageable(path)
+        self.debounced_send_message = debounce(self.manifest.input_debounce)(self.send_message)
         self._state_path = Path(f"{paths.EXTENSIONS_STATE}/{self.id}.json")
         self.state = ExtensionState.load(self._state_path)
 
--- a/ulauncher/modes/extensions/extension_mode.py
+++ b/ulauncher/modes/extensions/extension_mode.py
@@ -54,7 +54,7 @@ class ExtensionMode(BaseMode):
                 "args": [query.argument, trigger_id],
             }
 
-            ext.send_message(event)
+            ext.debounced_send_message(event)
             return
 
         msg = f"Extension could not handle query {query}"


friday:
@troycurtisjr @gornostal This is WIP, but feedback would be appreciated (see Status/concerns for my own concerns)

troycurtisjr:
My general thoughts on the approach:

In my experience overloading the usage of things is asking for complication, so reusing the general stdin/stdout for this specialized purpose, especially with a pile of existing extensions, seems like trouble that can be easily avoided. There are of course plenty of examples of ecosystems that do this, the primary MCP interfaces do this for instance. However, those were created from the start using it, which means any extensions/implementations had to be created with that in mind. 

I don't think it is the case that just grabbing sys.stdin/sys.stdout will necessarily work, as that is a Python layer. The actual file descriptors on the process are the same, and I believe will be inherited by any thing the extension process launches. I suspect going this path will mean continual battling of extension behavior corrupting the stdout fd or reading improperly from the stdin fd.  There are ways to do this, using dup systems calls, such that the actual underlying file descriptors are properly managed, but that quickly defeats the goal of simplification.

We do have the advantage in this case of controlling both sides of the communication channel, so I think that with that in mind using a newline to delimit things probably works fine. While the newline is perfectly valid in a json object, it also has no meaning and never has to be present, thus we can always ensure we only ever produce minimized json objects, and the newline will be safe enough. It does require such text oriented transport like json, and we couldn't migrate to a different more efficient and more well defined interchange protocol, but then again we aren't using that today so we don't necessarily need to build the protocol for that maybe-future-maybe-never need. Using newlines means that we can easily just use the already implemented `read_line_async()` and thus don't need the explicit framer. While it is less capable and flexible, it is workable under the current conditions.

I think it is pretty clear that this PR does represent a simplification, in that it deletes classes and associated code, and leverages more builtins (like basically replacing the `Framer` class with `read_line_async`). But some of the simplification is accomplished by removing features, which work for now in the happy path cases, but I suspect we'll ultimately add in more protection code to handle the new edge cases we just haven't found with this implementation yet.  A few capabilities that I can notice:

* Previously sends were async, now they blocking. The means the remote extension must be alive and processing messages correctly or it will block Ulauncher
* extension/get-all & extension/add were previously async, but are now sync, potentially blocking the main loop
* Previously extensions were running when `ExtensionControl.is_running`  was True, now it just means that it process was launched and has not yet exited. While this *could* be one definition of running, previously running meant the extension at least got initialized enough to check in. 

I think some of the reason from moving from immediate interaction is likely related to this:

> but the extension mode really had no good way to handle starting and stopping extensions on demand and interacting with them immediately 

However, I think this is only an illusion. While there is less explicit code is in place with this PR, all the underlying constraints are the same and this simplification trades control for simplicity. For the happy-path case, this is probably fine. However, as soon as an extension isn't well behaved, it risks locking up the main process. async/non-blocking code is generally more complex and certainly more challenging to follow, especially if the reader isn't used to it, but there are very good reasons to implement in more of a callback manner.   I suspect we could likely do some class updates to make better use of the async Python API to make using the code more clear. Less callbacks, and just using "await", if we think that could make the code more clear.

gornostal:
Sounds like @troycurtisjr speaks for experience :+1:
I don't have anything more to add.

> This also blocked me from implementing https://github.com/Ulauncher/Ulauncher/issues/1063. I made a prototype to test implementing this (https://github.com/Ulauncher/Ulauncher/pull/1020), but the extension mode really had no good way to handle starting and stopping extensions on demand and interacting with them immediately.

That's more of a code organization and interfaces problem than extension architecture, right? If so, it could be solved with refactoring, I guess.

friday:
@troycurtisjr Thanks for the feedback.

The first couple of paragraphs doesn't apply any more. They were things I considered experimental myself and I discovered the `sys.stdout = sys.stderr` redirect indeed didn't work with subprocesses (unless they captured the output). So I gave up the idea of exclusive access to stdout, removed that redirect, and changed to use stderr to send messages to the Ulauncher app instead of stdout. This should give us fewer messages to handle and parse, and we already had a stderr listener before (colleting error messages), so we now have one less stream to handle in ExtensionRuntime.

If the listener gets data that isn't JSON, it is printed rather than treated as an error.

I force-pushed these changed [just before you posted](https://github.com/Ulauncher/Ulauncher/compare/947ad3e1dab78d508257b9f07621780b4d8a2ea0..4e60eedbc0e3ec5c7a21e7acb1dd60eed7f51c9d).

> We do have the advantage in this case of controlling both sides of the communication channel, so I think that with that in mind using a newline to delimit things probably works fine

I was not really wondering in terms of being standard/conventional, but more so "safe", as in preventing breaking for extension devs. If they print JSON to stderr they probably expect to see the output in the log. The other possible concern is that they try to "hijack" Ulauncher communication by printing JSON to stderr. Although they can already do this since python doesn't have the concept of "private" properties and they already have access to `._client` now.

We could generate a unique hash every time we start an extension, and use that as the delimiter, or wrap the data in some way, ex `["SOME_SECRET", actual_message]`. I think I'll keep the newline delimiter and handle it this way instead.

>  We couldn't migrate to a different more efficient and more well defined interchange protocol, but then again we aren't using that today so we don't necessarily need to build the protocol for that maybe-future-maybe-never need.

This is not impossible, but very hypothetical scenario. I don't think it  would make sense for us to switch from JSON unless there was a noticeable performance issue with JSON. With our usage I see no way that would ever be the case anyway. I think  this makes more sense if you run a server handling thousands of requests per second, or if you want a typed transfer. In both cases I think you shouldn't use Python if you need that. 

Technically stdin/out/err does support binary, but it would be very flaky without exclusive access to stderr.

I haven't tested this with large base64 encoded images though, which is something @gornostal might want to support for his `ImageResult` proposal. I guess I can add that to the list of concerns to stress test (but I think the old socket communication would not be much better if there is a problem with JSON).

> But some of the simplification is accomplished by removing features, which work for now in the happy path cases, but I suspect we'll ultimately add in more protection code to handle the new edge cases we just haven't found with this implementation yet.

This is a WIP draft, so it's meant to satisfy the "happy path" first. I definitely don't mind feedback about specific details, but so far no one has mentioned anything fundamentally flawed about this approach that can't be worked around, so I'll continue working.

We of course lose is the ability to switch back to a binary format (as mentioned) and the low level control. But we weren't taking advantage of those anyway.

> * Previously sends were async, now they blocking. The means the remote extension must be alive and processing messages correctly or it will block Ulauncher

I don't know if this is a real issue in practice. If the process is not alive it should not be in `extension_runtimes` and then the controller will should just drop the message (silently though, which might be bad). I still pushed f03d8052 to switch to async methods since it was a small change and feels like a good idea in general. But I think I messed something up there because afterwards I got strange behavior, so I reverted it. In addition to this, we probably need a message queue to go this route, or the order the messages are sent are not guaranteed I think?

> * extension/get-all & extension/add were previously async, but are now sync, potentially blocking the main loop

This is not really true. The preferences "routes" are async and they are accessed from a method that runs each request in a new thread.

Also, while `ExtensionController.start()` is synchronous now, it's not more blocking than it was before. It works just like before by calling  `ExtensionRuntime.__init__` and creating the new subprocess etc, without wrapping it in some `asyncio.to_thread` or similar method, so it runs in the main thread and blocks the event loop until it's done, but it's very fast, so unless you have an excessive amount of extensions I don't imagine it's a problem we really need to handle. Maybe it's worth testing if this can introduce issues though. I never considered this could be a problem since we do blocking i/o operations all the time (in particular reading and writing conf files).

> Previously extensions were running when ExtensionControl.is_running was True, now it just means that it process was launched and has not yet exited. While this could be one definition of running, previously running meant the extension at least got initialized enough to check in.

This is a feature, not a bug. We have no reason to do this any more because the communication is ready as soon as the `ExtensionRuntime` instance is created. This is probably the biggest benefit of this approach.

> However, I think this is only an illusion. While there is less explicit code is in place with this PR, all the underlying constraints are the same and this simplification trades control for simplicity. For the happy-path case, this is probably fine. However, as soon as an extension isn't well behaved, it risks locking up the main process. async/non-blocking code is generally more complex and certainly more challenging to follow, especially if the reader isn't used to it, but there are very good reasons to implement in more of a callback manner. I suspect we could likely do some class updates to make better use of the async Python API to make using the code more clear. Less callbacks, and just using "await", if we think that could make the code more clear.

You clearly don't like this approach, but unless there is a real world problem inherent with this solution I will just continue working on the PR until I worked out the kinks. This is not rocket science. Using native Python for this is no less safe than these Glib/Gio methods, and I want the code to be maintainable and easy to reason with, which it's not now.

troycurtisjr:
I tend to have opinions on many specifics, but I've learned mostly they are just preferences and other ways are fine as well. However I do think there are a couple of core issues with the design, which if addressed I think would make things more robust and I would be content (not that my contentment is a necessary thing!)

1. Overloaded communication channel doing different jobs, distinguished not by protocol, but by heuristic that probably works most of the time
2. Blocking io to a finite buffer to a remote process that can behave arbitrarily.

Like in Ghostbusters, crossing the streams causes trouble. Stderr had a typical purpose, and even in the context of ulauncher it has been in use already for transmitting log messages. Overloading that increases complexity, not reduce it. Complexity is when there are multiple choices/branches. And in this case it isn't even well defined at a protocol level, instead we need to use best effort and a heuristic (try to parse it as a json object, if it fails, it is likely a log message, though we can't tell the difference between that and a merely a poorly formed message). The potential need to put a hash in the message is a smell indicating that inherently ill-defined overloaded use.

It is true there is a lot of blocking io already happening, though local file io isn't really a concern. If there is extreme latency in a user's disk access, I suspect ulauncher will not be the primary impacted app. Doing comms with a remote process however is a different beast. 20 years of real world network application development has repeatedly reinforced that a remote process cannot be relied upon to respond in a timely manner, even if it is running on the same machine 😄 all it takes is an extension establishing a tcp connection with a remote system, and that connection stops responding. That extension can then get stuck in a tcp retry back off that can take up to two weeks to give up. If that blocks reads, the os buffer of the stdin pipe can fill up, and will block the writes from the ulauncher process. This will not be immediate, the messages would need to build up for a bit, but if nothing is reading from the other side it will eventually block.

This is related to the very real distinction between an extension being launched, not dead, but also not necessarily running. However, if the writes are non blocking, the worst that can really happen is an extension not responding. So if registration and checkin from extensions isn't desired, I think it seems reasonable to leave that state distinction out.

I suspect implementing the framer added a lot of the perceived complexity in the current implementation, especially for readers unfamiliar with binary level protocol design. Especially when implemented in Python, since it uses a lot of uncommon (to Python code) byte manipulation. And if binary data is something we expect to support, it is arguably overkill. A strong case could be made that if we really needed to support something like that we should really be using a messaging or rpc library instead of rolling our own.

So in my view, if we have a dedicated extension messaging channel, and non blocking reads and writes  to that channel, the rest is just details and preferences. If we don't have this, I don't think things will immediately fall apart, they will just introduce subtle hard to reproduce bug reports of hangs and missed messages.

friday:
> dedicated extension messaging channel

That's what we fundamentally can't have with this approach. But it's not really a dealbreaker functionality-wise. Using stderr non-exclusively works just fine.

> Stderr had a typical purpose, and even in the context of ulauncher it has been in use already for transmitting log messages.

Which it still is. And even for that case it was really a best effort approach. I changed the log we set up ourselves to go to stdout instead, but there will still be other messages to stderr, and that is fine. 

> The potential need to put a hash in the message is a smell indicating that inherently ill-defined overloaded use.

This is a very dramatic way of saying "I don't like it" :joy:

I have already done something similar for an extension, and while it's not proudest work, it was necessary, and it's working:

https://github.com/friday/ulauncher-clipboard/blob/main/managers/clipster.py#L14
https://github.com/friday/ulauncher-clipboard/blob/main/managers/clipster.py#L63-L65

In the case of this PR I think it's actually less problematic than this solution, because we have more options for how to handle it. For example `["ulauncher-json-message", actual_json_message]` is still a pretty simple way to wrap the message and developers won't ever be accidentally printing that to stderr and not getting the log output. We can avoid having to parse by checking `startswith("[\"ulauncher-json-message")`, but in that case using the delimiter approach instead is probably better.

troycurtisjr:
I think perhaps there is an alternative that could potentially satisfy both our perspectives. IIUC a big issue with the previous setup was the need to wait around for the incoming connection from the extension before being able to do anything. Instead of using that approach, we can use a `socketpair` when launching the subprocess, which provides a unix connection that is already present immediately at launch, and utilizes the same underlying techniques as the std-pipes (only with a socket instead of 2 pipes). This lifts the messaging out of the stderr stream, while providing an immediately connected channel between parent and child. I've attached an example script showing its use.

[socketpair-exec.py](https://github.com/user-attachments/files/22085839/socketpair-exec.py)

There is a little juggling of file descriptors up front, but once they are in a stream object, they behave the same as the current stderr/stdin stream from an i/o standpoint.

friday:
> I think perhaps there is an alternative that could potentially satisfy both our perspectives. IIUC a big issue with the previous setup was the need to wait around for the incoming connection from the extension before being able to do anything. Instead of using that approach, we can use a `socketpair` when launching the subprocess, which provides a unix connection that is already present immediately at launch, and utilizes the same underlying techniques as the std-pipes (only with a socket instead of 2 pipes). This lifts the messaging out of the stderr stream, while providing an immediately connected channel between parent and child. I've attached an example script showing its use.
> 
> [socketpair-exec.py](https://github.com/user-attachments/files/22085839/socketpair-exec.py)
> 
> There is a little juggling of file descriptors up front, but once they are in a stream object, they behave the same as the current stderr/stdin stream from an i/o standpoint.

Looks great!

One potential deal breaker that came up for this PR yesterday, is if debugging requires that the user runs the extension separately. I hope we can find a solution where this is not necessary. So I will wait for this PR (#1505) before I continue with this one, but if I continue I will try to use this socket par approach instead.

friday:
Closing in favor of #1573

PR #1535, opened 2025-09-11
chore: improve the error message when mypy is not available

Mr-Sunglasses:


friday:
Implemented in #1536. It was actually pretty tricky due to weirdness with the makefile and shell environment so I spent some hours just on this. We unfortunately can't make this smarter due to https://superuser.com/a/1758397

PR #1537, opened 2025-09-14
chore: update git checkout action

friday:
Update checkout action and use double quotation marks (some linter I didn't know I had did that for me)


gornostal:
Thanks for keeping this up-to-date!

PR #1540, opened 2025-09-20
fix: timer method only works in GLib.MainContext

friday:
Maybe this isn't strictly speaking a fix. The `timer` method was always documented to work inside GLib.MainContext, and both the app and extension runtime run inside GLib.MainContext now.

But when I made a draft PR (#1534) where the extension runtime no longer runs inside GLib.MainContext, then this method would silently stop working. Even though this PR won't be merged in this state I found this to be a bit of a footgun, and would rather see it to work regardless of the context.

Furthermore, the unit tests were misleading imo. They didn't run inside of GLib.MainContext, and they also don't wait for the timer, so they were not really testing that the timer worked. `func.assert_called_once()` was effectively testing something else I think.

So I rewrote the timer method to use `threading.Timer` instead. It works similar to the GLib implementation but requires less code. I removed the repeat argument because we don't use it, and I think this would have required some more complexity to support with `threading.Timer` (not a lot, but I see no reason to add it if we don't need it).

And I (mostly copilot tbh) added more thorough tests.

I'm suspecting @troycurtisjr will say this is going to use more resources (or he wouldn't have implemented it like it was originally)?

### DIFF
--- a/tests/utils/test_timer.py
+++ b/tests/utils/test_timer.py
@@ -1,35 +1,65 @@
-from typing import Any
-from unittest.mock import MagicMock, Mock
-
-import pytest
-from pytest_mock import MockerFixture
+import time
+from unittest.mock import Mock
 
 from ulauncher.utils.timer import timer
 
+TIMER_BUFFER = 0.02  # seconds
+
+
+def padded_wait(seconds: float) -> None:
+    """Wait the given number of seconds plus a small buffer to ensure timer execution."""
+    time.sleep(seconds + TIMER_BUFFER)
+
 
 class TestTimer:
-    @pytest.fixture(autouse=True)
-    def glib(self, mocker: MockerFixture) -> MagicMock:
-        return mocker.patch("ulauncher.utils.timer.GLib")
-
-    def test_timer_subsecond(self, glib: MagicMock) -> None:
+    def test_timer_executes_function(self) -> None:
+        """Test that timer executes the function after the delay."""
         func = Mock()
-        subsecond_time = 0.5
-        ctx = timer(subsecond_time, func)
-        glib.timeout_source_new.assert_called_with(subsecond_time * 1000)
-        src: Any = ctx.source
-        ctx.trigger(None)
+        delay = 0.1
+
+        ctx = timer(delay, func)
+
+        # Function should not be called immediately
+        func.assert_not_called()
+
+        padded_wait(delay)
+
         func.assert_called_once()
         ctx.cancel()
-        src.destroy.assert_called_once()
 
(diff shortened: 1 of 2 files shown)


troycurtisjr:
The main issues with threads are the complexity of working in threads. Whatever function is run in the callback has a lot of non-obvious restrictions as a result in running in a different thread. Because of Python's Global Interpreter Lock (for now.. [PEP 703](https://peps.python.org/pep-0703/)), it is easier to get away with more things in Python without generating race conditions, but it is very non-obvious when you really need to be careful. In particular, anytime Python does a system call (I/O, create resources, select, etc), and very often when it calls into functions implemented in C (versus a library function written in Python) it releases its GIL. Without the GIL held, things may change in the middle of a Python function executing and could predicates to be violated as a result.  

Since thread race conditions are some of the hardest things to track down and reproduce, it probably just replaces one footgun for another one :grin: 

As long as we are careful what we do in the callback function, we should be OK. The general advice is if you don't use explicit locking, at most you should do is set/clear a flag, or push a message into a thread safe queue, or similar. 

It seems reasonable to use the Python built-in if we won't always have a Glib event loop, however I think we should just use the threading library directly then. Currently, the function doesn't really provide much value since the Threading usage is already trivial, and actually just obfuscates the fact that it is running in another thread, (even with the doc string indicating such). For something as tricky as running in other threads, I think we should have each place import `threading` explicitly to make it very clear to future readers that they should tread carefully.


On the testing, since these are unit tests it mocked out the GLib/timer part which was not implemented by the unit (`TimerContext`) itself, but by external libraries and with potential external side effects. Instead it was just testing the lifecycle of the object and exercising its methods. Note that the new test mostly tests `threading.Timer` and thus probably doesn't belong in application unit testing. Since this is Python, having a test that at least executes the function body probably makes sense, but it doesn't seem necessary to test the underlying standard library.

friday:
@troycurtisjr Good feedback.

I definitely agree that wrapping a native python method to save one line of code and adding a bunch of tests mostly testing the native code is questionable. The tests were actually written before the implementation became just a wrapper. I thought I commented on that, but I guess I forgot. The actual suggestion would be to replace `timer` with `threading.Timer`.

However, if this is not safe I think we should look for a different solution. This reminds me of a conversation we had years ago where I thought Python was not a good language for Ulauncher. I keep seeing more and more reasons for that.

I'll see if I can find another way to rewrite this not using either Threads or GLib.MainContext. We don't really need multi-threading, just a good solution for concurrency.

friday:
Closing in favor of https://github.com/Ulauncher/Ulauncher/pull/1573

PR #1544, opened 2025-09-22
fix: don't rely on ext controller cache to get extension by keyword

friday:
The old logic was using the extension controller cache as the source of truth, which I think we shouldn't do, because the cache doesn't mean the controllers are active or that it has all the extensions that are active (in case we need to invalidate/remove something from the cache).

This should be more efficient also since it now has a lookup table (dict) and doesn't have to loop.

ext_mode.get_triggers() is called when the window is shown, to refresh the data. So the lifetime of this cache is the same as that.


ulauncher/modes/extensions/extension_socket_server.py
```diff
@@ -100,11 +96,6 @@ def stop(self) -> None:
             self.service.close()
             self.service = None
 
-    def get_controller_by_keyword(self, keyword: str) -> ExtensionSocketController | None:
-        if ext_controller := ExtensionController.get_from_keyword(keyword):
```
gornostal, line 104: Dependencies are much cleaner without this :+1:

gornostal (approved):

PR #1552, opened 2025-10-04
refactor: in-memory extension registry

gornostal:
- follow single responsibility principle in extension module,
- replace extension cache with an explicit in-memory extension registry

The in-memory registry is necessary for the feature of preview extensions.

### DIFF
--- a/tests/modes/apps/extensions/test_extension_mode.py
+++ b/tests/modes/apps/extensions/test_extension_mode.py
@@ -8,9 +8,19 @@ from ulauncher.internals.query import Query
 from ulauncher.modes.extensions.extension_mode import ExtensionMode
 from ulauncher.modes.extensions.extension_socket_controller import ExtensionSocketController
 from ulauncher.modes.extensions.extension_socket_server import events as ess_events
+from ulauncher.utils import singleton
 
 
 class TestExtensionMode:
+    @pytest.fixture(autouse=True)
+    def reset_singleton(self) -> None:
+        # Reset singleton before each test
+        singleton._instances.pop(ExtensionMode, None)
+
+    @pytest.fixture(autouse=True)
+    def ext_registry(self, mocker: MockerFixture) -> MagicMock:
+        return mocker.patch("ulauncher.modes.extensions.extension_mode.ExtensionRegistry")
+
     @pytest.fixture(autouse=True)
     def ext_server(self, mocker: MockerFixture) -> Any:
         ess = mocker.patch("ulauncher.modes.extensions.extension_mode.ExtensionSocketServer").return_value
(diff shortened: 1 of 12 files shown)


gornostal:
It needs more work. I'll resubmit it soon

friday (commented):
The way I intended things to work these methods belong in the ext controller instance, at least that's how I intended it to work where the controller work on single instances. Before the ExtController there were lots of methods in other places taking the extension id as the argument to  do something on a single instance. And the point of the ExtController class was to have one class that would wrap all the other classes to avoid the cyclic imports at least for this tree. But there was no single helper module for operating on multiple extensions. The ext mode does that though, and the ext socket server too. I recently moved most of the logic from the socket server that doesn't belong there though.

So currently in the main branch, the ext server is the place that handles most of the "extension registry" logic, while the extension cache is in the controller class because the controller is the place you create/get controllers from.

I don't mind moving this to a separate `ExtensionRegistry` class though. What I consider to be important to preserve about the ExtController is having a convenient top-level instance that handles the other extension classes, so we can avoid circular references as far as possible, and also to make the methods intuitive and easy to use in the ExtController, although the underlying logic might be complicated (#1044).

Having said that I think these methods you mentioned for the reason for closing the PR belong to the ExtController, but if it adds complexity to have them there maybe having them on the ExtensionRegistry is not a bad idea?

I don't really believe in the "single responsibility principle". It is a bit like "separation of concerns" I think . You can decide one day that some set of concerns need to be separated and refactor, then the next day you decide that some other need to be separated and you end up with the first one being coupled/integrated again. Similarly what is and isn't the responsibility of a class can be fully arbitrary. As for the ExtController it's not strictly following the "single responsibility principle", but the responsibility of the ExtController is to control the extension, which is what makes it very wide. You can move responsibilities elsewhere and possibly wrap those from the controller (like we have with the ExtRunner, and ExtRemote), and I think it makes a lot of sense for those classes and moving the runner and remote logic into the controller would be a big mistake and make the code hard to maintain. And that's the reason why they should keep being separate, not because some idealistic programming pattern.

If it makes sense then we can move out more concerns we thing can be abstracted out, but likewise it has to make sense on it's own imo. Start simple by avoiding making too many abstractions. Separate the abstractions when there is a need.

As for an example if what I think is bad separation: The ItemNavigation class is tightly coupled with the window. The window has the ItemNavigation instance, the ResultWidget and the core class (previously QueryHandler) and it has to pass the other instances to the ItemNav. And the ResultWidget which is created and owned by the window uses `get_toplevel` to get the window, and in the on_click method it runs the activate method on the `nav`: `window.results_nav.activate(alt)`, so the ownership is super confusing here.

PR #1553, opened 2025-10-05
refactor: follow single responsibility principle in the extension module, use explicit in-memory extension registry

gornostal:
The main reason I need this is to have an in-memory extension registry. It will allow registering temporary (preview) extensions.
Other changes are to make the related classes cleaner.

### DIFF
--- a/ulauncher/modes/extensions/extension_controller.py
+++ b/ulauncher/modes/extensions/extension_controller.py
@@ -8,8 +8,7 @@ import tempfile
 from datetime import datetime
 from pathlib import Path
 from shutil import copytree, rmtree
-from typing import Any, Iterator
-from weakref import WeakValueDictionary
+from typing import Any
 
 from ulauncher import paths
 from ulauncher.cli import get_cli_args
@@ -27,6 +26,7 @@ from ulauncher.modes.extensions.extension_manifest import (
 )
 from ulauncher.modes.extensions.extension_remote import ExtensionRemote
 from ulauncher.modes.extensions.extension_runtime import ExtensionRuntime
+from ulauncher.utils.eventbus import EventBus
 from ulauncher.utils.get_icon_path import get_icon_path
 from ulauncher.utils.json_conf import JsonConf
 from ulauncher.utils.json_utils import json_load
@@ -60,8 +60,8 @@ class ExtensionState(JsonConf):
 
 
 logger = logging.getLogger()
-controller_cache: WeakValueDictionary[str, ExtensionController] = WeakValueDictionary()
 extension_runtimes: dict[str, ExtensionRuntime] = {}
+lifecycle_events = EventBus("extensions")
 
 
 def _load_preferences(ext_id: str) -> JsonConf:
@@ -69,6 +69,8 @@ def _load_preferences(ext_id: str) -> JsonConf:
 
 
 class ExtensionController:
+    """Manages the lifecycle of an extension."""
+
     id: str
     path: str
     state: ExtensionState
@@ -91,24 +93,12 @@ class ExtensionController:
             self.state.update(defaults)
 
     @classmethod
-    def create(cls, ext_id: str, path: str | None = None) -> ExtensionController:
-        if not path:
-            path = extension_finder.locate(ext_id)
-            if not path:
-                msg = f"Extension with ID '{ext_id}' not found"
-                raise ExtensionNotFoundError(msg)
-        if controller := controller_cache.get(ext_id):
-            return controller
-        new_controller = cls(ext_id, path)
-        controller_cache[ext_id] = new_controller
-        return new_controller
-
-    @classmethod
     async def install(
         cls, url: str, commit_hash: str | None = None, warn_if_overwrite: bool = True
     ) -> ExtensionController:
         logger.info("Installing extension: %s", url)
         remote = ExtensionRemote(url)
+        is_new_install = not Path(remote.target_dir).exists()
         commit_hash, commit_timestamp = remote.download(commit_hash, warn_if_overwrite)
 
         try:
@@ -120,7 +110,7 @@ class ExtensionController:
             rmtree(remote.target_dir)
             raise
         else:
-            controller = cls.create(remote.ext_id, remote.target_dir)
+            controller = cls(remote.ext_id, remote.target_dir)
             controller.state.save(
                 url=url,
                 browser_url=remote.browser_url or "",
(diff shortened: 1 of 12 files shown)


gornostal:
@friday _I'm moving the conversation from https://github.com/Ulauncher/Ulauncher/pull/1552#pullrequestreview-3302374715 to here_

>  while the extension cache is in the controller class because the controller is the place you create/get controllers from.

That's what I didn't like and wanted to refactor. ExtController is designed to control a single extension. The cache layer on top of it is more of a hack, which I wanted to replace with an explicit registry. 

> What I consider to be important to preserve about the ExtController is having a convenient top-level instance that handles the other extension classes

If by that you mean being able to get/iterate existing extensions, it's moved to the registry class now for the reasons above.

> Having said that I think these methods you mentioned for the reason for closing the PR belong to the ExtController

You and I had the same feeling about the install/uninstall methods then 😄  
I kept them in the controller class.

Re. class vs functions and the Extension prefix -- I totally agree with you. I think I just followed the bad examples blindly when working on this PR. I'll update the registry so it exports functions as you suggested.

Lastly, there are no circular dependencies in the proposed structure.

Please let me know if I missed any points from your review. And thanks for the review!

P.S. I probably should have kept the previous MR open instead of creating this one. I thought making another round of changes would actually take me another week and 10 more 180⁰ turns. So I wanted to reduce the noise from the commits.

friday:
I pushed the changes I proposed in the other PR and some other minor changes. Let me know if you agree or not.

(still not done with the review, but it was easier for me to push them first)

ulauncher/modes/extensions/extension_controller.py
```diff
@@ -233,6 +222,7 @@ async def update(self) -> bool:
                 raise
 
         await self.toggle_enabled(was_running)
+        lifecycle_events.emit("extensions:updated", self)
```
friday, line 225: This calls `extension_registry.updated()` which updates the registry to set this instance.

But if this instance was not already in the extension_registry, how did we get it?

ulauncher/modes/extensions/extension_socket_server.py
```diff
@@ -83,10 +86,11 @@ def handle_registration(self, framer: JSONFramer, event: dict[str, Any]) -> None
                 for msg_id in pended[1:]:
                     GObject.signal_handler_disconnect(framer, msg_id)
             ext_id: str | None = event.get("ext_id")
+            path: str | None = event.get("path")
             assert ext_id
+            assert path
             ExtensionSocketController(self.socket_controllers, framer, ext_id)
-            # TODO: This is ugly, but we have no other way to detect the extension started successfully
-            ExtensionController.create(ext_id).is_running = True
+            self.on_extension_registered(ext_id, Path(path))
```
friday, line 93: I don't think this is better than before. Sure, it looks better here, but you have the exact same logic in a method in the extension mode now instead, and you need extra logic to pass that method here.

The fundamental problem is needing ot set this property from the outside and it's still the same. Plus this adds more conflicts for #1534 where I actually did fix this proper (and I still haven't given up on that PR, I have kept rebasing it and improving it and trying to replace with socket pairs now)

ulauncher/modes/extensions/extension_mode.py
```diff
@@ -22,16 +26,25 @@ class ExtensionTrigger(Result):
...
 
     def __init__(self) -> None:
-        self.ext_socket_server = ExtensionSocketServer()
+        self.ext_socket_server = ExtensionSocketServer(self._on_extension_registered)
         self.ext_socket_server.start()
         events.set_self(self)
 
+    def _on_extension_registered(self, ext_id: str, _: Path) -> None:
+        """Callback when an extension successfully registers with the socket server."""
+        extension_registry.get(ext_id).is_running = True
```
friday, line 46: See https://github.com/Ulauncher/Ulauncher/pull/1553/files#r2408877650

ulauncher/modes/extensions/extension_registry.py
```diff
@@ -0,0 +1,69 @@
...
+    """Iterate over registered extension controllers."""
+    yield from _ext_controllers.values()
+
+
+def load(ext_id: str, path: str | None = None) -> ExtensionController:
+    """Load an extension controller into the registry without starting it."""
+    if not path:
+        path = extension_finder.locate(ext_id)
+        if not path:
+            _ext_controllers.pop(ext_id, None)
```
friday, line 37: I think this is a strange non-explicit side effect. If you call load it can do the opposite. Not that I haven't introduced side effects myself, but it should at least have a comment about why I think.

ulauncher/modes/extensions/extension_registry.py
```diff
@@ -0,0 +1,69 @@
...
+@lifecycle_events.on
+def installed(controller: ExtensionController) -> None:
+    """Handle install events dispatched by controllers."""
+    _ext_controllers[controller.id] = controller
+
+
+@lifecycle_events.on
+def updated(controller: ExtensionController) -> None:
+    """Handle update events dispatched by controllers."""
+    _ext_controllers[controller.id] = controller
```
friday, line 63: As I mentioned elsewhere I don't think we need `updated` as lifecycle event, because `_ext_controllers[controller.id]` will already be `controller`

Maybe with your upcoming PR for the preview feature this will be needed in that one though? (I would still suggest removing it from this PR and adding it there in that case).

friday (changes requested):
I have a lot of remarks and some of them needs to be fixed imo otherwise we will break the support for non-managed extensions as a fallback for example.

BUT all those were details, and overall I think this is looking great.

ulauncher/modes/extensions/extension_controller.py
gornostal: You're right, we don't need to listed for `updated` even in the ext. registry. I'll remove that

ulauncher/modes/extensions/extension_registry.py
friday: > It does, however, trigger reload dbus command, which would crash trying to load a deleted extension.

That makes sense :+1: I think if you add this as a comment so we understand this line in 6 months time when reading it, that's fine too.

friday (approved):
Since I found a bug in the main branch when reviewing this I wanted to fix it in main first. Then I had to fix the conflict and effectively addressed most of my own feedback here also. I thought it was easier to push the other comments too.

I was wrong on one of them since it was an async method. I added a TODO comment instead of that because I separately have been trying to make the start and stop methods non-async, so I think in the future we can do this.

You can see everything that was changed here (including an unrelated test fix to main that was not in this branch before): https://github.com/Ulauncher/Ulauncher/compare/9d696b9e95ca894d26edddc7a495adf24e666c75..58f8c1fe687bf80cb43d959140be69472b50a916

gornostal:
Thanks for the final review and cleanup!

PR #1560, opened 2025-10-26
feature: extension preview command #1505

gornostal:
See #1505

### DIFF
--- a/preferences-src/src/components/pages/ExtensionPreferences.vue
+++ b/preferences-src/src/components/pages/ExtensionPreferences.vue
@@ -37,10 +37,10 @@
              copy paths
           </a>
         </div>
-        <div class="notes alert-warning paddding-small" v-if="extension.shadowed_by_preview">
+        <div class="notes alert-warning padding-small" v-if="extension.shadowed_by_preview">
           <i class="fa fa-exclamation-triangle"></i> This extension cannot be used while another extension with the same ID is running in preview mode.
         </div>
-        <div class="notes alert-info paddding-small" v-if="extension.is_preview && extension.with_debugger && extension.is_stopped">
+        <div class="notes alert-info padding-small" v-if="extension.is_preview && extension.with_debugger && extension.is_stopped">
           <i class="fa fa-info-circle"></i> Extension is waiting for debugger connection (see console for details).
         </div>
       </div>
@@ -391,7 +391,7 @@ export default {
   font-style: italic;
 }
 
-.paddding-small {
+.padding-small {
   padding: 8px;
 }
 </style>


ulauncher/modes/extensions/extension_cli_handlers.py
```diff
@@ -130,3 +133,143 @@ def upgrade_extensions(_: ArgumentParser, args: Namespace) -> bool:
...
+
+
+def preview_extension(_: ArgumentParser, args: Namespace) -> bool:  # noqa: PLR0911, PLR0912, PLR0915 - intentionally single function for ease of maintenance
+    """
+    Run an extension in preview mode (without installing it).
+    Returns True on success, False otherwise.
+    """
+    # Check if Ulauncher is running
+    if not check_app_running(app_id):
+        logger.error("Error: Ulauncher is not running. Please start it first.")
```
friday: Nit: I don't like "Please start it first". Maybe "Error: Ulauncher needs to be running in order to preview extensions." or something.

ulauncher/modes/extensions/extension_mode.py
```diff
@@ -155,3 +155,104 @@ def trigger_event(self, event: dict[str, Any]) -> None:
...
+        )
+
+        # Stop the preview extension
+        stop_msg = f"[preview] Preview extension '{preview_ext_id}' stopped"
+        self.run_ext_batch_job([preview_ext_id], ["stop"], done_msg=stop_msg)
+
+        # Try to restart the original extension
+        try:
+            original_controller = extension_registry.get(original_ext_id)
+        except ExtensionNotFoundError:
```
friday: With the changes made to the get method, this should be `except KeyError` right?

friday (changes requested):
Overall great work, but I found some issues while testing and made some code comments. I'm not sure exactly how the debugging is supposed to work because I never debug python. I didn't get it to work, but it's likely I'm missing something.

I tested with for example https://github.com/Ulauncher/ulauncher-hash which I have installed already and have the source code locally. While it was running and enabled I ran the preview command which disabled it and started in debug mode instead (as expected). But:

1. If you save the preferences then restart Ulauncher, then the app wouldn't start next time (fixed in main here 66ada59e, but not in this branch yet until you rebase it).

Image

This was really caused by a bad assumption in main from the previous changes though.

2. I managed to enabling the disabled extension manually in the preferences and for example got both of them running at the same time:

Image

I don't think this should be possible. The "on" button for the overridden extension should be disabled imo, and "reload extensions" button should ignore the overridden extension.

3. If you run with the `--with-debugger`, press ctrl+c and then run with `--with-deubgger` again it errors with this message: `RuntimeError: Can't listen for client connections: [Errno 98] Address already in use  🔌 from com.github.ulauncher.ulauncher-hash.preview`

4. Also seems to only affect with debug:
I get this message and the preview extension view shows as "stopped" in the list view but not the extension details
Image 

Image

5. If Ulauncher is terminated, the preview process stays open, but then if you start Ulauncher again, it won't "reconnect" to the preview process, so in this case you have to ctrl+c and then restart. I was running both Ulauncher and the preview command from the same terminal window but different panes, so I ended up doing this accidentally, but still seems like a legit case. Not fully sure how to support it though. Maybe get the ulauncher pid when the preview runtime start and check it regularly and kill the preview process if the ulauncher pid dies.

gornostal:
Thanks for the review!
Very thorough. As always :) 

1. Rebased
2. Fixed. Now you won't be able to enable it or restart by clicking the reload button
3. Fixed. It was a bug in `run_ext_batch_job`
4. That's expected because the extension is waiting for a debug client to connect before executing the first line of code.
   I added a note to the pref. UI so the user knows what to do next.
5. Fixed. Same as point 3.

ulauncher/modes/extensions/extension_cli_handlers.py
gornostal: Fixed

ulauncher/modes/extensions/extension_mode.py
```diff
@@ -65,7 +65,7 @@ def handle_action(self, action_metadata: ActionMetadata | None) -> None:
     def get_triggers(self) -> Iterator[Result]:
         self._trigger_cache.clear()
         for ext in extension_registry.iterate():
-            if not ext.is_enabled or ext.has_error:
+            if not ext.is_enabled or ext.has_error or not ext.is_running:
```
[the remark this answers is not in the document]
gornostal: Pushed to `main`

friday (commented):
I tried to make time yesterday to review, but I didn't have time to test. The code looks good now. One concern is that I think you didn't disable or hide the "on" button for the `shadowed_by_preview` I think? So there is no indication that it's disabled. I'll try to test it tomorrow,

friday:
I just re-read and might add one more change request (small one).

gornostal:
Thanks for taking the time for this!

PR #1563, opened 2025-11-08
build: switch nix to use wrapGAppsHook3

nathanaelhoun:
Since two weeks, a nixos-unstable build using the flake for Ulauncher v6 break with the error:
```
error: 'wrapGAppsHook' has been renamed to/replaced by 'wrapGAppsHook3'
```

This PR replaces the reference to `wrapGAppsHook`. 

I have also removed the symlink created by a nix build in the repository, as I believe it's not needed for non-nix dev environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,5 @@ data/preferences
 .venv
 /nix/result
 /nix/dev
+/result
 ulauncher.1.gz
--- a/result
+++ b/result
@@ -1,1 +1,0 @@
-/nix/store/vcysl6bpqx7p2v43lk27v7vm4v67afjc-python3.13-ulauncher-v6


friday (approved):
Not tested but looks legit. Thanks!

friday:
I appreciate using conventional commits, but I will squash merge the PR and rename to use the prefix "build" instead if "fix" because it's a change to the nix build, not the app itself. I make mistakes like this too btw, but I fix them if I can catch myself before merging.

nathanaelhoun:
> I appreciate using conventional commits, but I will squash merge the PR and rename to use the prefix "build" instead if "fix" because it's a change to the nix build, not the app itself. I make mistakes like this too btw, but I fix them if I can catch myself before merging.

No problem, thanks for maintaining consistency!

PR #1568, opened 2025-11-17
feat: native gtk preferences

### LINKED ISSUE #1257: Rewrite Ulauncher preferences to use pure GTK (not webkit)

friday:
# Preferences Window Rewrite

Closes #1257

## Rewrite benefits
- Much lower memory usage (no web engine)
- Simpler architecture (no client-server communication)
- Simplified and faster build process
- Much smaller package size (and easier to package)
- Much faster startup (no WebKit initialization)
- No web dependencies (and no https://github.com/Ulauncher/Ulauncher/security warnings for outdated npm deps)

### User Experience
- Largely the UX same as before
- Respects system theme. I made some small overrides for themes because they can be really inconsistent, but I made them with safe principles and tested on different themes (light and dark themes).
- I didn't want to go overboard with the styling/small details as I was trying to respect the user themes and didn't want to spend too much time on the small details.
- Shortcuts editing experience looks better now.
- You can press ctrl+s to save changes now.
- The extensions list is kept up to date by polling and re-rendering if there are changes insteas of the "reload" button we used to have. The new app needs this because it can be started faster than the old app and showed extensions that hadn't been started yet as "stopped". But this turned out to be a very nice feature in general as it also updates when using the CLI or task killer etc to install, uninstall or stop extensions.
- Extensions that are "shadowed by the preview extension" are excluded from the list completely now (I think this makes more sense).

## Notes
- The code has a lot of hard coded "magic numbers" and similar, and lack utility methods/wrappers that could help make the code more readable, but also less flexible.
- There were some places we supported html, like the extension installation instructions. These won't work anymore, but we support pango markup now instead.. We can for example replace `<code>x</code>` with `<span face="monospace" bgcolor="#00000030">x</span>` to improve this, see: https://docs.gtk.org/Pango/pango_markup.html>, but I don't know if we want to do that.
- Gtk.Spinner won't be animated if animations are disabled in the Gsettings/dconf.
- Keyboard accessibility (tab index) hasn't been handled yet, but I think that was the case before also.
- When using scroll wheel or trackpad swipe/gesture to scroll and the cursor ends up over a combo box (select box), the scroll movement stops and the cursor changes to same as css "cursor: move" and doesn't change back after. This is a GTK bug imo.
- The cursor handling overall seems much worse than for web (no pointer/hand when over a button for example), but I don't think that's something we're supposed to handle. It doesn't seem easy to do (you can't set the cursor with normal css cursors at least).
- We could improve on hover effects for buttons and links (some themes have them and others do not).


gornostal (approved):
That's awesome! So much extra weight lost.

I'm curious, have you considered splitting ui and logic by defining UI in xml files using Glade? It would also be easier to build and preview screens in the visual editor.
Do you have performance concerns? I wonder how much slower it would work -- 20 -> 200ms kind of slower?

friday:
@Ulauncher/nix-contributors I updated default.nix‎ with the removed dependencies and the nix-build job is running, but I don't actually have Nix to test it.

Vortriz:
> @Ulauncher/nix-contributors I updated default.nix‎ with the removed dependencies and the nix-build job is running, but I don't actually have Nix to test it.

Did a quick run. Worked without any issues.

PR #1570, opened 2025-11-17
chore: replace type linter mypy with pyrefly

friday:
I tested Pyrefly when it was first released as alpha and was impressed, but it didn't seem to support out Gtk type stubs. [Now it just hit beta](https://github.com/facebook/pyrefly/releases/tag/0.42.0). I tested it again and it does support them now :+1:  It runs insanely fast, and it was even better in my IDE (Zed defaults to basedpyright, but the pyrefly extension made it **much** faster AND better)

I couldn't find any feature parity comparison with mypy, so I don't know if we should migrate away from mypy yet, so I didn't do that. But I noticed for example that despite being so fast, it had type inference for untyped code (which is why I had to add ignore comments, cast or similar.

To test it, run `make python-venv`  to install, then `pyrefly check` 

I also tested ty, which worked great and even ran faster, but it didn't feel as ready (more false positives), and it didn't infer untyped code to types.

The CI fails because it has an outdated ruff version in the dockerhub image, but it would still fail if it wasn't for that because we added pyrefly now which is not in ci either.

### DIFF
--- a/ulauncher/app.py
+++ b/ulauncher/app.py
@@ -29,7 +29,7 @@ class UlauncherApp(Gtk.Application):
     # So all methods except __init__ runs on the main app
     query = ""
     windows: WeakValueDictionary[Literal["main", "preferences"], Gtk.ApplicationWindow] = WeakValueDictionary()
-    _tray_icon: ulauncher.ui.tray_icon.TrayIcon | None = None
+    _tray_icon: ulauncher.ui.tray_icon.TrayIcon | None = None  # pyrefly: ignore[implicit-import]
 
     def __call__(self, *args: Any, **kwargs: Any) -> UlauncherApp:
         return cast("UlauncherApp", get_instance(super(), self, *args, **kwargs))
@@ -171,5 +171,5 @@ class UlauncherApp(Gtk.Application):
         self.hold() if value else self.release()
 
     @events.on
-    def quit(self) -> None:
+    def quit(self) -> None:  # pyrefly: ignore[bad-override]
         super().quit()
--- a/ulauncher/modes/extensions/extension_mode.py
+++ b/ulauncher/modes/extensions/extension_mode.py
@@ -5,7 +5,7 @@ import contextlib
 import html
 import logging
 from threading import Thread
-from typing import Any, Iterator, Literal
+from typing import Any, Callable, Iterator, Literal, cast
 
 from ulauncher.internals.query import Query
 from ulauncher.internals.result import ActionMetadata, Result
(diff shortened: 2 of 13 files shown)


ulauncher/modes/extensions/extension_mode.py
```diff
@@ -94,7 +95,10 @@ def activate_result(self, result: Result, query: Query, alt: bool) -> ActionMeta
         Called when a result is activated.
         Override this method to handle the activation of a result.
         """
-        handler = getattr(result, "on_alt_enter" if alt else "on_enter", DEFAULT_ACTION)
+        handler = cast(
+            "ActionMetadata | Callable[[Query], ActionMetadata]",
+            getattr(result, "on_alt_enter" if alt else "on_enter", DEFAULT_ACTION),
+        )
```
friday: We didn't have types for this. Mypy would just infer this as any, while pyrefly inferred as `bool | Any`

ulauncher/ui/tray_icon.py
```diff
@@ -99,7 +100,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         elif tray_icon_lib == "AyatanaIndicator":
             app_status = AyatanaIndicator.IndicatorCategory.APPLICATION_STATUS
-            self.aya_indicator = AyatanaIndicator.Indicator.new("ulauncher", icon_name, app_status)
+            self.aya_indicator = AyatanaIndicator.Indicator.new(
+                "ulauncher",
+                icon_name,
+                app_status,  # pyrefly: ignore[ bad-argument-type]
```
friday: Pyrefly detects that AyatanaIndicator can be one of two modules and that app_status can be a property of one of two modules, so it this is a false error because it's smart but not smart enough to understand AyatanaIndicator is the same class on row 103 and 104.

ulauncher/utils/ewmh.py
```diff
@@ -272,7 +273,7 @@ def getClientList(self):
         """
         return [self._createWindow(w) for w in self._getProperty("_NET_CLIENT_LIST")]
 
-    def getClientListStacking(self):
+    def getClientListStacking(self) -> list[xobject.drawable.Window]:
```
friday: This was untyped before, but pyrefly inferred the type from the usage. It uses an underlying method that can return all kinds of objects, but the way it's used will only return a this type.

ulauncher/app.py
```diff
@@ -171,5 +171,5 @@ def toggle_hold(self, value: bool) -> None:
         self.hold() if value else self.release()
 
     @events.on
-    def quit(self) -> None:
+    def quit(self) -> None:  # pyrefly: ignore[bad-override]
```
friday: I don't actually understand this error, but it works now, so I ignored it:

`UlauncherApp.quit` has type `BoundMethod[UlauncherApp, (...) -> Any]`, which is not assignable to `BoundMethod[UlauncherApp, (self: UlauncherApp) -> None]`, the type of `Application.quit`

friday:
> I would drop mypy completely instead of keeping it in CI.

I will keep it in ci (only) for now. It's intended to be short term and confirm that we don't have any type issues that pyrefly doesn't check for.

I learned a long time after upgrading pylint to ruff that bugs had began sneaking in that pylint would have caught, so I'd rather not do the same again. If it turns out that having mypy in the ci causes more pain than it helps then we can remove it then.

friday:
I'll merge this with mypy left for the CI @gornostal ^ We'll see how well that works (or not) and can always follow it up.

PR #1573, opened 2025-12-06
refactor: use socket pair for extension communication

friday:
This is very similar to, and substitutes #1534, but fixes all the problems mentioned in that branch by myself and others. I found a way where the extension runtime could use a GLib mainloop so we can use GLib timer functions safely, but yet use socket pairs as suggested by @troycurtisjr .

## Summary by CodeRabbit

* **Refactor**
  * Switched to an FD-based IPC flow with a unified mainloop, simplifying startup/shutdown and lifecycle handling.

* **New Features**
  * Added a JSON-over-socket messaging controller, a global event bus, and direct send/receive extension APIs with synchronous start and message dispatch.

* **Tests**
  * Added comprehensive messaging-controller tests and simplified/removed many legacy socket/framer tests.

* **Removed**
  * Deprecated legacy framing and socket-server/controller implementations and related tests.

* **Bug Fix**
  * Adjusted extension initialization order to reduce race conditions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/modes/apps/extensions/test_extension_runtime.py
+++ b/tests/modes/apps/extensions/test_extension_runtime.py
@@ -1,3 +1,4 @@
+import signal
 from typing import Any
 from unittest.mock import MagicMock, Mock
 
@@ -34,6 +35,11 @@ class TestExtensionRuntime:
     @pytest.fixture
     def time(self, mocker: MockerFixture) -> MagicMock:
         return mocker.patch("ulauncher.modes.extensions.extension_runtime.time")
+
+    @pytest.fixture
+    def mock_timer(self, mocker: MockerFixture) -> MagicMock:
+        """Mock the timer utility function used for scheduling delayed kills."""
+        return mocker.patch("ulauncher.modes.extensions.extension_runtime.timer")
 
     def test_run__basic_execution__is_called(self, subprocess_launcher: MagicMock) -> None:
         extid = "mock.test_run__basic_execution__is_called"
@@ -107,7 +113,8 @@ class TestExtensionRuntime:
             "Exited", 'Extension "mock.test_handle_exit" exited with code 9 after 5.0 seconds.'
         )
 
-    def test_stop(self) -> None:
+    def test_stop(self, mock_timer: MagicMock) -> None:
+        """Test that stop() closes the connection and schedules a kill timer."""
         extid = "mock.test_stop"
         exit_handler = Mock()
         runtime: Any = ExtensionRuntime(extid, ["mock/path/to/ext"], None, exit_handler)
@@ -115,5 +122,19 @@ class TestExtensionRuntime:
         runtime._subprocess.get_identifier.return_value = "ID"
         runtime._msg_controller = Mock()
         runtime.stop()
+
         runtime._msg_controller.close.assert_called_once()
-        # TODO: This doesn't test the SIGKILL signal ^
+        mock_timer.assert_called_once_with(0.5, runtime._kill)
+
+    def test_kill__sends_sigkill(self) -> None:
+        """Test that _kill() sends SIGKILL if the process is still running."""
+
+        extid = "mock.test_kill__sends_sigkill"
+        runtime: Any = ExtensionRuntime(extid, ["mock/path/to/ext"])
+
+        # Simulate process still running
+        runtime._subprocess.get_identifier.return_value = "ID"
+        runtime._kill()
+
+        # Verify SIGKILL is sent
+        runtime._subprocess.send_signal.assert_called_once_with(signal.SIGKILL)


ulauncher/ui/windows/preferences/utils/ext_handlers.py
```diff
@@ -99,7 +99,7 @@ def install_async() -> None:
             try:
                 ext = asyncio.run(ExtensionController.install(url))
```
gornostal: unrelated to this PR, but would be good to drop asyncio completely to simplify the code.

ulauncher/utils/message_socket.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from ulauncher.utils.gio_compat import Gio
+
+
+class MessageSocket:
+    """
+    Wraps a file descriptor from a socket pair to provide conventent read/write methods
+    """
```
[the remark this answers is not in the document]
gornostal: MessageSoket is fine. I also like MessageIO as the name.

Consider also renaming methods to `read` and `write` as the class  name already suggests that we're dealing with messages here.

ulauncher/utils/socket_msg_controller.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from ulauncher.utils.gio_compat import Gio
+
+
+class MessageSocket:
+    """
+    Wraps a file descriptor from a socket pair to provide conventent read/write methods
+    """
```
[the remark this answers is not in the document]
friday: I don't like any of the names I came up with myself. The short ones felt technically incorrect, and the technically correct ones were too long (some variant of `Socket[Message|IO][Controller|Handler]`).

I just renamed to `SocketMsgController` for now. It's at least a bit shorter than having to type out "Message". "IO" is even shorter, but that would make the name start with, "SocketIO" which is an extremely popular web socket library, and I don't want to imply get anyone confused that we use web sockets now.

And I shortened the method names per your suggestion. I thought about that before and forgot it.

ulauncher/ui/windows/preferences/utils/ext_handlers.py
friday: Yeah. We talked about this. I am not at all convinced that it will simplify the code. I think it will simplify some places and do the opposite in a bunch of other places. It might still be an overall improvement though.

ulauncher/modes/extensions/extension_mode.py
```diff
@@ -122,7 +126,9 @@ def run_ext_batch_job(
         async def run_batch_async() -> None:
             for job in jobs:
                 if job == "start":
-                    await asyncio.gather(*[c.start() for c in ext_controllers if c.is_enabled])
+                    for controller in extension_registry.load_all():
```
troycurtisjr: Should this be using the `ext_controllers` instead of going back to the registery to get all the controllers? (Caught by CodeRabbit)

ulauncher/modes/extensions/extension_runtime.py
```diff
@@ -79,6 +114,35 @@ def _kill(self) -> None:
...
         self.error_stream.read_line_async(GLib.PRIORITY_DEFAULT, None, self.handle_stderr)
 
+    def send_message(self, message: dict[str, Any]) -> None:
+        """Send a JSON message to extension."""
+        try:
+            json_str = json.dumps(message)
+            self.msg_socket.write(json_str)
+            logger.debug("Sent message to %s: %s", self.ext_id, message)
+        except Exception:
+            logger.exception("Failed to send message to %s", self.ext_id)
```
troycurtisjr: Should we close a connection or take some kind of action in these cases? Does it seem like it could actually be recoverable, such that the next message could be successful?

ulauncher/modes/extensions/extension_mode.py
friday: Yes. Having rebased the old branch and fixed a lot of conflicts I don't remember why it was like this, but it makes no sense. Fixed in b7865628

ulauncher/modes/extensions/extension_runtime.py
friday: I reworked the patch in 8b5788d4 now

troycurtisjr (approved):

ulauncher/modes/extensions/extension_mode.py
```diff
@@ -55,22 +62,19 @@ def handle_query(self, query: Query) -> None:
...
-        if self.active_ext and isinstance(action_metadata, list):
-            for result in action_metadata:
-                result["icon"] = self.active_ext.get_normalized_icon_path(result.get("icon"))
-
-            results = [Result(**res) for res in action_metadata]
-            events.emit("window:show_results", results)
-        else:
-            events.emit("mode:handle_action", action_metadata)
+        msg = f"Query not valid for extension mode '{query}'"
+        raise RuntimeError(msg)
```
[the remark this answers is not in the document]
friday: @coderabbitai. This is intentional. If the error is caught your suggestion would mean the extension reference would be using the outdated value instead of clearing

tests/utils/test_socket_msg_controller.py
```diff
@@ -0,0 +1,119 @@
...
+        self, controller_pair: tuple[SocketMsgController, socket.socket]
+    ) -> None:
+        """Test that closing the peer calls the controller on_close callback."""
+        controller, peer = controller_pair
+        on_close = Mock()
+        controller._on_close = on_close
+        controller.listen(lambda _on_msg: None)
+        peer.close()
+        process_pending_events()
+        assert on_close.called
```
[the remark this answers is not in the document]
friday: @coderabbitai direct access to private attribute in tests are fine. Do not request changes for this again. Only for direct access to private attribute outside of tests.

PR #1589, opened 2025-12-24
refactor: remove readthedocs build integration

friday:
Since we have moved these to the new static website build we should remove them here to avoid editing the data in the wrong place, getting outdated docs here or having to edit in both.


gornostal:
Thanks for this!

PR #1604, opened 2025-12-29
chore: add AGENTS.md

friday:
Add AGENTS.md (both for coding agents and coderabbit - the code review agent)

I tried to steer the agents towards producing better code and this has been an evolving document I have used for the last two weeks.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,315 @@
+# Guide for AI Coding Assistants
+
+This document contains guidelines and context for AI coding assistants working on the Ulauncher codebase.
+
+## General Guidelines
+
+### Communication
+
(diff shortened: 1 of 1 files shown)


friday:
@gornostal I'll merge this. Just wanted you to be aware because I saw you had some agents.md file in some local working branch. If you have suggestions make that a new PR and I'll look at it.

I tried to make this terse overall and for "smarter" agents like Claude rather than adding a ton of elaborate code examples for everything, but I'm not sure what you're using. I'm not sure I do want all of the current examples for modes, so I might trim that out later and just ask it to check core and BaseMode.

gornostal:
I have my agents file in git exclude. But my looks too generic compared to this. So let's keep it

PR #1632, opened 2026-01-14
fix: Don't set margin for window managers

SapphirusBeryl:
Essentially what this PR purposes to do is only set frame margins for a declared desktop environment with compositing. With the `None` constant present for `DESKTOP_ID`, assume the environment is a window manager and don't apply a margin to the window frame.

Opposed to adding yet another option, or re-adding the `--no-window-shadow`, this is perhaps a good compromise for allowing blur and shadow effects to be applied by both composited X11 and Wayland window managers. 

End result:

<p align="center">
01KEWSS4ANXH4TAE2VNBTW8BXP
</p>

Opposed to: 

<p align="center">
01KEWSP4BET0R33G2DYNAV7RWB
</p>

This can still be worked around in different ways. For example, hyprland provides the `ignore_alpha` field for layer rules, although this only works under hyprland with the GTK Layer Shell enabled (which is rather limiting). You still won't be able to apply compositor effects under X11 window managers or Wayland compositors of which don't implement such functionality.

Only caveat to this updated approach is the `DESKTOP_ID` variable has to be set in order for client-sided effects like `box-shadow` to be applied. Prior to this, in version 5.15.x, a CSS class setting a negative 20 pixel margin was applied with `--no-window-shadow`. 

Though far from ideal in my opinion, this workaround can still be replicated by setting a negative margin for the `app` class with a user theme, for example:

```css 
.app {
   margin: -20px;
}
```

Alternatively document a workaround? I'm not sure of any one-good solution to this particular edge case.

Relevant discussion: https://github.com/Ulauncher/Ulauncher/discussions/1349


friday:
I've been trying to think what's the most reasonable way to handle this, and I think this is something that probably apply to the same cases as the setting you just created (that I just renamed...).

Maybe renaming the setting to something more broad like, "Handle window decorations and positioning". Then we could add small print below the description to explain that this is for users who want to configure this for their window manager/DE instead

SapphirusBeryl:
> I've been trying to think what's the most reasonable way to handle this, and I think this is something that probably apply to the same cases as the setting you just created (that I just renamed...).
> 
> Maybe renaming the setting to something more broad like, "Handle window decorations and positioning". Then we could add small print below the description to explain that this is for users who want to configure this for their window manager/DE instead

Would that mean we'd instead check for `layer_shell` being `False` or something? This would also apply to i3 with Picom, for example, as well.

friday:
> Would that mean we'd instead check for `layer_shell` being `False` or something? This would also apply to i3 with Picom, for example, as well.

I think `layer_shell` would need to be renamed (again) to for exampled `managed_by_wm` or some other name that makes sense to clarify this is not about layr shell. It's about opting out completely of trying to handle positioning and decorations.

Then if `managed_by_wm` is true we apply no positioning or shadows (even for x11). But we should still check separately if `self.is_composited()` is True and disable shadows regardless in that case imo.

SapphirusBeryl:
> > Would that mean we'd instead check for `layer_shell` being `False` or something? This would also apply to i3 with Picom, for example, as well.
> 
> I think `layer_shell` would need to be renamed (again) to for exampled `managed_by_wm` or some other name that makes sense to clarify this is not about layr shell. It's about opting out completely of trying to handle positioning and decorations.
> 
> Then if `managed_by_wm` is true we apply no positioning or shadows (even for x11). But we should still check separately if `self.is_composited()` is True and disable shadows regardless in that case imo.

That might not actually be a bad idea. It'd solve the issue of having to parse `XDG_DESKTOP_NAME` in addition to everything else that's being done. But then, it's no longer fully automatic. And you still have the edge cases wherein other compositors don't provide the option to negate semi-opaque surfaces.

PR #1648, opened 2026-02-13
fix: local install OSError

wrental:
Local install of most recent ulauncher beta(27)  with `sudo pip install ./` from root of git repo resulted in error `OSError: usr/share/ulauncher` when  running `ulauncher` from terminal.  
This does not affect usage with `make run`, only local install.

Source of error:  
`ulauncher/paths.py`, line 9 - needs `/local`:  
`ASSETS = os.path.abspath(os.environ.get("ULAUNCHER_SYSTEM_DATA_DIR", f"{SYSTEM_PREFIX}/share/ulauncher"))`  
\>>> `ASSETS = os.path.abspath(os.environ.get("ULAUNCHER_SYSTEM_DATA_DIR", f"{SYSTEM_PREFIX}/local/share/ulauncher"))`

---
## Summary by cubic
Fixes OSError when running ulauncher after a local pip install. ASSETS now defaults to {SYSTEM_PREFIX}/local/share/ulauncher, matching XDG_DATA_DIRS and preventing missing path errors.

<sup>Written for commit 0f57780de47204ffd79a247fcfa7ced1401cf145. Summary will update on new commits.</sup>

### DIFF
--- a/ulauncher/paths.py
+++ b/ulauncher/paths.py
@@ -6,7 +6,7 @@ APPLICATION = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 # ULAUNCHER_SYSTEM_PREFIX is used by a third party packagers like Nix
 SYSTEM_PREFIX = os.environ.get("ULAUNCHER_SYSTEM_PREFIX", sys.prefix)
 # ULAUNCHER_SYSTEM_DATA_DIR is used when running in dev mode from source and during tests
-ASSETS = os.path.abspath(os.environ.get("ULAUNCHER_SYSTEM_DATA_DIR", f"{SYSTEM_PREFIX}/share/ulauncher"))
+ASSETS = os.path.abspath(os.environ.get("ULAUNCHER_SYSTEM_DATA_DIR", f"{SYSTEM_PREFIX}/local/share/ulauncher"))
 HOME = os.path.expanduser("~")
 XDG_DATA_DIRS = os.environ.get("XDG_DATA_DIRS", f"/usr/local/share/{os.path.pathsep}/usr/share/").split(os.path.pathsep)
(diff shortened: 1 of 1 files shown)


friday (changes requested):
What is sys.prefix for you? Mine contains "local", and since no one else complained I think that's the normal case. It seems like something is different on your system, and if this was merged as it is now, it would break for everyone else.

wrental:
my sys.prefix is just `/usr`, which explains it. a quick search reveals that might be due to a weird python install - ill troubleshoot some other time. i'm also having trouble with startup on login so symptomatic of a larger issue, but if no one else has the issue then likely not a problem. 

for a local beta install is pip the best way to do it, or do you recommend something else?

friday:
> all appears to be working as intended from terminal, but using gnome app launcher (which i'm assuming targets the desktop entry in `~/.local/share/applications`) does not. running `gapplication launch io.ulauncher.Ulauncher` from terminal works, as does `ulauncher` (does the same thing, verified with `--dev` option) but i neither have the option to launch at login from within ulauncher settings nor to launch at startup with gnome-tweaks. iirc you're on ubuntu, but what DE?
> 
> I can also move this to discussions or issues if you'd prefer, lmk

We don't officially support pip install, because it's out of our control where pip places things. You can investigate if it's fixable or create your own desktop entry. There is some related info in https://github.com/Ulauncher/Ulauncher/discussions/1348

Also, I guess you could add a make script that runs pip install and fixes the desktop icon after, but then you have to do that for the uninstall command also.

wrental:
ah okay. ill double check the execution perms which seems most likely to be the issue given that pip correctly installed in `~/.local/share/applications` and i did already validate, and exec is the same as the command that works in terminal. writing this from my phone so ill double check that in a bit but otherwise might have to settle for running gapp launch in terminal ¯\\_(ツ)_/¯ not the worst thing in the world. either way excited for official v6 release, tysm for all your work 🙌

PR #1656, opened 2026-03-13
feature: add multiline option to `Result` for wrapping result description text

dnewcome:
I'm submitting this PR to support a shell plugin I wrote:
https://github.com/dnewcome/ulauncher-term

I'm not sure if this is a feature you would want in the mainline, but the way Ulauncher handles key events makes it really amazing for seeing the results of an edited command instantly. When the result item is taller than usual (enabled by this patch) the icon is centered on the left which looks a little weird. I didn't explore changing that but I'd consider looking into it if you want to merge this.

I used Claude here to grok this codebase faster and write up a summary of my changes, but I did manually edit and review everything and run the tests. I ended up requiring `gi` to test this. I'm not sure if there's a way to  do it without pulling in another dependency.

- dnewcome

## Summary

- Adds a multiline boolean field to Result (default False) that controls how the description label is rendered in ResultWidget
- When multiline=True, the description label wraps text (WORD_CHAR mode) instead of truncating with an ellipsis, and the text container aligns to the top so multi-line content doesn't get vertically centered
- Adds PyGObject to requirements.txt so UI tests that import gi.repository can be type-checked and run without implicit dependency on the system install
- Adds tests covering both the multiline and ellipsizing (non-multiline) cases

## Motivation

Extensions that display structured or longer descriptions (e.g. multi-line file previews, error messages, command output, help text) currently have no way to show the full content — it always gets truncated. This opt-in flag lets extension authors render descriptions that wrap naturally within the result item.

---
## Summary by cubic
Adds an opt-in multiline flag to `Result` so description text can wrap instead of being ellipsized, with multi-line items top-aligned. Includes tests and adds `PyGObject` for UI tests.

- **New Features**
  - `Result.multiline` (default False) enables wrapped descriptions and top-aligns the text container; default behavior stays ellipsized.
  - Added `multiline` to `Result.__init__` to avoid a prototype mismatch.
  - Tests cover both multiline wrapping and non-multiline ellipsizing.

- **Dependencies**
  - Added `PyGObject` to `requirements.txt` for Gtk/Pango in tests.

<sup>Written for commit 41c3d948c0f0f9825f2b929bc2a6b6ffab29365e. Summary will update on new commits.</sup>

### DIFF
--- a/ulauncher/internals/result.py
+++ b/ulauncher/internals/result.py
@@ -39,6 +39,7 @@ class Result(BaseDataClass):
         self,
         *,
         compact: bool | None = None,
+        multiline: bool | None = None,
         highlightable: bool | None = None,
         searchable: bool | None = None,
         name: str | None = None,
@@ -55,6 +56,8 @@ class Result(BaseDataClass):
         init_kwargs: dict[str, Any] = {}
         if compact is not None:
             init_kwargs["compact"] = compact
+        if multiline is not None:
+            init_kwargs["multiline"] = multiline
         if highlightable is not None:
             init_kwargs["highlightable"] = highlightable
         if searchable is not None:


gornostal:
@dnewcome What I suggest here is for you to keep using the Ulauncher implementation from your branch while we are working on next version features that will include a solution for that problem 

What I have in mind for the next version is a completely new result type suitable for multi-line rich text with the ability of result streaming. It will help solve the problem you are facing and it will also be useful for extensions that stream LLM output.

PR #1674, opened 2026-04-13
feat: implement algorithm for app ranking based on recent launch history

### LINKED ISSUE #1434: feat: improve handling of "frequent apps"
### Description

The current logic doesn't care about or store data to weight up recent activity. See https://github.com/Ulauncher/Ulauncher/discussions/1272#discussioncomment-11708452

### Communication guidelines

### Ulauncher v6

burgerphilic18:
Closes: #1434

Description:
This PR improves the frequent apps logic by replacing the launch counter with a scoring algorithm. Currently, Ulauncher only stores the total number of launches, which means old favorite apps can remain at the top of the search results even after the user has switched to a new frequent app.

This change introduces a more dynamic system that weights recent launches much more heavily than old ones ensuring the list stays correct to current usage patterns. Last 10 app launch timestamps are stored to calculate the score.

I made this scoring pattern:

< 1 day = 100
< 3 day = 80
< 7 day = 60
< 14 day = 40
< 30 day = 20
older than that = 10

Total Count = 0.1/launch

Final Score = (total_count * 0.1) + sum(points_for_each_recent_launch)

---
## Summary by cubic
Replaces the simple launch counter with a recency‑weighted ranking so recently used apps appear first. Stores the last 10 launches and computes a score that better reflects current usage.

- **New Features**
  - Recency-based scoring per launch: <1d 100, <3d 80, <7d 60, <14d 40, <30d 20, else 10; plus 0.1 per total launch.
  - Persists last 10 launch timestamps and trims the list on bump.
  - Ranks apps by computed score in `get_top_app_ids`.
  - Backward-compatible load of `app_starts.json` (int-only entries mapped to `{count, last_launches: []}`).

- **Bug Fixes**
  - Hardened data loading: uses `BaseDataClass`; validates and coerces `count` (int or digit string) and `last_launches` (int/float to float), trimming to 10.

<sup>Written for commit 59d9f7559a1806fd88fcd6c07e27b2c00a1ab2af. Summary will update on new commits.</sup>

### DIFF
--- a/ulauncher/modes/apps/app_result.py
+++ b/ulauncher/modes/apps/app_result.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import contextlib
 import logging
 import operator
+import time
 from os.path import basename
+from typing import TypedDict
 
 from ulauncher import paths
 from ulauncher.gi import GioUnix
@@ -12,7 +14,18 @@ from ulauncher.utils.json_utils import json_load, json_save
 
 logger = logging.getLogger()
 app_starts_path = f"{paths.STATE}/app_starts.json"
-app_starts: dict[str, int] = json_load(app_starts_path)
+
+class AppStartData(TypedDict):
+    count: int
+    last_launches: list[float]
+
+_raw_app_starts = json_load(app_starts_path)
+app_starts: dict[str, AppStartData] = {}
+for _app_id, _data in _raw_app_starts.items():
+    if isinstance(_data, int):
+        app_starts[_app_id] = {"count": _data, "last_launches": []}
+    elif isinstance(_data, dict):
+        app_starts[_app_id] = {"count": _data.get("count", 0), "last_launches": _data.get("last_launches", [])}
 
 
 class AppResult(Result):
@@ -44,8 +57,29 @@ class AppResult(Result):
 
     @staticmethod
     def get_top_app_ids() -> list[str]:
-        sorted_tuples = sorted(app_starts.items(), key=operator.itemgetter(1), reverse=True)
-        return [*map(operator.itemgetter(0), sorted_tuples)]
+        scores = {app_id: AppResult._calculate_score(data) for app_id, data in app_starts.items()}
+        sorted_ids = sorted(scores.items(), key=operator.itemgetter(1), reverse=True)
+        return [app_id for app_id, score in sorted_ids]
+
+    @staticmethod
+    def _calculate_score(data: AppStartData) -> float:
+        now = time.time()
+        score = data["count"] * 0.1
+        for launch_time in data["last_launches"]:
+            diff = now - launch_time
+            if diff < 86400:
+                score += 100
+            elif diff < 259200:
+                score += 80
+            elif diff < 604800:
+                score += 60
+            elif diff < 1209600:
+                score += 40
+            elif diff < 2592000:
+                score += 20
+            else:
+                score += 10
+        return score
 
     def get_searchable_fields(self) -> list[tuple[str, float]]:
         frequency_weight = 1.0
@@ -62,6 +96,11 @@ class AppResult(Result):
         ]
 
     def bump_starts(self) -> None:
-        starts = app_starts.get(self.app_id, 0)
-        app_starts[self.app_id] = starts + 1
+        data = app_starts.get(self.app_id, {"count": 0, "last_launches": []})
+        data["count"] += 1
+        data["last_launches"].append(time.time())
+        if len(data["last_launches"]) > 10:
+            data["last_launches"].pop(0)
+
+        app_starts[self.app_id] = data
         json_save(app_starts, app_starts_path)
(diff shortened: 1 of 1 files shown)


friday:
Sorry for the AI overload :D Did you trigger the copilot review? I haven't found those to be great myself and have been trying a couple other free ones to evaluate.

Will try to take a look at this soon, but this whole week will be hard for me to find time.

ulauncher/modes/apps/app_result.py
```diff
@@ -3,16 +3,43 @@
...
+    _valid_launches: list[float] = []
+
+    if isinstance(_data, int):
+        _valid_count = _data
+    elif isinstance(_data, dict):
+        _raw_count = _data.get("count")
+        if isinstance(_raw_count, int):
+            _valid_count = _raw_count
+        elif isinstance(_raw_count, str) and _raw_count.isdigit():
+            _valid_count = int(_raw_count)
```
friday, line 36: This data was at no point stored as a string (if it was the old app_starts would not work), so these  lines are redundant.

ulauncher/modes/apps/app_result.py
```diff
@@ -3,16 +3,43 @@
...
 logger = logging.getLogger()
 app_starts_path = f"{paths.STATE}/app_starts.json"
-app_starts: dict[str, int] = json_load(app_starts_path)
+
+
+class AppStartData(BaseDataClass):
+    count = 0
+    last_launches: list[float] = []
+
+_raw_app_starts = json_load(app_starts_path)
```
friday, line 23: I have been trying to made cleaner abstractions in the code base to avoid needing manually working with json config files, but for this case it wasn't possible. I reworked that into a new class (#1684), so it is now.

ulauncher/modes/apps/app_result.py
```diff
@@ -44,8 +57,29 @@ def from_id(app_id: str) -> AppResult | None:
 
     @staticmethod
     def get_top_app_ids() -> list[str]:
-        sorted_tuples = sorted(app_starts.items(), key=operator.itemgetter(1), reverse=True)
-        return [*map(operator.itemgetter(0), sorted_tuples)]
+        scores = {app_id: AppResult._calculate_score(data) for app_id, data in app_starts.items()}
+        sorted_ids = sorted(scores.items(), key=operator.itemgetter(1), reverse=True)
+        return [app_id for app_id, score in sorted_ids]
```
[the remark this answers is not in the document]
friday: This is a pre-existing issue, although made worse by this logic. I fixed this in the main branch now (aa9117a46)

ulauncher/modes/apps/app_result.py
```diff
@@ -44,8 +71,29 @@ def from_id(app_id: str) -> AppResult | None:
...
+            elif diff < 259200:
+                score += 80
+            elif diff < 604800:
+                score += 60
+            elif diff < 1209600:
+                score += 40
+            elif diff < 2592000:
+                score += 20
+            else:
+                score += 10
```
friday, line 95: This list of breakpoints can be expressed with an algorithm using math functions instead. See 5633301b1 (this is very close to your breakpoints except it doesn't max out at 2592000, which may or may not be a good thing)

friday (commented):
Kudos for taking this on. I hadn't even started thinking about this feature myself but now I have and it's quite a heavy integrated part of the app and lots of decisions I want to get right to be able to merge it.

I thought the code base needed initial refactoring in order to add something like this cleanly to it, so I added those and then tried to fix the conflicts, but I can't push to this branch. I made a new branch with this branch but rebased (which moved the code a lot, but it's still the same logic). I also fixed the linting issues and. See #1691.

If you want to continue working on on this branch without handling the conflicts you can made a new branch from that one.

But I'm still not sure this is the best approach. This means we compare the total app launches vs the last 10. It may not be possible in a way that's fair and accurate, and it's hard to determine a suitable half life for the recency algorithm. For example you could run a small business and do accounting once a month, but you do it a lot on that day/days, so it's one of top 5 apps, while every time you need to start working on it, it'd be very much out of the recent apps. But the day(s) you do you it, you open it a lot. Then you quickly bump it up in the top apps after a few times, and it gradually falls away in the days after. Maybe that's fine? I don't know the expectations users would have. But I think I prefer it to be more stable.

I have been thinking a simple capped log file where you append the app id and calculate the score based on all the line numbers in the log matching the app id, relative to the total. So if you opened Firefox.app on line 1 in the log, that's going to become less and less relevant the longer the log grows, and if you go for vacation a month and come back, it won't change the balance.

We get the recency "for free" without needing to store dates.

I don't know what's a reasonable cap, but more data will affect performance.

Do you think that would work?

Edit: I think I have an even better idea that would accomplish the same but  more performant and with minimal changes.

friday:
#1701 was just merged instead. Thanks again for your time!

PR #1680, opened 2026-04-26
chore: make container and ci use venv too 

friday:
Make all environments use venv and align with the new `make venv` ways where we only create/update the venv if it's missing or outdated.

Pros:
1. No need to rebuild docker image after changing requirements.txt
2. We can restore the fix for pyrefly editor integrations without breaking CI

Cons:
1. CI will be slower on cache misses (GHA caches expire after 7 days of no access)

---
## Summary by cubic
Unifies local, CI, and dev container to share the same Python virtualenv via `make venv`. The venv only rebuilds when missing or when `requirements.txt` changes, so no Docker rebuild is needed and the `pyrefly` editor fix stays enabled.

- **Refactors**
  - GitHub Actions: cache `.venv` keyed by `requirements.txt` and `.github/workflows/tests.yml`; run `make venv` quietly.
  - Dockerfile: remove baked dependency installs and `NO_VENV`; rely on runtime venv.
  - Makefile/container: `venv` prints why it rebuilds and skips when fresh; always add `.venv/bin` to `PATH`; persist venv in the container by mounting `.container-env/.venv`.

<sup>Written for commit ce1024eeacddca3a72584c707886dc2b59d867a6. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,6 +12,12 @@ jobs:
     container: ulauncher/build-image:6.6
     steps:
       - uses: actions/checkout@v5
+      - uses: actions/cache@v4
+        with:
+          path: .venv
+          key: venv-${{ hashFiles('requirements.txt') }}
+      - name: venv
+        run: make venv QUIET=1
       - name: ruff
         run: make ruff
       - name: pyrefly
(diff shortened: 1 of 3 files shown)


friday:
Closing due to all the noise created by bad bot advice. Will make a new identical PR which also will use the same cache key (hopefully proving that works).

PR #1691, opened 2026-05-01
feat: frequency ranking

friday:
Rework of #1674

The original PR exposed a lot of things that needed improvements in main first, so I reworked the structure and semantics a lot for the sake of this feature. I tried to push that to the original branch but the author disabled that.

The main difference from this branch and #1674 is that all the logic involving this data was moved into a dedicated module (#1688), renamed to `app_history.py` (f30248ec0) and implemented with a new JsonKeyValueConf helper class #1684 to support typed data for this JSON configs in this shape.

I also added cache for app ranking (aa9117a46) to address one of the copilot complaints.and the `get_top_app_ids` method was renamed to get_app_ranking.

All these changes are already merged and the old branch is rebased on top of them, which heavily changes the code implementation.

I also fixed linting issues and rewrote the decay algorithm 5633301b1 (in the branch).

I still have some opinions/thoughts about this solution, but I figured this is a better base to start the discussion.

---
## Summary by cubic
Add frequency-based app ranking with exponential decay using recent launch times, replacing simple count sorting. Adds a small ranking cache and a typed history format to improve accuracy and performance.

- **New Features**
  - Ranks apps by score: count weight + decayed recent launches (half-life 5 days), keeping up to 10 timestamps per app.
  - Caches the computed ranking and invalidates on bump.
  - Stores typed history entries (count + last_launches) with value coercion and validation.

- **Migration**
  - Replace get_top_app_ids() calls with get_app_ranking().
  - Existing app_starts.json with integer counts is auto-coerced; no manual steps required.

<sup>Written for commit 5633301b176116abe5b6f5d9b0bae582ed5e6051. Summary will update on new commits.</sup>

### DIFF
--- a/ulauncher/modes/apps/app_history.py
+++ b/ulauncher/modes/apps/app_history.py
@@ -2,26 +2,43 @@ from __future__ import annotations
 
 import operator
 import time
-from typing import Any, TypedDict
+from typing import Any
 
 from ulauncher import paths
+from ulauncher.utils.base_data_class import BaseDataClass
 from ulauncher.utils.json_conf import JsonKeyValueConf
 
 _APP_HISTORY_PATH = f"{paths.STATE}/app_starts.json"
 
 
-class _AppHistoryData(TypedDict):
-    count: int
-    last_launches: list[float]
+class _AppHistoryData(BaseDataClass):
+    count = 0
+    last_launches: list[float] = []
 
 
 class _AppHistory(JsonKeyValueConf[str, _AppHistoryData]):
     _ranking_cache: list[str] | None = None
 
     def _coerce_value(self, value: Any) -> _AppHistoryData:
+        if isinstance(value, _AppHistoryData):
+            return value
+        valid_count = 0
+        valid_launches: list[float] = []
+
         if isinstance(value, int):
-            return {"count": value, "last_launches": []}
-        return {"count": value.get("count", 0), "last_launches": value.get("last_launches", [])}
+            valid_count = value
+        elif isinstance(value, dict):
+            raw_count = value.get("count")
+            if isinstance(raw_count, int):
+                valid_count = raw_count
+            elif isinstance(raw_count, str) and raw_count.isdigit():
+                valid_count = int(raw_count)
+
+            raw_launches = value.get("last_launches")
+            if isinstance(raw_launches, (list, tuple)):
+                valid_launches = [float(x) for x in raw_launches if isinstance(x, (int, float))][-10:]
+
+        return _AppHistoryData(count=valid_count, last_launches=valid_launches)
 
     def get_app_ranking(self) -> list[str]:
         if self._ranking_cache is None:
@@ -36,8 +53,8 @@ class _AppHistory(JsonKeyValueConf[str, _AppHistoryData]):
     @staticmethod
     def _calculate_score(data: _AppHistoryData) -> float:
         now = time.time()
-        score = data["count"] * 0.1
-        for launch_time in data["last_launches"]:
+        score = data.count * 0.1
+        for launch_time in data.last_launches:
             diff = now - launch_time
             if diff < 86400:
                 score += 100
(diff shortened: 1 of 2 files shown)


ulauncher/modes/apps/app_history.py
```diff
@@ -1,25 +1,77 @@
...
+import time
+from typing import Any
+
 from ulauncher import paths
+from ulauncher.utils.base_data_class import BaseDataClass
 from ulauncher.utils.json_conf import JsonKeyValueConf
 
 _APP_HISTORY_PATH = f"{paths.STATE}/app_starts.json"
+_MAX_APP_HISTORY = 10
+_SEARCH_SCORE_HALF_TIME = 5 * 24 * 60 * 60  # 5 days
```
friday, line 13: These two matched the original PR. I haven't put much thought into good values. _SEARCH_SCORE_HALF_TIME roughly aligned with the previous breakpoints.

Increasing `_SEARCH_SCORE_HALF_TIME` should help if you for example use different apps on different days. For example you are self-employed, working with software engineering, but you need to handle billing and taxes with accounting software once a month. With a 5 days half time the accounting software would be "halved" six times between every time you use it. Maybe that's what we want because it's so rarely used, but it's debatable. It could be one of your top 5 apps but used one day per month.

Increasing `_MAX_APP_HISTORY` means the opposite. If you open the accounting software 10 times in a short time span because you use it heavily that day, then it could quickly go from a low to a high rank, which makes history a bit unstable.

We still use count, which works to stabilize, but I'm not sure if we want to keep that long term? Maybe we do? If we do want to keep it, then the importance of these two doesn't matter as much

ulauncher/modes/apps/app_history.py
```diff
@@ -1,25 +1,77 @@
...
+        score = data.count * 0.1
+        # Exponential decay: score halves every SEARCH_SCORE_HALF_TIME, approaches 0 for old launches
+        for launch_time in data.last_launches:
+            score += 100 * math.exp(-_SCORE_DECAY * (now - launch_time))
+        return score
+
     def bump(self, app_id: str) -> None:
-        self[app_id] = self.get(app_id, 0) + 1
+        data = self.get(app_id)
+        if not data:
```
friday, line 66:
```suggestion
        if data is None:
```

Nit: data used to be an int and it's now a dict. Both int `0` and dict `{}` will evaluate to false, so it's better to be explicit.

I don't think it can actually go wrong in this case though.

ulauncher/modes/apps/app_history.py
```diff
@@ -1,25 +1,77 @@
...
+        valid_launches: list[float] = []
+
+        if isinstance(value, int):
+            valid_count = value
+        elif isinstance(value, dict):
+            raw_count = value.get("count")
+            if isinstance(raw_count, int):
+                valid_count = raw_count
+            elif isinstance(raw_count, str) and raw_count.isdigit():
+                valid_count = int(raw_count)
```
friday, line 38: This is not needed. The count was never stored as a string.

ulauncher/modes/apps/app_history.py
```diff
@@ -1,25 +1,77 @@
...
+
+
+class _AppHistoryData(BaseDataClass):
+    count = 0
+    last_launches: list[float] = []
 
 
-class _AppHistory(JsonKeyValueConf[str, int]):
+class _AppHistory(JsonKeyValueConf[str, _AppHistoryData]):
     _ranking_cache: list[str] | None = None
```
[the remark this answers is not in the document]
friday: This is funny. I said this would happen in https://github.com/Ulauncher/Ulauncher/pull/1690#discussion_r3174186486 when the cache was implemented:

> When you bump an app is when it's launcher and Ulauncher closes the window. This means the next time the app ranking cache is used the window by necessity must have reopened and triggered the app mode to clear it. Don't you agree?
>
> I did this originally anyway, but then another review agent I ran locally complained that I didn't invalidate it on .update(), setitem, delitem etc and I don't want to add all that code just to satisfy agents.

But then I added `self.clear_ranking_cache()` to the bump function anyway, and you did the same thing as the local review agent.

friday:
Closing, but will keep the branch open.

PR #1701, opened 2026-05-04
feat: app launch recency ranking

### LINKED ISSUE #1434: feat: improve handling of "frequent apps"

friday:
## App launch recency ranking

Replaces simple integer launch-count ranking with an adaptive decay model that gives more weight to recent usage. Inspired by #1674, but uses an event-driven approach rather than storing timestamps.

### How it works

The conceptual model is an append-only log where each launch adds the app's ID. An app's rank is proportional to how recently its entries appear Older entries are naturally outweighed by newer ones as the log grows.

Instead of actually storing a growing log, this simulates that behavior with a float score per app: on every launch, all scores are multiplied by a small decay factor, then the launched app gets +1. The decay factor adapts to history depth. The more total launches recorded, the smaller the decay per event. So the relative effort to overtake an established app stays proportional regardless of how long you've been using Ulauncher.

Concretely: if app A has N launches worth of history, app B needs about 19% of N launches to overtake it (at the default `DECAY_RATE = 3`). The constant controls aggressiveness. Higher values make old habits fade faster, lower values make rankings more stable.

### Strengths

- **Event-based, not time-based**: not using your computer or Ulauncher specifically for a priod doesn't shift your rankings.
- **No timestamps stored**: no record of when you launched each app (could be considered a privacy concern).
- **No extra state**: total launch count is recovered from the sum of scores. No separate counter needed.
- **Clean migration**: old integer counts convert to the new scale without needing two separate measures or an arbitrary weighting between them.

### Weaknesses

- The decay calibration can be changed later, but existing scores were accumulated under the old value. The effective decay rate during the transition will be somewhere between old and new, gradually self-correcting over subsequent launches. This is the same class of approximation as the initial migration. Which also can't retroactively weight old launches by recency.
- `DECAY_RATE` is not user-configurable. It could be, but then it wouldn't apply for old launch data.

### Migration

Creates `app_rankings.json` from `app_starts.json` on first run (skipped if the file already exists). Old integer counts are normalized to the new float scale. `app_starts.json` is left in place. After migration the initial ranking order matches the old frequency order; subsequent launches start shaping it by recency.

### Renames

- `app_history.py` / `AppHistory` → `app_rankings.py` / `AppRankings` (because the rankings are constantly reshaping now I think this terminology is more appropriate)
- `get_app_ranking()` → `get_app_ids()`


friday:
@burgerphilic18 Thanks again for your contribution. I had to think about it for a bit and came up with this counter suggestion. Do you think this works for you?

burgerphilic18:
> @burgerphilic18 Thanks again for your contribution. I had to think about it for a bit and came up with this counter suggestion. Do you think this works for you?

Hey, sorry for the delayed response, I was bit busy irl with university exams. Yes, this absolutely works for me. The decay model is a massive improvement over my timestamp approach which removes the privacy concern of storing timestamps and increasing the memory efficiency of the single float score.

friday:
I'll merge it. I don't feel certain about the amount of decay, but I guess we can always tune that.

gornostal:
Great idea with the decay feature!

Thank you both!

PR #1704, opened 2026-05-10
refactor: emphasize ui file structure as an isolated part

friday:
Rebased and more complete version of #1673

Enforces a clean GTK import boundary
* only modules in ulauncher/ui/ are allowed to touch GTK
* modules or parts of modules previously outside of ulauncher/ui/ that touched GTK, have now been moved into ulauncher/ui/
* removed gtk imports from gi.py

The only exceptions now is:
* `ulauncher/main.py` importing the ui (only when it's needed).
* `ExtensionController.get_normalized_icon_path()` lazy loads `get_icon_path` from the ui. This is only used from the app, so it's safe but it's a footgun we really should get rid of next.

There are also some "utils" like xinit and ewmh that I think should be moved too, but not in their current shape.


friday:
Both remaining :robot: issues are really the same issue, which
A: Worked exactly the same before this refactor
B: Isn't even a problem (it's part of Gtk, which we depend on)

PR #1743, opened 2026-06-07
feat(extensions): progressive (streamed) responses from generator listeners

PhilMeyr:
## Motivation

AI/LLM extensions are now among the most popular Ulauncher extensions (ChatGPT, Gemini, Ollama, Mistral…), and they all share the same UX problem: an LLM answer takes several seconds to complete, and today the extension API only allows **a single rendered response per request** — the user stares at "Loading…" until the full answer arrives, while every chat UI they know renders tokens as they stream.

This PR makes generator listeners stream: results are rendered progressively, exactly like the API v3 `yield` syntax already suggests. Follows up on the exchange in discussion #1736 ("Making it actually progressive is harder and hasn't been requested until now, but it should be possible").

Tested end-to-end on GNOME Wayland with a real extension streaming SSE chunks from the Mistral API ([demo branch](https://github.com/PhilMeyr/ulauncher-mistral/tree/feat/streaming-demo)).

## Design

- **Extension side** (`api/extension.py`): when a listener returns a generator, the accumulated results are sent as responses marked `"partial": true`, throttled (default 0.1s, tunable via `ULAUNCHER_PARTIAL_RESPONSE_INTERVAL`), followed by an unthrottled final response. Yielding a `Result` appends; yielding a `list[Result]` **replaces** the accumulated results — needed for content that grows over time (an LLM answer) rather than a growing list of items. The same semantics apply when generators are collected without streaming (shared helper `effect_utils.append_or_replace`).
- **App side** (`extension_mode.py`): `handle_response` keeps `_pending_callback` alive while responses are partial (result lists only — a malformed partial is logged and ends the request). The existing `request_id` staleness check applies to partials unchanged, and the extension stops consuming the generator as soon as its request is superseded.
- **Capability negotiation**: the app advertises support via the `ULAUNCHER_PARTIAL_RESPONSES=1` env var. Without it, extensions collect generators into a single response (today's behavior) — old app + new extension and new app + old extension both keep working. No manifest or API version change.
- **Robustness**: if the generator raises mid-stream, the error is logged and the accumulated results are sent as the final response, so the request always terminates.
- **Selection preservation** (`ui/`): re-renders previously reset the keyboard selection to the default, which would make the selection jump on every partial. `show_results` now carries deliberate (user-moved) selections over to the next render by result name. This also fixes the existing selection jump when the "Loading…" placeholder is replaced.

## Notes

- The UI callback was already safely re-callable (it's how the placeholder replacement works); the single-response constraint was only in `handle_response` and in the generator collection.
- No extension API change: `Iterable[Result]` listeners and `yield` are already legal — they just render progressively now.
- Out of scope (pre-existing behavior, unchanged): extension exceptions are not surfaced to the UI — a listener that raises still only logs in the extension process. Streaming extensions can catch their own errors and yield error results.
- Tests cover: partial/final sequencing, throttling, replace-vs-append semantics, superseded-request abort, mid-stream generator failure, no-capability fallback, the capability handshake, partial handling and staleness app-side, and selection preservation.
- I read CONTRIBUTING's preference for discussing features first — happy to move this conversation to a Discussion if you prefer; opening as a PR to make the proposal concrete, following up on #1736.

---
## Summary by cubic
Adds progressive (streamed) responses for extension generator listeners so results render as they arrive, greatly improving UX for LLM-backed extensions. Backward compatible: without the capability flag, generators are collected into a single final response.

- **New Features**
  - Stream partial result lists when `ULAUNCHER_PARTIAL_RESPONSES=1`, throttled by `ULAUNCHER_PARTIAL_RESPONSE_INTERVAL` (default 0.1s), followed by an unthrottled final response.
  - Yield semantics: yielding a `Result` appends; yielding `list[Result]` replaces the list (supports growing content like LLM answers).
  - App keeps the pending callback alive for partials and replaces the rendered list on each partial; staleness check via `request_id` still applies.
  - Generator errors mid-stream log and still produce a final response; iteration stops immediately when a request is superseded.
  - App advertises capability to extensions via `ULAUNCHER_PARTIAL_RESPONSES=1`; no manifest or API version change. Docs updated.

- **Bug Fixes**
  - Preserve deliberate keyboard selection across result re-renders to prevent selection jumps (also fixes jump when replacing the “Loading…” placeholder).

<sup>Written for commit 03d1109a41741faee987abc2a561f77032abc0d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Ulauncher/Ulauncher/pull/1743?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/architecture/extension-ipc.md
+++ b/docs/architecture/extension-ipc.md
@@ -56,3 +56,31 @@ Extensions run in **separate processes** as the same user as Ulauncher (not sand
 4. Extension handles action → may send new results back
 
 All communication is asynchronous using GLib's event loop.
+
+## Partial (Streamed) Responses
+
+Extension listeners that return a generator stream their results progressively: the
+accumulated results are sent as responses marked `"partial": true`, throttled to at most
+one per `ULAUNCHER_PARTIAL_RESPONSE_INTERVAL` seconds (default 0.1), followed by a final
+unmarked, unthrottled response with the complete list. Yields landing within the
+interval are not lost: they are included in the next response — the next partial, or the
+final one if nothing else is yielded. Ulauncher keeps the response callback alive until
+the final response, so each partial replaces the rendered list.
+
+`ULAUNCHER_PARTIAL_RESPONSES=1` is exported by the app when spawning extension
+processes. `ULAUNCHER_PARTIAL_RESPONSE_INTERVAL` is only read by the extension process
+from its inherited environment — a tuning knob the app never sets.
+
+Yielding a `Result` appends it to the streamed list; yielding a `list[Result]` replaces
+it, so producers can update results they already sent (e.g. an LLM answer growing as
+tokens arrive). If the generator raises, the error is logged and the accumulated results
+are sent as the final response, so the request always terminates.
+
+This only happens when the app advertises support via the `ULAUNCHER_PARTIAL_RESPONSES=1`
+environment variable; otherwise generators are collected into a single response, because
+older apps treat the first response as final. Stale partials are discarded through the
+existing `request_id` check; when newer input supersedes the request, the extension stops
+consuming the generator and sends nothing more — not even a final response.
+
+Primary use case: extensions backed by slow producers (LLM APIs, network searches) that
+can render results as they arrive.
--- a/tests/api/test_extension.py
+++ b/tests/api/test_extension.py
@@ -1,10 +1,14 @@
+from __future__ import annotations
+
 import logging
 import os
-from unittest.mock import patch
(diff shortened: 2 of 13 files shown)


friday:
Fixed in #1764 now.

PR #1746, opened 2026-06-07
refactor: add glib utils module with timer and GLib.idle_add wrapper

friday:
We already had a timer method, but this makes it use GLib.timeout_add internally so it can  share a context class and helper for a new GLib.idle_add wrapper: `run_when_idle`.

This point of this is that `GLib.idle_add` has a major foot gun: The callback function return type determines if the function should loop or not. We never want it to loop, and we want to avoid the risk of accidentally making it loop. by changing the return type (it happened in the past).

And since AI bots also nagging about using `GLib.idle_add`, this should get rid of the annoying comments.

Unfortunately it's not possible to use the ruff `tool.ruff.lint.flake8-tidy-imports.banned-api` rule to ban these APIs since they are functions, not modules.

---
## Summary by cubic
Add `glib_utils` with safe scheduling helpers and replace direct `GLib.idle_add` calls across the app. This prevents accidental rescheduling, unifies timers, and simplifies cancellation.

- New Features
  - Added `ulauncher/utils/glib_utils.py` with `run_when_idle`, `timer`, and `SchedulerContext` (idempotent `.cancel()`).
  - `_run_once` ensures callbacks run once and return False, avoiding `GLib.idle_add` looping.

- Refactors
  - Replaced all `GLib.idle_add` usages with `run_when_idle` and rewrote timers to use `timer` (`GLib.timeout_add` under the hood).
  - Removed `ulauncher/utils/timer.py` and `TimerContext`; updated imports and types.
  - Simplified spinner timing in prefs to use seconds with the new `timer`.
  - Tests: removed old timer tests; added coverage for `glib_utils` (context cancel, args forwarding, single-run behavior).

<sup>Written for commit 4ef1e43fd7b50cdaec48e945abc0450345f5f478. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Ulauncher/Ulauncher/pull/1746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


friday:
Going a different rout I think

PR #1784, opened 2026-07-31
nix: change the nix version to be aligned with the actual app version

dasj19:
Hi, I am a NixOS user using the NixOS unstable channel with flakes, and lately I was unable to build Ulauncher 6 unless the version string in nix matched the one in "_version.py"
 
 Relevant logs from build before the suggested change:
 ```
> Executing pythonMetadataCheckPhase
       > The 'ulauncher' derivation has version 'v6' but .dist-info/METADATA specifies version '6.0.0b31'.
       > This usually means that the wrong version is hardcoded in pyproject.toml or setup.{py,cfg}.
       > Use the pyprojectVersionPatchHook or patch the version manually so that the project metadata matches the derivation's version.

```
 
 I am unsure if the proposed change is the optimal one, as I am not very familiar with the codebase... but at least I am raising this issue that one can not use the latest Ulauncher 6 builds on NixOS/unstable using flakes.

<a href="https://cubic.dev/pr/Ulauncher/Ulauncher/pull/1784?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -33,8 +33,18 @@
 }:
 let
   pname = "ulauncher";
-  version = "v6";
 
+  # nixpkgs' pythonMetadataCheckHook compares this to the built dist-info metadata.
+  # Therefore the version string is extracted from _version.py.
+  versionFile = builtins.readFile ../ulauncher/_version.py;
+  versionMatch =
+    builtins.match ''
+      .*version = "([0-9]+\.[0-9]+\..*)".*
+    '' versionFile;
+  version =
+    if versionMatch == null then
+      throw "Failed to parse version"
+    else builtins.elemAt versionMatch 0;
   src.python = ../.;
 
   packages.tests.python = pp: (with pp; [

### CODE REVIEW

File: nix/default.nix [hunk 1]
```diff
@@ -33,7 +33,7 @@
 }:
 let
   pname = "ulauncher";
-  version = "v6";
+  version = "6.0.0b34";
```
friday: I think this should work, and prevent hard coding a value we have to bump repeatedly?

```suggestion
  # nixpkgs' pythonMetadataCheckHook compares this to the built dist-info metadata.
  version = lib.pipe ../ulauncher/_version.py [
    builtins.readFile
    (builtins.split ''version = "([^"]+)"'')
    (builtins.filter builtins.isList)
    builtins.head
    builtins.head
  ];
```

friday (commented):
Thanks for reporting this and the PR :+1: 

Apparently nixpkgs added validation to check the specified vs built version some ~20 days ago: [NixOS/nixpkgs#532778](https://github.com/NixOS/nixpkgs/pull/532778) 

They added a couple of escape hatches, but since `v6` was never a valid version string anyway we should probably rather fix than suppress it (like you did).

But I don't want to hardcode it. `_version.py` is already the single source of truth, and I think the code suggestion I made (with some AI help) will work to make it dynamic? Please verify (I don't use Nix).

dasj19:
> But I don't want to hardcode it. `_version.py` is already the single source of truth, and I think the code suggestion I made (with some AI help) will work to make it dynamic? Please verify (I don't use Nix).

Thanks for the feedback. I added code that extracts the version string from _version.py ... and I tested it locally.

friday (approved):

friday:
Merged. Thanks again!

PR #1793, opened 2026-08-16
fix: schedule focus-out hide on GLib main loop, not a Python thread

### LINKED ISSUE #1792: Intermittent SIGSEGV in on_focus_out_event — GTK called from non-main thread
## Summary

`UlauncherWindow.on_focus_out_event` (in `ulauncher/ui/windows/UlauncherWindow.py`) schedules `self.hide()` via `threading.Timer(0.07, ...)`:

```python
def on_focus_out_event(self, widget, event):
    self.is_focused = False
    t = threading.Timer(0.07, lambda: self.is_focused or self.hide())
    t.start()
```

`threading.Timer` fires its callback on a plain Python thread, not the GLib main loop. GTK/GDK are not thread-safe, so calling `self.hide()` off the main thread intermittently segfaults deep inside GDK/GTK's internal signal-emission machinery (`gdk_window_hide` / `gtk_widget_unmap` cascade).

## Evidence

Six `systemd-coredump` SIGSEGV crashes over about a month (`coredumpctl list ulauncher`), all in the `python3.14` interpreter running Ulauncher. `coredumpctl debug`'s backtrace of the crashing thread shows it is **not** the main GTK thread, with the stack:

```
PyGObject ffi_call → gtk_widget_hide → repeated g_signal_emit/gtk_widget_unmap → gdk_window_hide → crash inside libglib-2.0
```

The bug itself appears to have been present for a while (found identical code in Timeshift snapshots going back to at least June), but the crash is intermittent/racy since it depends on the 70ms focus-out/focus-in timing lining up with the GTK main loop's own activity.

## Suggested fix

Replace the `threading.Timer` with `GLib.timeout_add`, which schedules the callback on the main loop instead of a background thread:

```python
def on_focus_out_event(self, widget, event):
    self.is_focused = False
    GLib.timeout_add(70, lambda: self.is_focused or self.hide())
```

`GLib` is already imported in this file, so this is a self-contained one-line change.

## Environment

- Ulauncher 5.15.15 (Ubuntu PPA build)
- Python 3.14.4
- KDE Plasma / X11

skjerns:
## Summary

Fixes #1792.

`on_focus_out_event` used `threading.Timer` to delay the hide check by 70ms. That runs the callback on a separate Python thread, and GTK/GDK are not thread-safe — calling `self.hide()` from that thread intermittently segfaults deep inside GDK/GTK's signal-emission internals (`gdk_window_hide`/`gtk_widget_unmap` cascade).

This switches to `GLib.timeout_add`, so the delayed check runs on the main loop like every other GTK call in the codebase, instead of a bare Python thread.

## Evidence

Confirmed via `coredumpctl` on a system with several months of SIGSEGV crash history: every available core dump shows the same signature — segfault inside `libglib-2.0` reached through `gdk_window_hide`/`gdk_window_withdraw` via a `gtk_widget_unmap`/`g_signal_emit` cascade, on a thread that is not the process's main thread. `threading.Timer` here is the only place in the codebase that invokes a GTK method from a spawned thread, which matches.

## Test plan

- [x] `grep` confirms `threading` had no other use in the file after removing the import
- [ ] Manually verified no more crashes after a period of normal use with rapid focus-out/focus-in (e.g. Alt-tabbing, switching input sources)

<a href="https://cubic.dev/pr/Ulauncher/Ulauncher/pull/1793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ulauncher/ui/windows/UlauncherWindow.py
+++ b/ulauncher/ui/windows/UlauncherWindow.py
@@ -2,7 +2,6 @@
 import os
 import time
 import logging
-import threading
 
 import gi
 gi.require_version('Gtk', '3.0')
@@ -151,8 +150,13 @@ class UlauncherWindow(Gtk.Window, WindowHelper):
         # this is a simple workaround to avoid hiding window
         # when user hits Alt+key combination or changes input source, etc.
         self.is_focused = False
-        t = threading.Timer(0.07, lambda: self.is_focused or self.hide())
-        t.start()
+
+        def check_focus_and_hide():
+            if not self.is_focused:
+                self.hide()
+            return GLib.SOURCE_REMOVE
+
+        GLib.timeout_add(70, check_focus_and_hide)
 
     def on_focus_in_event(self, *args):
         if self.settings.get_property('grab-mouse-pointer'):


friday:
Thank you!

